PR #3, opened 2019-10-15
New test/test target teams distribute nowait

josemonsalve2:
Taken from 
https://bitbucket.org/crpl_cisc/sollve_vv/pull-requests/48/new-test-test-target-teams-distribute


josemonsalve2:
Josh,

Be ready to hate us. We had a conversation about this code and Swaroop raised a good point. She does not think that having a data race on the C array is the best solution. I agree. This is because this is non conformant with the OpenMP standard.

So we have decided to go with a better approach:
```
int useless[N_TASKS][N];
int order[N_TASKS]; // Each possition represents the order in which each task executed
int ticket[1] = {0};

#target enter data map(to: ticket[:], order[:])

for (int i = 0; i < N_TASKS; ++i) {
  #pragma omp target teams distribute map(alloc:useless[i][:], ticket[:])
  for (int j = 0; j < N; ++j) {
    //Create amount of work that makes the
    //first task take longer than the last one
    for (int k = 0; k < N*(N_TASKS-i); ++k)
      useless[i][j] += k*i*j; // Value does not really matter
    }
    // Capture the ticket
    int myTicket = 0;
    #pragma omp atomic capture
    myTicket = ticket[0]++;
    order[i] = myTicket;
  }
}
# pragma omp taskwait

#target exit data map(from:ticket[:], order[:]) 

// Order[0] must be equal to number of tasks:

OMPVV_ERROR_IF(ticket[0] != N_TASKS, "The test registered different number of target regions than spawned");

int wasAsync = 0;
for (int i = 0; i < N_TASKS; ++i) {
  if (order[i] != i) {
    wasAsync == 1;
    break;
  }
}

OMPVV_WARNING_IF(!wasAsync, "We could not detect asynchronous behavior between target regions")
```
What do you think?

josemonsalve2:
This is a test

jhdavis8:
I have updated this test to use a ticket-based approach, with some slight modifications to Jose's posted code. Right now, only xlc is showing the desired asynchronous behavior. Also, xlf is not compiling, complains about non-contiguous memory in the map clause on line 45.

tests/4.5/target_teams_distribute/test_target_teams_distribute_nowait.F90
```diff
@@ -0,0 +1,79 @@
...
+    INTEGER,DIMENSION(1):: ticket
+
+    ticket(1) = 0
+    errors = 0
+    was_async = 0
+
+    !$omp target enter data map(to: ticket(1:1), order(1:N_TASKS), my_ticket)
+
+    DO x = 1, N_TASKS
+       !$omp target teams distribute map(alloc: work_storage(x, 1:N), ticket(1:1)) private(my_ticket) nowait
```
jhdavis8: xlf won't compile because of non-contiguous memory here, I believe it relates to the work_storage mapping in particular.

PR #4, opened 2019-10-15
New test/test target enter data classes

josemonsalve2:
Taken from 
https://bitbucket.org/crpl_cisc/sollve_vv/pull-requests/7/new-test-test-target-enter-data-classes

This test aims to check mapping of classes as supported in OpenMP 4.5.
OpenMP 4.5 does not support the use of this pointer. For this reason many work arounds have to be used every time we are trying to reference a member of the classes. There are two tests. One of the uses inheritance and the other one does not.


josemonsalve2:
See issue #17

josemonsalve2:
@dmcdougall I have made changes to this test, which will be the base to modify the one in test/target_enter_exit_data.  They are both really similar. But I will split the PRs for both

tests/4.5/target_enter_data/test_target_enter_data_classes_simple.cpp
```diff
@@ -0,0 +1,146 @@
...
+#define N 1000
+
+class Simple {
+private:
+  int *d_array;
+  int size;
+  int sum;
+
+public:
+  // Constructor. Maps the data into the de3vice
```
dmcdougall: Got a typo here: `de3vice`.

tests/4.5/target_enter_data/test_target_enter_data_classes_simple.cpp
```diff
@@ -0,0 +1,146 @@
...
+  Simple(int s) : size(s) { 
+    this->sum = 0;
+    this->d_array = new int[size];
+    std::fill(d_array, d_array+size, 0);
+    // Removing the  direct use  of attributes to avoid problems
+    // with 4.5 specifications 
+    int* helper = d_array;
+    int &hs = size;
+    int &hsum = sum;
+#pragma omp target enter data map(to: helper[0:hs]) map(to: hs) map(to:hsum)
```
dmcdougall: Why take the reference of an `int`?  It's not a heavy-weight object.

And a related question: does this do what you expect?  I.e., does mapping an `int &` map the whole `int`?

tests/4.5/target_enter_data/test_target_enter_data_classes_simple.cpp
```diff
@@ -0,0 +1,146 @@
...
+        h_array[i] = helper[i];
+    }
+  }
+  int* getArray() {
+    return d_array;
+  }
+  
+  int* getSum() {
+    return &sum;
+  }
```
dmcdougall: This is where I _would_ return an `int &` (rather than a raw pointer) if I wanted my client to see a modifiable property.  Just my preference, of course :)

tests/4.5/target_enter_data/test_target_enter_data_classes_simple.cpp
josemonsalve2: I believe that some compilers take the whole this-> pointer too serious. I would have to test it again to see if this was the real reason, but I think that using a variable inside that is a member implied a use of the this pointer in the implicit mapping.

This is part of the whole confusion with the this pointer and 4.5

I will try it out without it and see what happens.

dmcdougall: I think there's a misunderstanding here.

I understand why you're avoiding referring to member symbols directly, that's not what I'm asking.  What I'm asking is:  why did you do `int &hsum = sum;` and not `int hsum = sum;`?

josemonsalve2: I wanted to make sure what gets map is the same memory location. If we map without the reference then it won't correspond to the same location across target regions.  However, since here we are mapping always tofrom it should not make a difference. 

One possible change would be to do:
```C
/// On constructor
int &hsum = sum;
#pragma omp target enter data map(to: hsum)
///...
/// on modify
int &hsum = sum 
#pragma omp target map (alloc:hsum)
/// ...
/// on get values
int &hsum = sum 
#pragma omp target exit data map(from:hsum

```

dmcdougall: Ahhh, I see.  Thanks for explaining that.

Issue #11, opened 2019-10-15
segfault in test_error_verbose 


josemonsalve2:
When sollve_vv checks a result by invoking `OMPVV_TEST_VERBOSE`, it passes a constant literal as the second argument to `test_error_verbose` here. However, inside the implementation of` test_error_verbose` the string literal is being modified. This causes a segmentation fault for seven tests (in my case).

A different approach would be to copy the `conditionStr` argument to `clean_condition` before the `DO WHILE` loop and modify only the copy in the loop. Doing this locally on my box fixes up those seven run-time failures.

josemonsalve2:
@dmcdougall Here We are

dmcdougall:
You folks moved home?  :)

josemonsalve2:
Yeah!, SOLLVE team suggested we are all in the same spot! I miss bitbucket already!

dmcdougall:
Mis condolencias.

:)

dmcdougall:
For what it's worth, I don't care where you folks live as long as it is clear where to open bug reports and pull requests.  Are you expecting all that to take place on GitHub now?  If so, you might consider putting something in the BitBucket readme that points to the GitHub repo and disable issue reporting on the BitBucket repo.  Just a thought.

josemonsalve2:
Yes, we will make the other repo read only and remove the options to create new issues and PRs. We will have a bot updating it, but also add all the information in the README and all that! 

It's so much fun moving haha! I also have to update our website which has bitbucket all over it

dmcdougall:
Understood.

josemonsalve2:
Thanks for your help Damon!

dmcdougall:
No problem.

PR #18, opened 2019-10-22
fix infrastructure. OMPVV.F90 modifies argument that is in read only memory

josemonsalve2:
See issue #11. This is its fix.


josemonsalve2:
I wonder if by using `intent(out)` in the parameter `conditionStr` we could also avoid this issue.

Issue #13, opened 2019-10-16
Verfication issue in test_target_teams_distribute _parallel_for_firstprivate.c


timcramer:
The target_teams_distribute_parallel_for_firstprivate.c test verification works on the assumption that the first private variable 'privatized' will be reset to the value 10 for each iteration. We think this assumption is only true when exactly one thread executes one interation of the outer loop. This might be the case a GPU. However, this means that the test fails on devices with fewer threads, so we think this might not correct and portable OpenMP code. We believe that the verification only works if we reset the value of 'privatized' to 10 between the two loops, which is not the case if we (for instance) only have one team with 8 threads.

‌

Do you agree or did we made a wrong assumption?

timcramer:
Old comments from bitbucket:


Jose Monsalve Diaz
Ah! I see what you mean! This is an interesting point. Let me explore further. I need to double check with the specs! Will update later today.


Jose Monsalve Diaz
We have studied this and you are right.

We will separate the test for the “first”ization and the privatization part.

We also will include a check to make sure there are multiple number of teams and number of threads

We could also attempt to use scheduling to encourage a thread to take over multiple iterations.


Tim Cramer
Ok, thanks for the confirmation. Separation of the privatization part sounds good. However, I dont see how a loop schedule can fix the test since even for that you need to make assumptions on the amount of threads in order to guarantee a specific value for the firstprivate variable (at least as long you dont use a coditional for the increment of the variable).

josemonsalve2:
@timcramer I have fixed the test and created a new one thanks to this conversation. I sent you an invitation to collaborate so I can add you as pull request reviewer. Thanks for your help!

josemonsalve2:
Tim, 

I was thinking about your problem with the number of threads. I think this is a configuration parameter we need to implement for the whole suite. What I was thinking is we push this test as it is, and then we create a new issue were we try to add new flags for the makefile `NUM_DEVICE_THREADS`, `NUM_DEVICE_TEAMS`, `NUM_HOST_THREADS`

There will be some corner cases we will need to cover, but I feel this is a larger change than this issue covers

timcramer:
Sounds good.

josemonsalve2:
This issue was solved by PR #21 and #19, The remaining feature will be discussed in issue #22

PR #19, opened 2019-10-22
Fixing test  test_target_teams_distribute_parallel_for_firstprivate.c

josemonsalve2:
This is a fix from issue #13 

As in a `parallel for` construct, privatization happens at the thread level. We were assuming that privatization occurred at the iteration level.

### DIFF
--- a/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_firstprivate.c
+++ b/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_firstprivate.c
@@ -23,7 +23,11 @@ int test_target_teams_distribute_parallel_for_firstprivate() {
   int b[SIZE_N];
   int c[SIZE_N];
   int d[SIZE_N];
-  int privatized = 10;
+  int firstized = 10;
+  int privatized;
+  int reported_num_teams[SIZE_N];
+  int reported_team_num[SIZE_N];
+  int reported_num_threads[SIZE_N];
   int errors = 0;
   int i, j, dev;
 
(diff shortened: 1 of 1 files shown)


tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_firstprivate.c
```diff
@@ -36,17 +40,37 @@ int test_target_teams_distribute_parallel_for_firstprivate() {
   }
 
   // check multiple sizes. 
-#pragma omp target data map(to: a[0:SIZE_N], b[0:SIZE_N], c[0:SIZE_N]) map(from: d[0:SIZE_N])
+#pragma omp target data map(to: a[0:SIZE_N], b[0:SIZE_N], c[0:SIZE_N])
   {
-#pragma omp target teams distribute parallel for firstprivate(privatized)
+#pragma omp target teams distribute parallel for firstprivate(privatized, firstized) num_teams(10) num_threads(256)
```
timcramer: Thanks for addressing this. Although I didnt test if it works now for us, I have two (minor) concerns:
(1) `privatized` should be `private` now instead of `firstprivate`, because its initial value is not used at all, is it?
(2) For "our" device (SX-Aurora) hard-coded `num_threads(256)` might be a problem, because we only have eight cores. Since the device is not running an OS, the context switch is done by involving the host, which is quite expensive.

josemonsalve2: I wanted to keep the privatized in the `firstprivate` clause as we are trying to test if this clause actually does the privatization. But for an application you are correct. 

This threads number is clearly an issue. Uhmm I wonder if we should try to expose these configurations to the outside world. IT would require us to change a lot of tests though. Interesting point to consider. 

My I ask what system are you running on? It's ok if you cannot say.

timcramer: Ok, I see. Then it makes sense to keep it in the `firstprivate` clause.

I think the `num_threads` configuration is not generic enough. However, I am fine with your suggestion in issue #13, of course.

josemonsalve2:
Change the variable name for `privatized` to `privatization_check_variable`

josemonsalve2:
change `num_threads(256)` to `num_threads(64)`

ams-cs:
What is the purpose of the num_threads clause here?
My specific concern is that AMD GCN only permits 16 threads per team, so this will not do what it says.
I came here having observed that the (unpatched) test passes on devices with 64 compute units, and fails on devices with only 60. I think it's better to test what the toolchain will actually do in practice (unless you plan to test a range of launch dimensions, of course).

josemonsalve2:
@ams-cs Since we are trying to check for privatization, we want to make sure that multiple teams and threads are created. However, this is still a problem. We are thinking of exposing these numbers to the user through an option in the Makefile, as seen in issue #22, so it can be modified by the user. However, for 4.5 it is still possible to create less teams than requested, and it does not seem to be an issue if you request more than available in hardware.

dmcdougall:
> it does not seem to be an issue if you request more than available in hardware.

That surely depends on the runtime, no?

PR #20, opened 2019-10-22
New test/test target teams distribute num teams

jhdavis8:
Adding C and Fortran versions of this test. They both pass all compilers, but the Fortran test throws the "less teams than requested" warning on both gfortran and xlf.


tests/4.5/target_teams_distribute/test_target_teams_distribute_num_teams.c
```diff
@@ -0,0 +1,76 @@
...
+    a[x] = 1;
+    b[x] = x;
+    c[x] = 0;
+  }
+
+#pragma omp target data map(tofrom: default_num_teams)
+  {
+#pragma omp target teams distribute map(tofrom: default_num_teams, c[0:1024]) map(to: a[0:1024], b[0:1024])
+    for (int x = 0; x < 1024; ++x) {
+      default_num_teams = omp_get_num_teams();
```
spophale: Have only one thread update this.

tests/4.5/target_teams_distribute/test_target_teams_distribute_num_teams.c
```diff
@@ -0,0 +1,76 @@
...
+
+  if (default_num_teams == 1) {
+    OMPVV_WARNING("Test operated with one team.  Testing num_teams clause cannot be done.");
+  } else if(default_num_teams <= 0) {
+    OMPVV_ERROR("Test returned num_teams <= 0.  Maybe omp_get_num_teams() is not returning correct number of teams.");
+    errors = 1;
+  } else {
+#pragma omp target data map(to: a[0:1024], b[0:1024]) map(from: c[0:1024], num_teams)
+    {
+#pragma omp target teams distribute num_teams(default_num_teams - 1) map(alloc: num_teams)
```
josemonsalve2: if default_num_teams == 1, then this sets num_teams(0)

tests/4.5/target_teams_distribute/test_target_teams_distribute_num_teams.c
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  if (default_num_teams == 1) {
+    OMPVV_WARNING("Test operated with one team.  Testing num_teams clause cannot be done.");
+  } else if(default_num_teams <= 0) {
+    OMPVV_ERROR("Test returned num_teams <= 0.  Maybe omp_get_num_teams() is not returning correct number of teams.");
+    errors = 1;
+  } else {
+#pragma omp target data map(to: a[0:1024], b[0:1024]) map(from: c[0:1024], num_teams)
+    {
```
josemonsalve2: Remove these

tests/4.5/target_teams_distribute/test_target_teams_distribute_num_teams.c
```diff
@@ -0,0 +1,76 @@
...
+#pragma omp target teams distribute num_teams(default_num_teams - 1) map(alloc: num_teams)
+      for (int x = 0; x < 1024; ++x) {
+	num_teams = omp_get_num_teams();
```
josemonsalve2: atomic, single team, or array of teams (now that we know how many there are). If array of teams, then you need to check all the values are the same

tests/4.5/target_teams_distribute/test_target_teams_distribute_num_teams.c
jhdavis8: Since this is wrapped in the else clause and the first if clause above it checks if default_num_teams == 1, I think this isn't possible to happen.

jhdavis8:
I have updated this PR to address comments from this week's meeting, except for the num_teams(0) concern which I replied to above, I think it is incorrect.

tests/4.5/target_teams_distribute/test_target_teams_distribute_num_teams.F90
```diff
@@ -0,0 +1,91 @@
...
+       END IF
+       c(x) = a(x) + b(x)
+    END DO
+
+    IF (default_num_teams .eq. 1) THEN
+       OMPVV_WARNING("Test operated with one team. Testing num_teams")
+       OMPVV_WARNING("cannot be done.")
+    ELSEIF (default_num_teams .le. 0) THEN
+       OMPVV_ERROR("omp_get_num_teams() returned result less than or")
+       OMPVV_ERROR("equal to 0.  Maybe omp_get_num_teams is not returning")
```
spophale: I don't think we should have the "Maybe..." part.

jhdavis8: Agreed, will remove.

PR #26, opened 2019-10-28
Data race and verification issue in test `mmm_target_parallel_for_sim…

simoatze:
Fix data race and verification mechanism.

1 - [Line 69](https://github.com/SOLLVE/sollve_vv/blob/master/tests/4.5/application_kernels/mmm_target_parallel_for_simd.c#L69) is missing a `private(k)` clause causing a data race on `k`. The `collapse(2)` will take care of the first two loop, but the loop on `k` will be executed by all threads which will be sharing the `k` index variable causing a race condition.

2 - [Line 85](https://github.com/SOLLVE/sollve_vv/blob/master/tests/4.5/application_kernels/mmm_target_parallel_for_simd.c#L85) is not actually incrementing correctly the number of errors (`error += error` will always be 0).


josemonsalve2:
This one is a bad bug! Will take care of it! Thanks for letting us know!

Issue #15, opened 2019-10-16
test_target_data_use_device_ptr not standard compliant?


ams-cs:
The test_target_data_use_device_ptr test-case places "use_device_ptr" before "map". This ordering fails with GCC 9, because the implementation requires the "map" first.

According to OpenMP 4.5: "A list item in a use_device_ptr clause must have a corresponding list item in the device data environment." (Section 2.10.1)

This does not say explicitly that the ordering within a directive matters, but does imply that the "use_device_ptr" should not occur until the variable has been mapped. This was the interpretation taken by the GCC implementation.

OpenMP 5.0 appears to agree with this interpretation because new language has been added to cover the case: "If one or more of the use_device_ptr or use_device_addr clauses and one or more map clauses are present on the same construct, the address conversions of use_device_addr and use_device_ptr clauses will occur as if performed after all variables are mapped according to those map clauses."

GCC 10 accepts the test-case, as is, but that's because the implementation has been updated to OpenMP 5.0 (in this regard).

I believe this test-case needs to have those clauses reversed to be truly considered 4.5 compliant.

josemonsalve2:
Hi Andrew, 

I think that even though 4.5 does not say that explicitly, the description of directives in OpenMP says the order of constructs should not matter. 

In 4.5, section 2.1 Directive format. Page 27 line 10
"The order in which clauses appear on directives is not significant."

It is a good point though. But I am not sure if the wording you are describing means that the order in which the clauses appear matters.

tob2:
_"I think that even though 4.5 does not say that explicitly, the description of directives in OpenMP says the order of constructs should not matter."_

One can consider this part of the problem: independent what the user wrote, the compiler can always map first, always use_device_ptr first, process in order, process in reverse order or does so depending of the phase of the moon.

What 4.5 explicitly states: the item must be be **in** "device data environment" – and the only way to do this reliably is by doing so in a **separate** directive.

Additionally, if it were obvious enough that it should work as is, OpenMP 5 had not a very lengthy clause added which explicitly spells out that the mapping happens first. – _I assume the issue only became apparent after OpenMP 4.5 was used; hence, no one thought about spelling it out in the spec before 4.5 was released._

Hence, something like the following should work – the dummy is needed as OpenMP 4.5 requires at least one mapping per "omp target data". (Also in this regard, OpenMP 5 is better.)

```
--- test_target_data_use_device_ptr.c.orig      2019-10-17 09:25:48.856929341 +0200
[...]

PR #28, opened 2019-10-30
Update test_target_data_use_device_ptr.c

spophale:
Separating mapping and use_device_pointer so that the array is in device data environment before using use_device_pointer.

### DIFF
--- a/tests/4.5/target_data/test_target_data_use_device_ptr.c
+++ b/tests/4.5/target_data/test_target_data_use_device_ptr.c
@@ -4,7 +4,7 @@
 
 // Test for OpenMP 4.5 target data with use_device_ptr
 int main() {
-  int errors = 0, len = 10000, isHost = 0;
+  int errors = 0, len = 10000, isHost = 0, map_dummy;
   int *array_device = NULL;
   int *array_host = NULL;
 
@@ -14,17 +14,19 @@ int main() {
   for (int i = 0; i < len; ++i)
     array_host[i] = i;
 
-#pragma omp target data map(tofrom: array_device[0:len])   \
-    use_device_ptr(array_device)
+#pragma omp target data map(tofrom: array_device[0:len])
   {
-#pragma omp target is_device_ptr(array_device) map(tofrom: array_host[0:len]) map(tofrom: isHost)
+#pragma omp target data map(map_dummy) use_device_ptr(array_device)
     {
-      isHost = omp_is_initial_device();
-      for (int i = 0; i < len; ++i) {
-        array_device[i] = i;
-        array_host[i] += array_device[i];
-      } 
-    } // end target
+#pragma omp target is_device_ptr(array_device) map(tofrom: array_host[0:len]) map(tofrom: isHost)
+      {
+        isHost = omp_is_initial_device();
+        for (int i = 0; i < len; ++i) {
+          array_device[i] = i;
+          array_host[i] += array_device[i];
+        } 
+      } // end target
+    } // end target data
   } // end target data
 
   // checking results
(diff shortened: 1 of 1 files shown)


spophale:
Addressing issue #15.

josemonsalve2:
This test should be converted to the header file

tests/4.5/target_data/test_target_data_use_device_ptr.c
```diff
@@ -1,45 +1,57 @@
...
-    } // end target
+#pragma omp target is_device_ptr(array_device) map(tofrom: array_host[0:N]) map(tofrom: isHost)
+      {
+        for (int i = 0; i < N; ++i) {
+          array_device[i] = i;
+          array_host[i] += array_device[i];
+        } 
+      } // end target
+    } // end target data
   } // end target data
```
josemonsalve2: I feel like it would be more interesting if we do:

```C
int *array_device_ptr;
#pragma omp target data map(tofrom: array_device[0:N])
#pragma omp target data map(map_dummy) use_device_ptr(array_device)
array_device_ptr = array_device;
#pragma omp target is_device_ptr(array_device_ptr) map(tofrom: array_host[0:N]) map(tofrom: isHost)
      {
        for (int i = 0; i < N; ++i) {
          array_device[i] = i;
          array_host[i] += array_device_ptr[i];
        } 
      } // end target
    } // end target data
  } // end target data

Comment (spophale) [reply tests/4.5/target_data/test_target_data_use_device_ptr.c]: On second thoughts, I think the assignment should happen before use_device_ptr is used. If we want to reply go in this direction.

**Comment by spophale** [comment]
I don't want to change the test as it is closer to the usage seen in applications.

File: tests/4.5/target_data/test_target_data_use_device_ptr.c [hunk 2]
```diff
@@ -1,45 +1,57 @@
...
-#pragma omp target is_device_ptr(array_device) map(tofrom: array_host[0:len]) map(tofrom: isHost)
+#pragma omp target data map(map_dummy) use_device_ptr(array_device)
     {
-      isHost = omp_is_initial_device();
-      for (int i = 0; i < len; ++i) {
-        array_device[i] = i;
-        array_host[i] += array_device[i];
-      } 
-    } // end target
+#pragma omp target is_device_ptr(array_device) map(tofrom: array_host[0:N]) map(tofrom: isHost)
```
josemonsalve2: IsHost is not needed

tests/4.5/target_data/test_target_data_use_device_ptr.c
```diff
@@ -1,45 +1,57 @@
...
 #include <stdio.h>
 #include <stdlib.h>
+#include "ompvv.h"
+
+#define N 1000
 
-// Test for OpenMP 4.5 target data with use_device_ptr
 int main() {
-  int errors = 0, len = 10000, isHost = 0;
+  int errors = 0, isHost = 0, map_dummy;
```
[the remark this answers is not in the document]
josemonsalve2: This still applies

tob2:
That patch has one big problem — `isHost` is no longer declared but still used for mapping (see commit ff84b5f):
```
--- a/tests/4.5/target_data/test_target_data_use_device_ptr.c
+++ b/tests/4.5/target_data/test_target_data_use_device_ptr.c
…
-  int errors = 0, len = 10000, isHost = 0;
+  int errors = 0, map_dummy;
…
+#pragma omp target is_device_ptr(array_device) map(tofrom: array_host[0:N]) map(tofrom: isHost)
```

tests/4.5/target_data/test_target_data_use_device_ptr.c
jhdavis8: @spophale I think this still needs to be removed, see @tob2 's comment?

Issue #16, opened 2019-10-16
"omp master" does not protect against data race across "omp teams"


jeffreysandoval:
I noticed an issue with an “omp master” construct appearing in an “omp teams” construct, in these tests:

```
test_target_teams_distribute.c
test_target_teams_distribute_collapse.c
test_target_teams_distribute_default_none.c
```

For example:

```
#pragma omp target data map(tofrom: a[0:ARRAY_SIZE], num_teams) map(to: b[0:ARRAY_SIZE])
  {
#pragma omp target teams distribute map(alloc: a[0:ARRAY_SIZE], b[0:ARRAY_SIZE], num_teams)
    for (int x = 0; x < ARRAY_SIZE; ++x) {
#pragma omp master
      {
        num_teams = omp_get_num_teams();
      }
      a[x] += b[x];
    }
  }
```

The “omp master” does not bind to the “omp teams” construct, and therefore doesn’t protect against data races across the threads in different teams.  To fix this code, I think you’d need a “lastprivate(num_teams)” clause, or replace “master” with “atomic write”, or just make the code conditional based on “omp_get_team_num() == 0” (or maybe even “x == 0”).

josemonsalve2:
Yes, you are right, this would be only in `parallel`. We will fix it.

spophale:
We could potentially have the num_teams as an array if you have a consistency check later.

PR #29, opened 2019-11-02
Fix data races from use of omp master in omp teams

jhdavis8:
This PR addresses issue #16. The following tests were affected:
- `test_target_teams_distribute.c`
- `test_target_teams_distribute_collapse[.c|.F90]`
- `test_target_teams_distribute_default_none.c`

Since `omp master` doesn't bind to `omp teams`, we can't use it to protect num_teams from data races. I resolved this issue by changing `test_target_teams_distribute` to use an array instead of a scalar, properly mapped, and for the rest of the tests using a conditional: `omp_get_team_num == 0`.


jhdavis8:
> LGTM. Are these the only tests that have this problem?

I just confirmed that these are the only tests to have this issue.

PR #33, opened 2019-11-15
 Adding the configuration for AOMP

josemonsalve2:
Adding the aomp as a compiler. 

aomp actually uses clang, but to differentiate it we use CC=aomp CXX=aomp. We then overwrite them with clang and clang++

AOMP infrastructure has some way to check for the available hardware, I took this from one of their makefile. Adding Ron Lieberman to check if this is all right.

### DIFF
--- a/sys/make/make.def
+++ b/sys/make/make.def
@@ -71,6 +71,35 @@ ifeq ($(CC), clang)
   CLINKFLAGS += -lm -O3 -fopenmp $(COFFLOADING)
 endif
 
+# AOMP compiler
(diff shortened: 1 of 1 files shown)


dmcdougall:
Oh cool!  Thanks for doing this :)

josemonsalve2:
I noticed there is a problem when combining with Modules, I will push a new
update soon

ronlieb:
are you creating a modulefile for aomp ? we do have a template in /usr/lib/aomp/bin/modulefile   

its lightly tested, so if you see issues in it let me know.

PR #39, opened 2019-11-28
New feature/adding git commit to log header

josemonsalve2:
I have added additional information to the log file to be able to track what was the commit that generated the log entry. 

The new header and footer for the log file looks like this:
```
*-*-*BEGIN*-*-*<action> <lang>=<compilation line>*-*-*<date>*-*-*<system>*-*-*<test path>*-*-*<compilerVersion>*-*-*<commit>*-*-*
*-*-*END*-*-*<action> <lang>=<compilation line>*-*-*<date>*-*-*<system>*-*-*<action result>*-*-*<additional Notes>*-*-*<commit>*-*-*
```

Then this information get extracted by the createSummary.py script, and added into the JSON format. 

This JSON format is then used by the report template in html. It checks if the commit exists in github, if so, it will create the appropriate links. The commit numbers are link to the github commit source files, while an extra link is added to the particular test.

PR #40, opened 2019-12-02
New feature/adding online report

josemonsalve2:
I have created the necessary infrastructure for online reports submissions from the makefile.

Here is how it works:

1. I have created a back end RESTful API that lives in our server which allows us to upload JSON files into the server. This API allows creating new reports that are named by a 9 digit hex tag (obtained through RSA-1 hashing). With this tag it is possible to upload, append and remove the result from the server.  It also allows to obtain a JSON file with results.
2. I took the already existing report html file and modified it to obtain the JSON file directly from the API, and hosted it in our website under https://crpl.cis.udel.edu/ompvvsollve/result_report/results.html . It obtains the tag as a get parameter of the URL:  https://crpl.cis.udel.edu/ompvvsollve/result_report/results.html?result_report=<TAG HERE>
3. I created a front end script to be distributed with our testsuite called `onlineConnection.py`. This uses all the API operations and allows us to submit reports directly. 
4. Then I modified the makefile to use this script with the make rule report_online. This will automatically upload the json file to our servers through the `onlineConnection.py` script, create a tag, and returns an URL that the user can use to visualize the report. 
5. I have also added the option to REPORT_ONLINE_APPEND and REPORT_ONLINE_TAG which allows the user to continue using the same report for multiple runs (or multiple systems). 

I recommend we push this PR after #39 . While it is not a direct dependency, the other PR allows tracking of the code (git commit) that generated the report, making it possible for users to do some sort of versioning of the tests. (this is not intended to replace our versioning mechanism)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ results_report
 ompvv/*.mod
 results.json
 results.csv
+recent_online_report_tags
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ RESULTS_JSON_OUTPUT_FILE=results.json
 RESULTS_CSV_OUTPUT_FILE=results.csv
 RESULTS_HTML_OUTPUT_FOLDER=results_report
 RESULTS_HTML_REPORT_TEMPLATE=$(CURDIR)/sys/results_template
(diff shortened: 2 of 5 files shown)


josemonsalve2:
I have updated our documentation to reflect this new feature. I think it's pretty cool to have it

PR #43, opened 2019-12-16
Addition of condensed summary to `make report_summary`

colleeneb:
This adds a few printed out lines at the end of `make report_summary` that look like:
```
 .cpp pass rate: 4/10 (40%) [ 2 build failures ]
 .c pass rate: 63/86 (73%) [ 1 build failures ]
 .f90 pass rate: 1/48 (2%) [ 43 build failures ]
```
(The pass rates will be different for different compilers, of course.)
 
The reason for adding a condensed summary like this is that typically I wanted to just copy and paste a few summary lines about how the compiler was doing into an email, not the whole output. I find this useful, at least.

### DIFF
--- a/sys/scripts/createSummary.py
+++ b/sys/scripts/createSummary.py
@@ -332,11 +332,20 @@ def main():
   elif args.format and args.format[0].lower() == 'summary':
     num_tests = len(results)
     failures = []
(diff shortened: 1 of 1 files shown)


spophale:
Thanks @colleeneb. It is very useful.

PR #46, opened 2019-12-17
New test/test target teams distribute depend

jhdavis8:
This is a extensive test of various combinations of depend clauses for target teams distribute. The general format is setting up a dependency between two distributes, and making sure that the dependency is preserved and no race conditions are created. Summit results:

**gcc**: Runtime error (libgomp: libgomp: GOMP_OFFLOAD_async_run unimplemented, because of nowait)
**gfortran**: Pass (with warnings for failure to show asychronous behavior)
**clang**: Pass
**xlc**: Pass
**xlf**: Pass (with warnings for failure to show asychronous behavior)


josemonsalve2:
Just as reminder, this test needs to be split in multiple tests.

jhdavis8:
> Just as reminder, this test needs to be split in multiple tests.

Just finished this, should be ready for further review now.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
```diff
@@ -0,0 +1,69 @@
...
+    c[x] = 0;
+  }
+
+#pragma omp target data map(to: a[0:N], b[0:N]) map(tofrom: c[0:N])
+  {
+#pragma omp target teams distribute nowait depend(out: c[0:N/2]) map(alloc: a[0:N], b[0:N], c[0:N])
+    for (int x = 0; x < N; ++x) {
+      c[x] += a[x] + b[x];
+    }
+#pragma omp target teams distribute nowait depend(out: c[N/2:N/2]) map(alloc: a[0:N], b[0:N], c[0:N])
```
spophale: c[n/2:n-1]

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
```diff
@@ -0,0 +1,69 @@
...
+
+  if (!(all_valid == 1 && race_found == 1)) {
+    OMPVV_WARNING("Test could not prove asynchronous operations of tasks dependent on disjoint array sections");
+  }
+  return 0;
+}
+
+int main() {
+  int errors = 0;
+  int isOffloading = 0;
```
spophale: use macro

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
jhdavis8: Fixed this to c[N/2:N] since the end index for array slices in C is exclusive.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
jhdavis8: Looked into this further, Kyle uses a variable here to store the results from TEST_AND_SET_OFFLOADING(isOffloading), so it does actually use the macro. He just wanted to use the value later for use in messages. But there are a bunch of warnings and errors which need to use the WARNING_IF and ERROR_IF macros which I'll fix.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
jhdavis8: Actually, according to spec section 2.4 the second number in an array section in C is the length, so c[N/2:N/2] is correct. Fortran's built-in array sections, however, use the inclusive upper bound, which is why it is different in the Fortran test.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
```diff
@@ -0,0 +1,66 @@
...
+      c[x] += a[x] + b[x];
+    }
+#pragma omp target teams distribute nowait depend(out: c[N/2:N/2]) map(alloc: a[0:N], b[0:N], c[0:N])
+    for (int x = 0; x < N; ++x) {
+      c[x] += 2 * (a[x] + b[x]);
+    }
+  }
+
+  for (int x = 0; x < N; ++x) {
+    if (!(c[x] == 3 * x || c[x] == 6 * x || c[x] == 9 * x)) {
```
spophale: Surer test if all of 'c' ==9 then the no wait was ignored and the sibling "tasks" were executed sequentially.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+#pragma omp target data map(to: a[0:N], b[0:N]) map(tofrom: c[0:N])
+  {
+#pragma omp target teams distribute nowait depend(out: c[0:N/2]) map(alloc: a[0:N], b[0:N], c[0:N])
+    for (int x = 0; x < N; ++x) {
+      c[x] += a[x] + b[x];
+    }
+#pragma omp target teams distribute nowait depend(out: c[N/2:N/2]) map(alloc: a[0:N], b[0:N], c[0:N])
+    for (int x = 0; x < N; ++x) {
```
spophale: Both try to update c --> race condition

jhdavis8: To address this comment and the above comment, I have updated this test and the `depend_in_in` test. Now the two tasks update different arrays, and the check for dependency is more strong and specific.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.F90
```diff
@@ -0,0 +1,80 @@
...
+    DO x = 1, N
+       IF ((c(x) .ne. 6*x) .and. (c(x) .ne. 9*x)) THEN
+          invalid_found = .TRUE.
+          errors = 1
+          OMPVV_ERROR("Found invalid values")
+          exit
+       END IF
+       IF (c(x) .eq. 6*x) THEN
+          race_found = .TRUE.
+       END IF
```
spophale: If c(x) is always 3*(a(x)+b(x)) -->9*x == No async behavior.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
```diff
@@ -0,0 +1,74 @@
...
+      race_found = 1;
+    }
+  }
+
+  if (invalid_found == 0 && race_found == 1) {
+    OMPVV_INFOMSG("Found independence between depend clauses on disjoint array sections.");
+  } else {
+    OMPVV_WARNING("Could not find independence between depend clauses on disjoint array sections.");
+  } 
+
```
spophale: Add test or warning for no async behavior.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_in_out.c
```diff
@@ -0,0 +1,92 @@
...
+    }
+#pragma omp target teams distribute nowait depend(out: c) map(alloc: b[0:N], c[0:N], d[0:N])
+    for (int x = 0; x < N; ++x) {
+      d[x] = c[x] + b[x];
+    }
+  }
+
+  for (int x = 0; x < N; ++x) {
+    if (d[x] != 5 * x) {
+      errors = 1;
```
spophale: use macro

jhdavis8:
Need to add taskwait to the ends of the target data regions, to ensure data is mapped back correctly.

jhdavis8:
Should be ready for review now

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.F90
```diff
@@ -0,0 +1,85 @@
+!===--- test_target_teams_distribute_depend_disjoint_section.F90------------===//
+!
+! OpenMP API Version 4.5 Nov 2015
+!
+! This test checks the out-out dependency of two tasks when the array
+! sections in the depend lists are disjoint (non-overlapping). If no
+! asynchronous behavior can be shown, then the test gives only a warning,
+! since this is still complaint. This test will always pass.
```
josemonsalve2: It seems like this test can fail if the value is neither one or the other.

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.c
```diff
@@ -0,0 +1,74 @@
+//===--- test_target_teams_distribute_depend_disjoint_section.c--------------===//
+//
+// OpenMP API Version 4.5 Nov 2015
+//
+// This test checks the out-out dependency of two tasks when the array
+// sections in the depend lists are disjoint (non-overlapping). If no
+// asynchronous behavior can be shown, then the test gives only a warning,
+// since this is still complaint. This test will always pass.
```
josemonsalve2: Shouldn't it fail if the results are not one or the other?

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_in_in.F90
```diff
@@ -0,0 +1,85 @@
+!===--- test_target_teams_distribute_depend_in_in.F90-----------------------===//
+!
+! OpenMP API Version 4.5 Nov 2015
+!
+! This test checks if two depend(in) tasks are independent by trying to
+! create a asynchronous behavior. If no asynchronicity can be shown, then
+! the testgives only a warning, since this is still complaint. This test
+! will always pass.
```
josemonsalve2: same comments as for out out

tests/4.5/target_teams_distribute/test_target_teams_distribute_depend_disjoint_section.F90
jhdavis8: Ah yes, I forgot to change the description after adding that error condition.

josemonsalve2: If you want it to fail you need OMPVV_TEST(_AND_SET), OMPVV_ERROR does not guarantee that the test will fail, it just displays an error.

jhdavis8: Should be fixed now

josemonsalve2:
On Fatnode: 
All GCC failures are the same. Not supported. We should file a bug saying to stop doing that
Checked 24 runs
Reported errors(8):
  test_target_teams_distribute_depend_array_section.c on gcc 8.2.0 (runtime) 
  test_target_teams_distribute_depend_disjoint_section.c on gcc 8.2.0 (runtime) 
  test_target_teams_distribute_depend_in_in.c on gcc 8.2.0 (runtime) 
  test_target_teams_distribute_depend_in_out.c on gcc 8.2.0 (runtime) 
  test_target_teams_distribute_depend_list.c on gcc 8.2.0 (runtime) 
  test_target_teams_distribute_depend_out_in.c on gcc 8.2.0 (runtime) 
  test_target_teams_distribute_depend_out_out.c on gcc 8.2.0 (runtime) 
  test_target_teams_distribute_depend_unused_data.c on gcc 8.2.0 (runtime) 

  .c pass rate: 8/16 (50%) [ 0 build failures ]
  .f90 pass rate: 8/8 (100%) [ 0 build failures ]

josemonsalve2:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93481

PR #51, opened 2020-01-08
Add 5.0 versions of target_teams_distribute_reduction tests in C

jhdavis8:
These are the same as the 4.5 C reduction tests, except they use specific mapping attributes rather than just `defaultmap(tofrom:scalar)`.


tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
```diff
@@ -0,0 +1,88 @@
...
+  int warned = 0;
+  srand(1);
+
+  for (int itr_count = 0; itr_count < 16; ++itr_count) {
+    for (int x = 0; x < N; ++x) {
+      if (rand() / (double)(RAND_MAX) < false_margin) {
+	a[x] = 1;
+      } else {
+	a[x] = 0;
+      }
```
josemonsalve2: These lines could be simply changed for `a[x] = rand() / (double)(RAND_MAX) < false_margin`

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
```diff
@@ -0,0 +1,88 @@
...
+      // come out to true under the 'and' operator, and the rest false. For the and
+      // operator, a test array that comes out true requires every entry to be true, 
+      // which is why this margin is so close to 100%.
+  double false_margin = pow(exp(1), log(.5)/N);
+  int errors = 0;
+  int num_teams[N];
+  int warned = 0;
+  srand(1);
+
+  for (int itr_count = 0; itr_count < 16; ++itr_count) {
```
josemonsalve2: I was wondering if we should change this for a while that checks that we will test for all true (and resulting in a `true` result. And for at least one false. 
Either way I am fine with this as well, it's just that there is a change we won't check for an all true result.

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
```diff
@@ -0,0 +1,88 @@
...
+    for (int x = 0; x < N; ++x) {
+      host_result = host_result && a[x];
+    }
+
+    if (itr_count == 0) {
+      for (int x = 1; x < N; ++x) {
+	if (num_teams[x-1] != num_teams[x]) {
+	  OMPVV_WARNING("Kernel reported differing numbers of teams.  Validity of testing of reduction clause cannot be guaranteed.");
+	  warned += 1;
+	}
```
josemonsalve2: Fix indent

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
```diff
@@ -0,0 +1,88 @@
...
+      }
+      if ((num_teams[0] == 1) && (warned == 0)) {
+	OMPVV_WARNING("Test operated with one team.  Reduction clause cannot be tested.");
```
josemonsalve2: indent

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
```diff
@@ -0,0 +1,88 @@
...
+      } else if ((num_teams[0] <= 0) && (warned == 0)) {
+	OMPVV_WARNING("Test reported invalid number of teams.  Validity of testing of reduction clause cannot be guaranteed.");
```
josemonsalve2: indent

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_bitand.c
```diff
@@ -0,0 +1,79 @@
...
+  double false_margin = pow(exp(1), log(.5)/N); // See the 'and' operator test for
+  int errors = 0;                               // an exaplantion of this math.
+  int num_teams[N];
+  int warned = 0;
+  srand(1);
+
+  for (int x = 0; x < N; ++x) {
+    for (int y = 0; y < 16; ++y) {
+      if (rand() / (double) RAND_MAX < false_margin) {
+	a[x] += 1 << y;
```
josemonsalve2: indent

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_bitand.c
```diff
@@ -0,0 +1,79 @@
...
+      }
+    }
+    num_teams[x] = -x;
+  }
```
josemonsalve2: Shall we try these many times like the one in and? anyway it's fine with me

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_bitor.c
```diff
@@ -0,0 +1,76 @@
...
+  double false_margin = pow(exp(1), log(.5)/N);
+  int errors = 0;
+  int num_teams[N];
+  int warned = 0;
+  srand(1);
+
+  for (int x = 0; x < N; ++x) {
+    for (int y = 0; y < 16; ++y) {
+      if (rand() / (double) RAND_MAX > false_margin) {
+	a[x] += (1 << y);
```
josemonsalve2: indent

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_bitor.c
```diff
@@ -0,0 +1,76 @@
...
+      }
+    }
+    num_teams[x] = -x;
+
+  unsigned int b = 0;
```
josemonsalve2: Same suggestion from before

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_multiply.c
```diff
@@ -0,0 +1,75 @@
...
+      num_teams[x + y] = omp_get_num_teams();
+    }
+    host_result = 1;
+    for (int y = 0; y < 16; ++y) {
+      host_result *= a[x + y];
+    }
+    OMPVV_TEST_AND_SET_VERBOSE(errors, host_result != result);
+    OMPVV_INFOMSG("Device result is %d and host result is %d.", result, host_result);
+    if (host_result != result) {
+      break;
```
josemonsalve2: This break will cause that the num_teams will not be filled completely, causing false errors later on in lines 53

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_or.c
```diff
@@ -0,0 +1,76 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+#define N 1024
+
+int test_or() {
+  char a[N];
+  double false_margin = pow(exp(1), log(.5)/N);   // See the 'and' operator test for
```
josemonsalve2: Shouldn't this be the other way. We want to have the chance to test all zeros to obtain a false result for or

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_or.c
```diff
@@ -0,0 +1,76 @@
...
+  int errors = 0;                                 // an explanation of this math.
+  int num_teams[N];
+  int warned = 0;
+  srand(1);
+
+  for (int x = 0; x < N; ++x) {
+    if (rand() / (double)(RAND_MAX) > false_margin) {
```
josemonsalve2: It might be a matter of changing the order of ths `>`

jhdavis8: `      if (rand() / (double)(RAND_MAX) < false_margin) {`
This is the code in the and test, it is already flipped. This name is the same, which is misleading so I will change it.

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
```diff
@@ -0,0 +1,88 @@
...
+#include "ompvv.h"
+#include <math.h>
+
+#define N 1024
+
+int test_and() {
+  char a[N];
+  char result;
+  char host_result;
+      // The below calculation is meant to ensure about half the test arrays will
```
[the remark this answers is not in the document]
jhdavis8: The description is slightly confusing. It's supposed to be the case that roughly half of all the different arrays we will test will have all true elements, while the other half will have a few falses. This will mean half of the test iterations (loop on line 33) will give a true result, and the other half false.

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
jhdavis8: I think this is a good idea, but should we be worried that the test might never terminate, since it relies on random behavior? Not sure.

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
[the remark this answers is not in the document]
jhdavis8: Edited the description.

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
jhdavis8: This is done for all tests where it was possible

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_multiply.c
jhdavis8: Fixed

tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_and.c
josemonsalve2: Yes, that's also a possibility, large threshold? `while (tested_all && num_attempts < THRESHOLD)`?

jhdavis8:
I decided to go with a repeated test generation approach for the `or` and `bitxor` reduction test to match `and`, and I also adjusted the condition for continuing to test, so it checks if we have already tested both the true and false possibilities before exiting (up until a macro-defined threshold). I took a similar approach for `bitand` and `bitor` in which I check to make sure there are at least some true and false bits in the test arrays before proceeding to test.

jhdavis8:
Summit results:
- gcc fail on the and operator (timeout)
- All other pass

PR #52, opened 2020-01-08
Add 4.5 and 5.0 Fortran versions of reduction tests

jhdavis8:
These are Fortran versions of the 4.5/5.0 reduction tests. As with C, the only difference between 4.5 and 5.0 versions is the use of specific mapping attributes in 5.0, instead of `defaultmap(tofrom:scalar)`.


jhdavis8:
I have carried over (and will continue to carry over) over any requested changes I make on the C versions to the Fortran versions.

jhdavis8:
Summit results:
- gfortran/xlf fail multiply test
- xlf fails subtract test
- All others pass

jhdavis8:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93510

Gfortran bug report

jhdavis8:
Response from on report pointed out an out-of-bounds issue with the loop termination conditions, which I fixed. I also found an issue on the subtract test with an extra defaultmap that was causing it to fail to compile. Now all are passing on all compilers on summit.

PR #54, opened 2020-01-09
New test/test target teams distribute dist schedule

jhdavis8:
This test checks the behavior of `dist_schedule(kind [, chunk_size])` on target teams distribute by using reported team num to check that the distribution matches the specified schedule. The clause is tested with and without a chunk size, and for kind `static`, which is the only method availible on this construct.

These tests pass all summit compilers.

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_dist_schedule.F90
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_dist_schedule.F90
@@ -25,17 +25,78 @@ PROGRAM main
 
   OMPVV_TEST_SHARED_ENVIRONMENT
 
-  OMPVV_TEST_VERBOSE(test_function() .ne. 0)
+  OMPVV_TEST_VERBOSE(test_dist_schedule() .ne. 0)
 
   OMPVV_REPORT_AND_RETURN()
 
 CONTAINS
   INTEGER FUNCTION test_dist_schedule() 
-    INTEGER:: errors
-    test_function = 1;
-    !$omp target map(from: errors)
-    test_function = 0;
-    !$omp end target
+    INTEGER:: errors, num_teams, x, counter
+    INTEGER,DIMENSION(N):: a, b
+
+    errors = 0
+    counter = -1
+    
+    DO x = 1, N
+       a(x) = -1
+       b(x) = -1
+    END DO
+
+    !$omp target teams distribute map(from: num_teams) map(tofrom: a(1:N))&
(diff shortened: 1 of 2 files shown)


tests/4.5/target_teams_distribute/test_target_teams_distribute_dist_schedule.F90
```diff
@@ -0,0 +1,102 @@
+!===------ test_target_teams_distribute_dist_schedule.F90 ----------------===//
+! 
+! OpenMP API Version 4.5 Nov 2015
+! 
+! This test checks that the dist_schedule clause (which must have kind 
+! static) correctly causes CHUNK_SIZE iterations to be split among the
+! number of teams the test is run with, in a round-robin faction in order
+! of the team number, when a chunk size is given. The test also confirms
+! that when no chunk size is given, that each team receives no more than
+! one chunk.
```
spophale: implementation defined

tests/4.5/target_teams_distribute/test_target_teams_distribute_dist_schedule.c
```diff
@@ -0,0 +1,98 @@
...
+    OMPVV_ERROR("omp_get_num_teams() returned a value less than 1.");
+    return 1;
+  }
+
+  int counter = -1;
+  for (int i = 0; i < N; ++i) {
+    if (i % CHUNK_SIZE == 0) {
+      counter = (counter + 1) % num_teams;
+    }
+    if (a[i] != counter) {
```
spophale: use macro

tests/4.5/target_teams_distribute/test_target_teams_distribute_dist_schedule.c
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  if (num_teams == 1) {
+    OMPVV_WARNING("Cannot test dist_schedule(static) because num_teams was 1.");
+  } else if (num_teams < 1) {
+    OMPVV_ERROR("omp_get_num_teams() returned a value less than 1.");
+    return 1;
+  }
+
+  for (int i = 1; i < N; ++i) {
```
spophale: use omp_get_schedule to find out implementation defined and then compare.

jhdavis8: It turns out that the OMP_SCHEDULE environment variable only pertains to loops with schedule type `runtime`. Since we can only use `static` on a `distribute` clause, this method will not work.

PR #56, opened 2020-01-09
Add Fortran version of default none test and minor clean up

jhdavis8:
This test is a Fortran version of the existing `target teams distribute default(none)` test. Also included is minor cleanup on the C version of the test. Both tests pass all summit compilers.


tests/4.5/target_teams_distribute/test_target_teams_distribute_default_none.F90
```diff
@@ -0,0 +1,87 @@
...
+
+PROGRAM test_target_teams_distribute_default_none
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+  OMPVV_TEST_OFFLOADING
+
+  OMPVV_WARNING("Test only uses default(none) clause and does not")
+  OMPVV_WARNING("guarantee that the default(none) is enforced.")
```
josemonsalve2: if you want these to be on a single line it is really annoying. This is fine with me, but I used to do:
```
! Declare this
         CHARACTER(len=300) :: infoMessage
! Use it as storage for the whole message
WRITE(infoMessage, *) " working on a shared memory &
           & environment. Test might be inconclusive"
           OMPVV_WARNING(infoMessage)
```

tests/4.5/target_teams_distribute/test_target_teams_distribute_default_none.F90
```diff
@@ -0,0 +1,87 @@
...
+       c(x) = 2*x
+       d(x) = 0
+    END DO
+
+    !$omp target data map(from: d(1:N)) map(to: a(1:N), b(1:N), c(1:N))
+    !$omp target teams distribute default(none) shared(a, b, c, d) &
+    !$omp& private(x, privatized)
+    DO x = 1, N
+       privatized = 0
+       DO y = 1, a(x) + b(x)
```
josemonsalve2: what about the datasharing attribute for y? 

should x be privitized if it is the iteration variable? This I am not sure

tests/4.5/target_teams_distribute/test_target_teams_distribute_default_none.F90
```diff
@@ -0,0 +1,87 @@
...
+    errors = 0
+    share = 0
+
+    DO x = 1, N
+       a(x) = x
+    END DO
+
+    !$omp target data map(to: a(1:N)) map(tofrom: share)
+    !$omp target teams distribute default(none) private(x) &
+    !$omp& shared(share, a) defaultmap(tofrom:scalar)
```
josemonsalve2: for example, here you don't privatize x

jhdavis8: I think there is a `private(x)` on the end of line 72

josemonsalve2: Yes, you are right haha! But still, It's a good question I don't know the answer for. If you use default(none) do you have to specify the iteration variable as well?

jhdavis8: Yes definitely- I'm not sure either....
Pg. 58: "If any of the loop iteration variables would otherwise be shared, they are implicitly made private on the loop construct." possibly relevant?

tests/4.5/target_teams_distribute/test_target_teams_distribute_default_none.F90
jhdavis8: I decided to go ahead an privatize it.

tests/4.5/target_teams_distribute/test_target_teams_distribute_default_none.F90
```diff
@@ -0,0 +1,90 @@
...
+  USE omp_lib
+  implicit none
+  OMPVV_TEST_OFFLOADING
+
+  CHARACTER(len=300) :: infoMessage
+  WRITE(infoMessage, *) "Test only uses the default(none) clause and &
+       & does not guarantee that the default(none) is enforced."
+  OMPVV_WARNING(infoMessage)
+
+  OMPVV_TEST_VERBOSE(default_none1() .ne. 0)
```
spophale: Add comment what the tests are checking.

jhdavis8: Added short comments to each

tests/4.5/target_teams_distribute/test_target_teams_distribute_default_none.F90
```diff
@@ -0,0 +1,90 @@
...
+#define N 1024
+
+PROGRAM test_target_teams_distribute_default_none
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+  OMPVV_TEST_OFFLOADING
+
+  CHARACTER(len=300) :: infoMessage
```
josemonsalve2: This definition must be located before the macro of line 21. Fortran definiitions must all happen at the beginning

josemonsalve2:
@jhdavis8 this code has an error in fortran.

tests/4.5/target_teams_distribute/test_target_teams_distribute_default_none.F90
jhdavis8: Fixed on master

PR #57, opened 2020-01-09
Add Fortran version of shared test and minor cleanup

jhdavis8:
This is the Fortran version of the existing test of `target teams distribute shared`, in addition to some minor cleanup of the C version. Both test are passing all summit compilers, *except* gfortran, which is failing the write test.

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_shared.F90
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_shared.F90
@@ -6,11 +6,12 @@
 ! tests in a few ways that the variable is shared between the teams.  In the
 ! first test, the atomic directive is used to indicate that all operations on
 ! the variable should be done atomically.  If the value is the correct value
-! at the end of the region, then all teams operated on the same variable.
+! at the end of the region, then all teams operated on the same variable, and
+! the variable was not privatized.
 !
 ! The second test, instead of writing to the variable, only reads from the
 ! variable.  This tests that the value of the shared variable has not been
-! initiallized improperly or privatized.
+! initialized improperly.
 !
 !//===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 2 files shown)


tests/4.5/target_teams_distribute/test_target_teams_distribute_shared.F90
```diff
@@ -0,0 +1,77 @@
...
+!
+! This test uses the shared clause on a target teams distribute directive and
+! tests in a few ways that the variable is shared between the teams.  In the
+! first test, the atomic directive is used to indicate that all operations on
+! the variable should be done atomically.  If the value is the correct value
+! at the end of the region, then all teams operated on the same variable.
+!
+! The second test, instead of writing to the variable, only reads from the
+! variable.  This tests that the value of the shared variable has not been
+! initiallized improperly or privatized.
```
spophale: Does not test privatized.

tests/4.5/target_teams_distribute/test_target_teams_distribute_shared.c
```diff
@@ -56,8 +50,7 @@ int main() {
   }
 
   OMPVV_TEST_AND_SET_VERBOSE(errors, (share != 0));
-  OMPVV_ERROR_IF(errors != prev_errors, "The value of share is = %d", share);
-  prev_errors = errors;
+  OMPVV_ERROR_IF(errors != 0, "The value of share is = %d", share);
```
spophale: Expected: Correct value  Error:

jhdavis8: Fixed this on both tests

jhdavis8:
At the moment this test is failing xlf and gfortran on the shared write part of the test (the first half).

jhdavis8:
Discovered a missing initialization was causing the test fails for fortran compilers. Fixed and the test is now passing all compilers.

josemonsalve2:
Ready to push then!

PR #58, opened 2020-01-10
Adding a test for collapse and SIMD

josemonsalve2:
This test uses the same methodology used for `test_target_teams_distribute_collapse` but for `target_simd`

### DIFF
--- a/tests/4.5/target_simd/test_target_simd_collapse.c
+++ b/tests/4.5/target_simd/test_target_simd_collapse.c
@@ -1,0 +1,113 @@
+//===--- test_target_simd_collapse.c------ testing collapse with SIMD--------===//
+//
+// OpenMP API Version 4.5 Nov 2015
+//
+// This test uses the collapse clause and tests that for simd loops out of the scope
+// of the collapsed loops are not parallelized.  This test tests using one and
+// two collapsed loops.
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+#define ARRAY_SIZE 128 //Array Size of 128 uses 16MB target memory and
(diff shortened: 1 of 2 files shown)


josemonsalve2:
There was an indentation problem in `tests/4.5/target_teams_distribute/test_target_teams_distribute_collapse.c` that we solved

josemonsalve2:
We tested on summit and it passes on all the compilers

tests/4.5/target_simd/test_target_simd_collapse.c
```diff
@@ -0,0 +1,113 @@
...
+int test_collapse1() {
+  OMPVV_INFOMSG("Testing for collapse(1)");
+  int * a_mem = malloc(ARRAY_SIZE*ARRAY_SIZE*sizeof(int));
+  int * b_mem = malloc(ARRAY_SIZE*(ARRAY_SIZE+1)*sizeof(int));
+  int (*a)[ARRAY_SIZE] = (int (*)[ARRAY_SIZE])a_mem;
+  int (*b)[ARRAY_SIZE + 1] = (int (*)[ARRAY_SIZE+1])b_mem;
+  int errors = 0;
+
+  // a and b array initialization
+  for (int x = 0; x < ARRAY_SIZE; ++x) {
```
spophale: why?

josemonsalve2: Is the question why do we initialize the array b?

jhdavis8: I think @spophale was wondering why separately initialize b[x][0].... looks like it's because b has size ARRAY_SIZE + 1.

josemonsalve2: Ah, It's because B[1:N+1] is all overwritten. Never read. But we could initialize it as well

PR #62, opened 2020-01-13
Add Fortran version of map test, fix format in C version

jhdavis8:
Fortran version of `test_target_teams_distribute_map`. Checks each mapping attribute individually, like in C version.

Passes on all summit compilers.

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_map.F90
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_map.F90
@@ -21,7 +21,10 @@ PROGRAM test_target_teams_distribute_map
 
   OMPVV_TEST_OFFLOADING
   OMPVV_TEST_SHARED_ENVIRONMENT
-  OMPVV_TEST_VERBOSE(test_map_to() + test_map_from() + test_map_alloc() + test_map_tofrom() .ne. 0)
+  OMPVV_TEST_VERBOSE(test_map_to() .ne. 0)
+  OMPVV_TEST_VERBOSE(test_map_from() .ne. 0)
+  OMPVV_TEST_VERBOSE(test_map_alloc() .ne. 0)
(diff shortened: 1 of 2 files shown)


tests/4.5/target_teams_distribute/test_target_teams_distribute_map.F90
```diff
@@ -0,0 +1,135 @@
...
+
+PROGRAM test_target_teams_distribute_map
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
+  OMPVV_TEST_OFFLOADING
+  OMPVV_TEST_SHARED_ENVIRONMENT
+  OMPVV_TEST_VERBOSE(test_map_to() + test_map_from() + test_map_alloc() + test_map_tofrom() .ne. 0)
```
josemonsalve2: It might be better to keep these separate, so the VERBOSE will mention which one was the one that actually failed?

jhdavis8: Should be separate now

jhdavis8:
I spent some time looking at Kyle's comments on the bitbucket PR for the C version of this test as well as the specs. 

I think it is good to leave the enter/exit data pragmas in as they are since this way each possible version clause of the map clause is tested independently with target teams distribute. Especially since we want to have seperate test of the `to` map type and the `from` map type, whereas if we remove the enter/exit data pragmas the to/from tests will have to be made identical, which would eliminate the ability to pinpoint one or the other as failing.

If a fail is actually caused by a bad enter or exit data, then the user should be able to identify that problem by checking if the compiler passes the dedicated enter/exit data tests (already in the suite), which have no target teams distribute regions.

Any thoughts?

jhdavis8:
Test compiles and passes gfortran on fatnode.

PR #64, opened 2020-01-13
Add Fortran test for device and update c test

jhdavis8:
This is a Fortran version of `test_target_teams_distribute_device`.

At present it passes xlf, but fails gfortran with the error: `libgomp: GOMP_target_enter_exit_data unhandled kind 0x05`


tests/4.5/target_teams_distribute/test_target_teams_distribute_device.F90
```diff
@@ -0,0 +1,95 @@
...
+    DO dev = 1, num_devices
+       !$omp target enter data map(to: a(1:N), b(1:N), c(1:N, dev:dev), &
+       !$omp& num_teams(dev:dev)) device(dev)
+    END DO
+
+    DO dev = 1, num_devices
+       !$omp target teams distribute map(alloc: a(1:N), b(1:N), &
+       !$omp& c(1:N, dev:dev), num_teams(dev:dev)) device(dev)
+       DO x = 1, N
+          num_teams(dev) = omp_get_num_teams()
```
josemonsalve2: data race

tests/4.5/target_teams_distribute/test_target_teams_distribute_device.F90
```diff
@@ -0,0 +1,97 @@
...
+    END DO
+
+    DO x = 1, num_devices
+       num_teams(x) = 0
+       errors(x) = 0
+    END DO
+
+    DO dev = 1, num_devices
+       !$omp target enter data map(to: a(1:N), b(1:N), c(1:N, dev:dev), &
+       !$omp& num_teams(dev:dev)) device(dev)
```
spophale: num_teams is not a valid clause for "target enter data"

jhdavis8: I think in this pragma `num_teams` is actually the name of an array being mapped to the device (hence the `dev:dev`) to record the number of teams used on a particular device. I can rename the array if it is too misleading.

jhdavis8:
I think this branch was made from an old new_test branch for a different test, rather than master, so I need to fix it before this is merged.

Issue #10, opened 2019-10-15
 Older tests not conforming to macro usage 


josemonsalve2:
All of the following tests do not conform to macro usage and should be updated:

- [x] linked_list.c
- [x] mmm_target.c
- [x] mmm_target_parallel_for_simd.c
- [x] test_target_is_device_ptr.c
- [x] test_target_map_global_arrays.c
- [x] test_target_map_local_array.c
- [x] test_target_map_pointer.c
- [x] test_target_map_pointer_default.c 
- [x] test_target_map_scalar_default.c
- [x] test_target_map_struct_default.c
- [x] test_target_private.c 
- [x] test_target_data_map.c 
- [x] test_target_data_map_array_sections.c (ready for review)
- [x] test_target_data_map_classes.cpp
- [x] test_target_data_map_devices.c
- [x] test_target_data_use_device_ptr.c
- [x] test_target_enter_data_classes.cpp (no longer exist, new versions have macros) 
- [x] test_target_enter_data_classes.cpp (no longer exist, new versions have macros)
- [x] test_target_enter_data_malloced_array.c 
- [x] test_target_enter_exit_data_classes.cpp
- [x] test_target_enter_exit_data_depend.c (ready for review)
- [x] test_target_enter_exit_data_devices.c (ready for review)
- [x] test_target_enter_exit_data_if.c (ready for review)
- [x] test_target_enter_exit_data_map_global_array.c
- [x] test_target_enter_exit_data_map_malloced_array.c
- [x] test_target_enter_exit_data_struct.c
- [x] test_target_teams_distribute_collapse.c (ready for review)
- [x] test_target_update_from.c 
- [x] test_target_update_if.c 
- [x] test_target_update_to.c

josemonsalve2:
@jhdavis8 let's try to get this one out of the way in the next week. I want to do this before tackling issue #36 
We can divide it up. Let's discuss it on Slack.

josemonsalve2:
Done with `application_kernels/linked_list.c`

josemonsalve2:
Done with `application_kernels/mmm_target.c`

josemonsalve2:
Done with `application_kernels/mmm_target_parallel_for_simd.c`

josemonsalve2:
Done with `target_data/test_target_data_map_devices.c` Also fixed `target/test_target_device.c` that was missing the descirption

josemonsalve2:
Done with `test_target_is_device_ptr.c`

[...]

PR #66, opened 2020-01-14
Fixing fix_test/test_target_map_global_arrays.c to use the headerfile

tmh97:
Fixing test_map_global_arrays.c to use the headerfile

### DIFF
--- a/tests/4.5/target/test_target_map_global_arrays.c
+++ b/tests/4.5/target/test_target_map_global_arrays.c
@@ -1,4 +1,4 @@
-//===------------------- test_target_map_global_arrays.c---------------------===//
+//===--- test_target_map_global_arrays.c ------------------------------------===//
 // 
 // OpenMP API Version 4.5 Nov 2015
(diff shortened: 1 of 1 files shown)


tmh97:
Tackling issue #10 , still needs to be tested on summitdev.

tests/4.5/target/test_target_map_global_arrays.c
```diff
@@ -1,33 +1,40 @@
-//===--- test_target_map_array.c - test mapping an array to the device ------===//
+//===------------------- test_target_map_global_arrays.c---------------------===//
```
jhdavis8: Usually we keep the file title "left-aligned" so it looks something like this:
`//===--- test_target_map_global_arrays.c ------------------------------------===//`
Three dashes and a space before the filename, and then a space and dashes after until it matches the length of the footer. Some tests omit the space after the filename... not sure how strictly we are enforcing this tho.

PR #67, opened 2020-01-15
Fixing  fix_test/test_target_map_local_array.c to use the headerfile

tmh97:
Fixing fix_test/test_target_map_local_array.c to use the headerfile


tmh97:
Part of tackling issue #10 , still needs to be tested on summitdev.

PR #68, opened 2020-01-21
Add array reduce tests to address issue #8

jhdavis8:
These are a pair of tests on array reductions that come from issue #8 . They are attempting to perform array operations in a set of nested loops on the device, reducing to array elements.

The two versions (v1 and v2) differ in how they handle the reduction. The first test uses the default OpenMP addition reduction, while the second declares a custom reduction operation, which should get the same results.

I converted these codes into tests by just performing the same math and the host and on the device and then checking the results, on some dummy generated data, since none was provided.

Right now, the only test that is passing is the first verion on gfortran. The second test will not compile on gfortran, complaining about a type conversion in the arrayadd module. I suspect that the target reduction might not be written correctly as I had to correct other errors as well, so it would be good to get some more opinions on that. On IBM compilers both tests fail, as mentioned in the issue.


tests/4.5/application_kernels/array_reduce_v1.F90
```diff
@@ -0,0 +1,117 @@
...
+  OMPVV_TEST_OFFLOADING
+  OMPVV_TEST_SHARED_ENVIRONMENT
+
+  OMPVV_TEST_VERBOSE(test_array_reduce() .ne. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+CONTAINS
+  INTEGER FUNCTION test_array_reduce()
+
+    ! xl : mpif90 -O3 -qstrict -qsmp=omp -qoffload array_reduce_v1.F90 -o array_reduce.x
```
josemonsalve2: what is this comment for?

tests/4.5/application_kernels/array_reduce_v1.F90
```diff
@@ -0,0 +1,117 @@
...
+          DO yp = 1, my
+             vx(x, z, yp) = yp
+          END DO
+       END DO
+    END DO
+
+    !$omp target data map(to:vx) map(from:rrs)
+    !$omp target
+    rrs(:,:) = 0.
+    !$omp end target
```
josemonsalve2: Why initialize it in the device?

tests/4.5/application_kernels/array_reduce_v2.F90
```diff
@@ -0,0 +1,156 @@
...
+  OMPVV_TEST_OFFLOADING
+  OMPVV_TEST_SHARED_ENVIRONMENT
+
+  OMPVV_TEST_VERBOSE(test_array_reduce() .ne. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+CONTAINS
+  INTEGER FUNCTION test_array_reduce()
+
+    ! xl : mpif90 -O3 -qstrict -qsmp=omp -qoffload array_reduce.F90 -o array_reduce.x
```
josemonsalve2: what is this for?

tests/4.5/application_kernels/array_reduce_v2.F90
```diff
@@ -0,0 +1,156 @@
...
+    !$omp target data map(to:vx) map(from:rrs)
+    !$omp target
+    rrs(:,:)%x=0.
+    !$omp end target
+
+    ! separation in x
+    !$omp target teams distribute parallel do collapse(4) default(none) &
+    !$omp private(yp, z, x, x1, ir, term1, term2, sqterm) shared(my, &
+    !$omp nz, nx, nxh, inc, norder, vx) &
+    !$omp reduction(+:rrs)
```
josemonsalve2: Really interesting test! 

I wonder if this is supported by 4.5

tests/4.5/application_kernels/array_reduce_v1.F90
jhdavis8: I believe this comment was added to record how they compiled the test. I will remove it.

tests/4.5/application_kernels/array_reduce_v1.F90
jhdavis8: I'm not sure actually, I'll change it

tests/4.5/application_kernels/array_reduce_v2.F90
```diff
@@ -0,0 +1,148 @@
...
+! compares the result. The reduction operator is a custom declared add
+! operator over reals. The reduction occurs within a four-level collapsed
+! loop.
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+#define N 768
+
+MODULE arrayadd_module
```
jhdavis8: Try to take out the reduction and see if the error stays

tests/4.5/application_kernels/array_reduce_v1.F90
```diff
@@ -0,0 +1,114 @@
+!===--- array_reduce_v1.F90 ------------------------------------------------===//
+!
+! OpenMP API Version 4.5 Nov 2015
+!
+! This test conducts an array reduction on both the host and device and
+! compares the result. The reduction operator is the default OpenMP add
+! operator over reals. The reduction occurs within a four-level collapsed
+! loop.
```
jhdavis8: Add comment describing where it came from

jhdavis8:
Update on the status of this test: only v1 is passing, and only on gfortran.

array_reduce_v1 xlf error: 1587-175 The underlying GPU runtime reported the following error "an illegal memory access was encountered". 

array_reduce_v2 gfortran error: /autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/4.5/application_kernels/array_reduce_v2.F90:50:6:

   USE arrayadd_module
      1
Error: Can't convert REAL(4) to TYPE(myreal) at (1)

array_reduce_v2 xlf error: "/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/4.5/application_kernels/array_reduce_v2.F90", 1516-044 (S) A conversion from type REAL*4 is not permitted.
"/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/4.5/application_kernels/array_reduce_v2.F90", 1516-064 (S) Operands to the + operator must be compatible with the intrinsic uses of the operator, or 
with a specific interface within an accessible defined operator generic interface.

jhdavis8:
Update: neither tests pass any compiler on Summit. It might be a good idea to contact the authors to see if the errors make sense, since I am unsure if I fully understand this code.

tests/4.5/application_kernels/array_reduce_v1.F90
```diff
@@ -0,0 +1,114 @@
...
+#define N 768
+
+PROGRAM array_reduce_v1
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
+  OMPVV_TEST_OFFLOADING
+  OMPVV_TEST_SHARED_ENVIRONMENT
```
spophale, line 22: remove.

jhdavis8:
I am closing this PR for now since it has been sitting on the docket for quite a while with no progress made. This test could be resurrected with some time, but it should try to use a simpler example/testing strategy.

PR #69, opened 2020-01-21
Fixing fix_test/test_target_map_pointer.c to use headerfile

tmh97:
Fixing test_target_map_pointer.c to use headerfile, as per issue #10.

### DIFF
--- a/tests/4.5/target/test_target_map_pointer.c
+++ b/tests/4.5/target/test_target_map_pointer.c
@@ -17,10 +17,12 @@
 #define N 1000
 
 int main() {
-  int compute_array[N], *p;
-  int sum = 0, result = 0;
-  int i, errors;
-
+  int compute_array[N];
+  int *p;
+  int sum = 0, result = 0, errors = 0;
+  int i;
+  
+  
   // Array initialization
   for (i = 0; i < N; i++)
     compute_array[i] = 0;
--- a/tests/4.5/target/test_target_map_pointer.c
+++ b/tests/4.5/target/test_target_map_pointer.c
@@ -1,36 +1,40 @@
-//===--- test_target_map_pointer.c - test target with map pointer p[:N] -----===//
+//===--- test_target_map_pointer.c ------------------------------------------===//
 // 
 // OpenMP API Version 4.5 Nov 2015
 //
 // This test check if it is possible to map an array and a pointer to that array,
 // and then access the array through the pointer. It is necessary
 // to specify the array size with [:N]. If offloading is used, the value of p[] 
-// is copied over the device. The array will be updated inside de omp target
+// is copied over the device. The array will be updated inside the omp target
 // region and compared afterwards
 //
 ////===----------------------------------------------------------------------===//
 
 #include <stdio.h>
 #include <omp.h>
+#include "ompvv.h"
 
 #define N 1000
 
 int main() {
   int compute_array[N], *p;
   int sum = 0, result = 0;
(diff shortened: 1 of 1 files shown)


josemonsalve2:
I suggest we improve this test by adding some additional checks that the pointer and array are the same. 

One option involves using a target data region which I am not sure it is a good idea to combine, but I feel it will add an extra element to the tests:


```C
int array[N], *p;
p = array;
#pragma omp target data map(tofrom:array)
#pragma omp target map(alloc: p[0:N])
for (...) 
   p[i] ++;

#pragma omp target map(alloc: array)
for (...)
   p[i] ++; 
```

The second option involves having the pointer being a function parameter

```C

void some_func(int * p) {
#pragma omp target map(alloc: p[0:N])
for (...) 
   p[i] ++;
}
main () {
  int array[N];
  p = array;
  #pragma omp target data map(tofrom:array)
   some_func(array);

  #pragma omp target map(alloc: array)
  for (...)
     p[i] ++; 
...
}
```

We should still keep the current target region that just uses the pointer.

josemonsalve2:
Just to consider, the test I suggested might be a `target data` test instead of a `target` test, so it could even be a second PR. Up to you

tests/4.5/target/test_target_map_pointer.c
```diff
@@ -1,36 +1,40 @@
...
 #pragma omp target data map(tofrom: compute_array) //To test default pointer behavior, array must be mapped before the pointer
-#pragma omp target map(to: p[:N]) map(tofrom: isHost)
+#pragma omp target map(to: p[:N]) 
   {
-    // Record where the computation was executed
-    isHost = omp_is_initial_device();
-  
     // Array modified through the pointer
     for (i = 0; i < N; i++)
       p[i] = i;
```
josemonsalve2: See my ticket comment, and figure if it is better to do those changes in `test/target_data` adding a new test, or if you want to modify this one. I think that creating the new test might be a better idea.

tmh97: I think to create a separate test is the best way to proceed.

josemonsalve2:
Do you want to add it to the same PR?

tmh97:
Would you recommend I create a new PR since it is an entirely new test and not a fix like the previous, or keep them together since they will be very similar?

josemonsalve2:
I recommend keeping them together

tmh97:
@jhdavis8 @josemonsalve2 Test passes fine with gcc, but when compiling with clang the tests fails on device with large error value (sum != result). Have either of you seen issues with how clang handles pointers, or any issue with the code that would cause to fail on clang?

josemonsalve2:
We should discuss this further. I feel like this should work, but I am also confused if there is a need to use "`use_device_pointer()`" in the `target data` region

tests/4.5/target/test_target_map_pointer.c
```diff
@@ -1,36 +1,40 @@
...
 #include <omp.h>
+#include "ompvv.h"
 
 #define N 1000
 
 int main() {
   int compute_array[N], *p;
   int sum = 0, result = 0;
-  int i, isHost = -1;
+  int i, errors;
```
spophale: Initialize errors to 0

tmh97: Thank you Swaroop!

tmh97:
New changes tested on Fatnode, tested and passed with gcc and clang

Issue #17, opened 2019-10-18
g++ compiler error when using `this` in update clause


dmcdougall:
[Right here](https://github.com/SOLLVE/sollve_vv/blob/fc544d0226c7d8cc5c2634e43df0f3488a2e541b/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_classes.cpp#L35) is where the error occurs.  I thought this was a reasonable thing to do until one of the Mentor Graphics folks pointed out to me that [you explicitly avoid using `this->` in a different test](https://github.com/SOLLVE/sollve_vv/blob/fc544d0226c7d8cc5c2634e43df0f3488a2e541b/tests/4.5/target/test_target_map_classes_default.cpp#L42-L46).

We should probably use the same workaround here.

josemonsalve2:
Yes, this was a discussion we had a while ago.  There is a PR right now on on of the C++ tests that is trying to fix that, so I will add this code to it as well. Would it be ok if I add you for code review? I think this is really complicated. XLC actually supports this, but the wording in 4.5 is clear that you are not supposed to have expressions in there (IIIRC that is the issue). 

Anyhow. I will leave this open until we solve it.

dmcdougall:
Happy to review code for you.

> but the wording in 4.5 is clear that you are not supposed to have expressions in there

Can you point me to the part of the spec that spells that out?

josemonsalve2:
The description of the map clause says that the map clause receives a list item:

Section 2.15.5.1, page 216, line 17
```
The syntax of the map clause is as follows:
map([ [map-type-modifier[,]] map-type : ] list)
```

And the definition of list item is 

Section 2.1, page 27, line 20
```
A list item is a variable or array section. An extended list item is a list item or a function name.
```

In 5.0 this was fixed. The new syntax of the map clause says locator-list:
section 2.19.7.1, page 315, line 18-19
```
The syntax of the map clause is as follows:
19 map([ [map-type-modifier[,] [map-type-modifier[,] ...] map-type : ] locator-list)
```

And they extended the description to include what a locator list is:

In Section 2.1, page 39, line 7-10
```
A locator-list consists of a comma-separated collection of one or more locator list items.
A list item is a variable or an array section. An extended list item is a list item or a function name. A locator list item is any lvalue expression, including variables, or an array section.
```

josemonsalve2:
@jhdavis8 you probably want to be aware of this as well

dmcdougall:
[...]

PR #71, opened 2020-01-23
Fix test/target enter exit data/test target enter exit data classes.cpp

josemonsalve2:
This is to tackle issue #17 as well as issue #10 . 

This test splits the old target `enter exit data` c++ test into two different tests for simple classes and a complex class that uses hierarchy. This test passes in gcc and xlc, but fails in clang


tests/4.5/target_enter_exit_data/test_target_enter_exit_data_classes_complex.cpp
```diff
@@ -0,0 +1,194 @@
+//===--test_targe_enter_exit_data_classes_complex.c - test inheritance class -===//
+// 
+// OpenMP API Version 4.5 Nov 2015
+// 
+//
+// This test uses target enter data and exit data to map multiple levels of 
+// inheritance in classes and templates. The target enter and exit data are used
+// in the constructor and destructor of all the levels of the hierarchy. 
+// It requires the use of helper_ variables since it
+// is not legal to use the "this" pointer (implicitely or explicitely) in 4.5.
```
jhdavis8: No 'e' in implicitly/explicitly

josemonsalve2: Thanks Josh!

josemonsalve2:
Yes, it is pretty much it, plus I added a test on the #target exit data to guarantee that the data is moved back during the destructor.  this is actually what's failing in Clang.

spophale:
What is the message on clang failure ?

josemonsalve2:
No particular error message, other than the values being incorrect:
See detailed information here. 
    https://crpl.cis.udel.edu/ompvvsollve/result_report/results.html?result_report=4b8a27d7d

PR #72, opened 2020-01-23
Add Fortran default private test

jhdavis8:
This is a test of the default(private) clause, which is only available in Fortran. It uses the same approach as the private test to check that a variable is given the private data-sharing attribute. Passes both Summit compilers.


jhdavis8:
Passes gfortran on fatnode

PR #73, opened 2020-01-23
Add Fortran default firstprivate test

jhdavis8:
This is a test for the default(firstprivate) clause, which is only available in Fortran. It uses a similar approach to the private test, except it gives an initial value to the privatized variable on the host.

Right now both gfortran and xlf are failing this test. I suspect this is a bug since moving the initialization to inside the target teams distribute region causes it to pass.

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90
@@ -32,7 +32,6 @@ CONTAINS
   INTEGER FUNCTION test_default_firstprivate()
     INTEGER,DIMENSION(N):: a, b, c, d
     INTEGER:: privatized, errors, x, y
-    CHARACTER(len=300):: errMsg
     errors = 0
 
     DO x = 1, N
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90
@@ -1,61 +1,118 @@
-!===--- test_target_teams_distribute_default_private.F90--------------------===//
+!===--- test_target_teams_distribute_firstprivate.F90-----------------------===//
 !
 ! OpenMP API Version 4.5 Nov 2015
 !
-! This test uses the default firstprivate clause on a target teams distribute
-! directive to indicate that a scalar variable should be by default made
-! firstprivate to each team executing the teams distribute region.  The test
-! then operates on the firstprivatized variable in such a way that would most
-! likely cause competiting operations if the variable is not privatized, and
-! such that the outcome depends on the initial host value of the variable. If
-! the computation completes without errors, we assume that the
-! firstprivatization occured.
+! This test uses the default firstprivate clause and tests it in two separate
+! parts. The test first tests the privatization of the firstprivatized scalars
+! and then separately tests the proper initialization of them separately
 !
-!===------------------------------------------------------------------------===//
+!//===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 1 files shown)


spophale:
@jhdavis8 did you investigate this further ?

tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90
```diff
@@ -0,0 +1,61 @@
...
+    !$omp& b, c, d)
+    DO x = 1, N
+       DO y = 1, a(x) + b(x)
+          privatized = privatized + 1
+       END DO
+       d(x) = c(x)*privatized
+    END DO
+
+    DO x = 1, N
+       OMPVV_TEST_AND_SET(errors, (d(x) .ne. (2*x)*(6 + x)))
```
spophale: check the math.
Looks like the final check does not account for the incremented value of privatized in the last iteration.

jhdavis8: I believe this math is correct since in the last iteration privatized will be incremented 1 + N times, setting the value to 5 + 1 + N = 6 + N in the last iteration before being multiplied by 2*N on line 52 to give 2*N*(6 + N) as the final value of d(x).

jhdavis8: Add num_teams()

spophale: Add num_teams, num_threads and thread_limit if you want to be sure. You need to catch the catch the case where only one team is created.

jhdavis8: I tried these suggestions and there were the results:
Adding a privatized = 5 after line 52: PASS
Adding a privatized = 5 after line 48: PASS
Adding num_teams(1): FAIL
Adding num_teams(1) and thread_limit(1): FAIL

This suggests to me that this tests exposes a bug, but I will continue to try other changes.

jhdavis8: Try with more than 1 team

jhdavis8: With some more investigation I think I've figured it out. Looks like `firstprivate` is functioning at the teams level, not the iteration level. This means that `privatized` is initialized only for the first iteration that each team gets. So the test isn't compliant as written, but by adding a `privatized = 5` at the end of each iteration the test passes all compilers and I believe is compliant and tests what we expect.

tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90
```diff
@@ -0,0 +1,61 @@
...
+
+    !$omp target teams distribute default(firstprivate) shared(a, &
+    !$omp& b, c, d)
+    DO x = 1, N
+       DO y = 1, a(x) + b(x)
+          privatized = privatized + 1
+       END DO
+       d(x) = c(x)*privatized
+       privatized = 5
+    END DO
```
jhdavis8: Change this to a separate first and private check like previous tests of firstprivate.

jhdavis8:
I've fully rewritten the test now to separately check first and private. Right now the test is passing xlf and failing gfortran on summit. On gfortran d(x) is reading as 0 for both parts of the test. This seems like a bug so I added the label to report a bug.

jhdavis8:
Bug report: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94290

Issue #22, opened 2019-10-24
Allow the user to select NUM_THREADS and NUM_TEAMS for the tests that use this


josemonsalve2:
**The problem**
Not all the devices would support large number of threads. In order to avoid being GPU specific we need to be flexible with respect to the requested number of threads. This is somehow different to the OpenMP Environment variables as we don't want to depend on the implementation, but rather pass the information from the test suite to OpenMP.

List of tests that this applies to:
- [x] test_target_firstprivate
- [x] test_target_private
- [x] test_target_teams_distribute_firstprivate
- [x] test_target_teams_distribute_num_teams
- [x] test_target_teams_distribute_private
- [x] test_target_teams_distribute_shared
- [x] test_target_teams_distribute_collapse
- [x] test_target_teams_distribute_thread_limit
- [x] test_target_teams_distribute_dist_schedule
- [x] test_target_teams_distribute_default_none
- [x] test_target_teams_distribute_default_shared
- [x] test_target_teams_distribute_default_private
- [x] test_target_teams_distribute_default_firstprivate
- [x] test_target_teams_distribute_parallel_for
- [x] test_target_teams_distribute_parallel_for_firstprivate
- [x] test_target_teams_distribute_parallel_for_if_no_modifier
- [x] test_target_teams_distribute_parallel_for_if_parallel_modifier
- [x] test_target_teams_distribute_parallel_for_if_target_modifier
- [x] test_target_teams_distribute_parallel_for_num_teams
- [x] test_target_teams_distribute_parallel_for_num_threads
- [x] test_target_teams_distribute_parallel_for_private
- [x] test_target_teams_distribute_parallel_for_schedule_private
- [x] test_teams
- [x] test_teams_distribute_default_none

**Describe the solution you'd like**
Create a new set of options to be included in the Makefile/system File:
* `OMPVV_NUM_THREADS_DEVICE`
* `OMPVV_NUM_TEAMS_DEVICE`
* `OMPVV_NUM_THREADS_HOST`

These variables are sent to the source codes through -D option in the compiler line, and used everywhere we have num_teams and num_threads. Some cases will have to require special attention. Specially those where we are trying to test for the num_threads and num_teams clauses themselves. 

We probably want to also have a set of default values. These values should live in ompvv.h header files. 

**Additional context**
This request was a consequence of the discussion occurred in Issue #15

jhdavis8:
List of tests that this applies to:
- [ ] test_target_firstprivate
[...]

PR #74, opened 2020-01-27
Start adding ability to specify number of teams, threads for issue #22

jhdavis8:
In this PR I make changes to the Makefile and ompvv.c/.F90 to allow the user to specify how many threads and teams the tests should run with, to avoid making the suite GPU-specific. This proposal was made in issue #22.

To test the feature I added the new macros, `OMPVV_NUM_THREADS_DEVICE`, `OMPVV_NUM_TEAMS_DEVICE`, and `OMPVV_NUM_THREADS_HOST`, to a couple of existing tests to make sure to functionality worked and demonstrate how it might be used. I implemented this in the way that seemed most simple and similar to what we already have, but I am open to changing it.

The issue page has a list of tests that need to be changed to use this feature.


Makefile
```diff
@@ -27,16 +27,31 @@ endif
...
+
+ifdef OMPVV_NUM_TEAMS_DEVICE
+	DTEAMS = -DOMPVV_NUM_TEAMS_DEVICE=${OMPVV_NUM_TEAMS_DEVICE}
+endif
+
+ifdef OMPVV_NUM_THREADS_HOST
+	HTHREADS = -DOMPVV_NUM_THREADS_HOST=${OMPVV_NUM_THREADS_HOST}
+endif
+
+###################################################
```
josemonsalve2: Shall we have these in /sys/make.def or here?

jhdavis8: Decided to move them to make.def

jhdavis8:
Tested on Fatnode

Issue #35, opened 2019-11-20
Another reduction tests


josemonsalve2:
Separated directives reduction tests. Taken from:
https://github.com/ROCm-Developer-Tools/aomp/issues/51

```
#include <iostream>

int main()
{
  int counts1 = 0;
  int counts2 = 0;
  #pragma omp target teams map(from:counts1)
  {
    int counts_team = 0;
    #pragma omp parallel
    {
      #pragma omp for
      for (int i=0; i<4; i++)
        #pragma omp atomic
        counts_team += 1;
    }
    counts1 = counts_team;
  }

  #pragma omp target teams map(from:counts2)
  {
    int counts_team = 0;
    #pragma omp parallel
    {
      #pragma omp for reduction(+:counts_team)
      for (int i=0; i<4; i++)
        counts_team += 1;
    }
    counts2 = counts_team;
  }

  if (counts1 != 4)
    std::cout << " wrong counts1 = " << counts1 << " should be 4!" << std::endl;
  if (counts2 != 4)
    std::cout << " wrong counts2 = " << counts2 << " should be 4!" << std::endl;
}
```

jhdavis8:
See pull request #77

jhdavis8:
See pull request #77

PR #77, opened 2020-01-28
New test/reduction separated directives.c (issue #35)

jhdavis8:
See issue #35. This is a test of the reduction clause with an and operator, using separated target teams parallel for directives. The test tries to perform the same calculation using both a reduction and atomic writes. I added a C and C++ version.

Both are passing the compilers on fatnode. Summit is down for maintenance today.


jhdavis8:
Fixed the data races by having only team 0 do the writes.

jhdavis8:
Right now this test is passing gnu but failing clang on fatnode. Need to submit a bug report.

spophale:
@jhdavis8 
Is there a reason not to merge the test ?

josemonsalve2:
It's waiting on bug report

jhdavis8:
Just tested on summit and clang and the tests are passing all compilers. Not sure what the cause of the errors I saw earlier was, so I will remove the label and merge this.

Issue #60, opened 2020-01-11
CORAL-2 Test suite integration


josemonsalve2:
Include these tests 

> Jose, Kyle, Sunita, and Oscar,
> 
> I recently read about work in the CORAL-2 procurement using the SOLLVE OpenMP test suite to demonstrate compiler functionality.  This is clearly very valuable work and having a comprehensive test suite is very important to us.
> 
> In the course of preparing for Sierra at LLNL, Aaron Black identified a large number of bugs in XLF’s OpenMP offload support.  We have released the collection of reproducers as open source.  It can be found here:  https://github.com/LLNL/FGPU/
> 
> Would you be interested in reviewing our collection to find examples that would be good additions to the SOLLVE suite?
> 
> Thanks,
> 
> Dave

Test suggestions to review from FGPU (tests that are out of scope are removed from the list):

- [x] array_property_issues
- [x] declare_target
- [ ] first_private_clause_not_working
- [x] map_derived_type
- [ ] omp_assumed_shape_array_failure
- [ ] omp_parallel
- [ ] omp_parallel_barrier
- [ ] omp_privatizer_bug
- [ ] target_enter_data_map_failure
- [x] target_map
- [ ] target_unstructured_data
- [x] teams
- [ ] https://github.com/LLNL/FGPU/tree/master/examples/array_slices

aaroncblack:
@josemonsalve2 - can you also add the omp map operation support on FORTRAN array slices at:
https://github.com/LLNL/FGPU/tree/master/examples/array_slices
Thanks

josemonsalve2:
Thanks @aaroncblack ! I have added the link to the list

spophale:
All positive feature tests have been added. Not covering negative testing at this point.

PR #78, opened 2020-01-28
Add declare target subroutine test from CORAL-2 (issue #60)

jhdavis8:
See issue #60. These tests are checking the behavior of the declare target clause, derived from tests on page 116 of the examples document. They are expected to fail of xlf, although I have not checked this as Summit is down today. 

Both involve calling subroutines on the device, which perform some simple math which is then checked. One checks calling a subroutine from a module, and the other a standalone subroutine.

Right now the module test is passing gfortran on fatnode, while the subroutine test won't compile, complaining about the lack of an explicit interface at the `declare target (test_declare)`, but based on the spec and examples document I believe this test is compliant and gfortran is wrong. If I add an interface statement, the test passes, but the examples document clearly states that the `declare target` is supposed to alleviate that requirement.


tests/4.5/application_kernels/declare_target_module.F90
```diff
@@ -0,0 +1,52 @@
+!===--- declare_target_module.F90 ------------------------------------------===//
+!
+! OpenMP API Version 4.5 Nov 2015
+!
+! This test uses the declare target construct within a subroutine, called
+! in a target region. It checks the result of a simple offloaded addition.
+! This test was provided by the LLNL FGPU repository, derived from the
+! OpenMP 4.5 examples document.
```
jhdavis8: Add link to original repo?

tests/4.5/application_kernels/declare_target_module.F90
```diff
@@ -0,0 +1,52 @@
...
+  implicit none
+  INTEGER:: x
+
+  OMPVV_TEST_OFFLOADING
+
+  OMPVV_TEST_SHARED_ENVIRONMENT
+
+  x = N
+
+  !$omp target if(x > THRESHOLD) map(tofrom: x)
```
josemonsalve2: Do we know why the THRESHOLD Thingy?

jhdavis8: This is from the original 4.5 examples test, https://www.openmp.org/wp-content/uploads/openmp-examples-4.5.0.pdf, pg  115-116 (122-123 on the pdf viewer). It isn't necessary for this test, however.

josemonsalve2: Let's leave it like this and blame it on the Examples

spophale:
declare_target_module.F90  compiles and runs correctly with xlc on Summit.
declare_target_subroutine.F90 does not compile. See output below:

** ompvv_lib   === End of Compilation 1 ===
"declare_target_subroutine.F90", line 26.24: 1514-480 (S) The entity test_declare that appears in the DECLARE TARGET directive must be declared in the same scoping unit.
** declare_target_subroutine   === End of Compilation 2 ===
** test_declare   === End of Compilation 3 ===
1501-511  Compilation failed for file declare_target_subroutine.F90.

jhdavis8:
Split into two PRs

tests/4.5/application_kernels/declare_target_module.F90
```diff
@@ -0,0 +1,53 @@
...
+
+#define N 1024
+
+MODULE module_declare
+CONTAINS
+  SUBROUTINE test_declare(x)
+    INTEGER:: x, m
+    !$omp declare target
+    x = x + 1
+  END SUBROUTINE test_declare
```
jhdavis8: Try putting the subroutine inside the program

tests/4.5/application_kernels/declare_target_subroutine.F90
```diff
@@ -0,0 +1,56 @@
...
+! Thanks to David Richards and Aaron Black for providing this test.
+!
+!===------------------------------------------------------------------------===//
+
+
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM declare_target_subroutine
```
jhdavis8: Add CONTAINS to move SUBROUTINE inside the program scope.

jhdavis8: Tried this, no change to results or the error message.

jhdavis8: Need to ping the authors about this test to see if anything has changed

jhdavis8: Also check with the examples document. Consider sending a note to vendors.

tob2: "test_declare" needs an explicit interface (or be in a use-assocated module or being a contained procedure) as it uses a deferred-shape array argument.

If you don't want to make it a contained procedure or put into a module, you could use something like:
```Fortran
INTERFACE
  SUBROUTINE test_declare(a, sum)
    !$omp declare target(test_declare)
    INTEGER,POINTER,INTENT(in) :: a(:)
    INTEGER,INTENT(inout):: sum
  END SUBROUTINE
END INTERFACE
```

jhdavis8: Thanks @tob2! This resolves it.

jhdavis8:
Test passes gfortran and xlf on summit.

spophale (approved):

PR #84, opened 2020-02-05
Test for the iterator modifier inside task depend on OMP 5.0

fabianmcg:
Test for the iterator modifier inside task depend clause on OMP 5.0.

### DIFF
--- a/tests/5.0/tasks/test_task_depend_iterator.cpp
+++ b/tests/5.0/tasks/test_task_depend_iterator.cpp
@@ -11,7 +11,6 @@
 
 #include <omp.h>
 #include <algorithm>
-#include <chrono>
 #include <cstdlib>
 #include <iostream>
 #include <thread>
@@ -22,7 +21,7 @@ int test_task_depend_iterator() {
   int ptr[] = {0, 4, 5, 6, 7, 8, 9, 10, 11}, cols[] = {1, 2, 3, 4, 5, 5, 6, 6, 7, 7, 8};
   std::vector<int> threadOrder;
   bool threadOrderError = false;
-#pragma omp parallel num_threads(4)
+#pragma omp parallel num_threads(OMPVV_NUM_THREADS_HOST)
   {
 #pragma omp single
     {
@@ -34,7 +33,6 @@ int test_task_depend_iterator() {
           {
             threadOrder.push_back(i);
           } // end critical section
-          std::this_thread::sleep_for(std::chrono::milliseconds(100));
         } // end task depend
       }
     } // end single
--- a/tests/5.0/tasks/test_task_depend_iterator.cpp
+++ b/tests/5.0/tasks/test_task_depend_iterator.cpp
@@ -20,42 +20,37 @@
 
 int test_task_depend_iterator() {
   int ptr[] = {0, 4, 5, 6, 7, 8, 9, 10, 11}, cols[] = {1, 2, 3, 4, 5, 5, 6, 6, 7, 7, 8};
-  std::vector<int> threadOrder, threadAssignment;
-  bool threadOrderError=false, threadAssignmentWarning=false;
+  std::vector<int> threadOrder;
+  bool threadOrderError = false;
 #pragma omp parallel num_threads(4)
   {
 #pragma omp single
     {
       for (int i = 0; i < 8; ++i) {
-        int pos = ptr[i], size = ptr[i+1] - ptr[i];
-#pragma omp task depend(iterator(it = 0 : size), in : ptr[cols[pos+it]]) depend(out : ptr[i])
+        int pos = ptr[i], size = ptr[i + 1] - ptr[i];
+#pragma omp task depend(iterator(it = 0 : size), in : ptr[cols[pos + it]]) depend(out : ptr[i])
         {
 #pragma omp critical
           {
             threadOrder.push_back(i);
-            threadAssignment.push_back(omp_get_thread_num());
           } // end critical section
           std::this_thread::sleep_for(std::chrono::milliseconds(100));
         } // end task depend
       }
     } // end single
   } // end parallel
-  std::sort(threadOrder.begin()+1,threadOrder.begin()+5);
-  std::sort(threadOrder.begin()+5,threadOrder.begin()+7);
-  std::sort(threadAssignment.begin()+1,threadAssignment.begin()+5);
+  std::sort(threadOrder.begin() + 1, threadOrder.begin() + 5);
+  std::sort(threadOrder.begin() + 5, threadOrder.begin() + 7);
   for(int i = 0; i < 8; ++i)
     threadOrderError = (threadOrder[i] != i) || threadOrderError;
-  for(int i = 0; i < 4; ++i)
-    threadAssignmentWarning = (threadAssignment[i+1] != i) || threadAssignmentWarning;
-  threadAssignmentWarning = (threadAssignment[5] == threadAssignment[6]) || threadAssignmentWarning;
-  OMPVV_WARNING_IF(threadAssignmentWarning, "The thread id assigned to each task may be erroneous.");
   OMPVV_ERROR_IF(threadOrderError, "The dependencies between tasks were not enforced in the correct order.");
   return threadOrderError ? 1 : 0;
 }
 
 
 int main() {
+  OMPVV_TEST_OFFLOADING;
   int errors = 0;
   OMPVV_TEST_AND_SET_VERBOSE(errors, test_task_depend_iterator());
-  OMPVV_RETURN(errors);
+  OMPVV_REPORT_AND_RETURN(errors);
 }
--- a/tests/5.0/tasks/test_task_depend_iterator.cpp
+++ b/tests/5.0/tasks/test_task_depend_iterator.cpp
@@ -1,0 +1,61 @@
+//===-- test_task_depend_iterator.cpp - test of the iterator clause inside the task depend clause ----===//
+// 
+// OpenMP API Version 5.0 Nov 2020
+// 
+// This file is a test for the iterator modifier when used with the task depend
+// clause. This modifier should create an iterator that expand into multiple values
+// inside the clause they appear. In this particular test case the iterator expands into
+// several values creating several dependencies at the same time.
+//
+//===-------------------------------------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <algorithm>
+#include <chrono>
+#include <cstdlib>
+#include <iostream>
+#include <thread>
+#include <vector>
+#include "ompvv.h"
+
+int test_task_depend_iterator() {
+  int ptr[] = {0, 4, 5, 6, 7, 8, 9, 10, 11}, cols[] = {1, 2, 3, 4, 5, 5, 6, 6, 7, 7, 8};
+  std::vector<int> threadOrder, threadAssignment;
+  bool threadOrderError=false, threadAssignmentWarning=false;
+#pragma omp parallel num_threads(4)
+  {
+#pragma omp single
+    {
+      for (int i = 0; i < 8; ++i) {
+        int pos = ptr[i], size = ptr[i+1] - ptr[i];
+#pragma omp task depend(iterator(it = 0 : size), in : ptr[cols[pos+it]]) depend(out : ptr[i])
+        {
+#pragma omp critical
+          {
+            threadOrder.push_back(i);
+            threadAssignment.push_back(omp_get_thread_num());
+          } // end critical section
+          std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        } // end task depend
+      }
+    } // end single
+  } // end parallel
+  std::sort(threadOrder.begin()+1,threadOrder.begin()+5);
+  std::sort(threadOrder.begin()+5,threadOrder.begin()+7);
+  std::sort(threadAssignment.begin()+1,threadAssignment.begin()+5);
+  for(int i = 0; i < 8; ++i)
+    threadOrderError = (threadOrder[i] != i) || threadOrderError;
+  for(int i = 0; i < 4; ++i)
+    threadAssignmentWarning = (threadAssignment[i+1] != i) || threadAssignmentWarning;
+  threadAssignmentWarning = (threadAssignment[5] == threadAssignment[6]) || threadAssignmentWarning;
+  OMPVV_WARNING_IF(threadAssignmentWarning, "The thread id assigned to each task may be erroneous.");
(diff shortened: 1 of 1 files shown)


tests/5.0/tasks/test_task_depend_iterator.cpp
```diff
@@ -0,0 +1,61 @@
...
+    threadAssignmentWarning = (threadAssignment[i+1] != i) || threadAssignmentWarning;
+  threadAssignmentWarning = (threadAssignment[5] == threadAssignment[6]) || threadAssignmentWarning;
+  OMPVV_WARNING_IF(threadAssignmentWarning, "The thread id assigned to each task may be erroneous.");
+  OMPVV_ERROR_IF(threadOrderError, "The dependencies between tasks were not enforced in the correct order.");
+  return threadOrderError ? 1 : 0;
+}
+
+
+int main() {
+  int errors = 0;
```
josemonsalve2: Add an OMPVV_TEST_OFFLOADING; macro here

tests/5.0/tasks/test_task_depend_iterator.cpp
```diff
@@ -0,0 +1,61 @@
...
+  OMPVV_TEST_AND_SET_VERBOSE(errors, test_task_depend_iterator());
+  OMPVV_RETURN(errors);
```
josemonsalve2: Also, use OMPVV_REPORT_AND_RETURN(errors);

tests/5.0/tasks/test_task_depend_iterator.cpp
josemonsalve2: BTW I know this is offloading check, but we are considering on changing the name of this macro to something else, But we want to keep a macro we can extend and affect all the tests at once, that's why I am asking. This is so we can do (for example) time measurements, or other initialization stuff.

tests/5.0/tasks/test_task_depend_iterator.cpp
```diff
@@ -0,0 +1,61 @@
...
+            threadAssignment.push_back(omp_get_thread_num());
+          } // end critical section
+          std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        } // end task depend
+      }
+    } // end single
+  } // end parallel
+  std::sort(threadOrder.begin()+1,threadOrder.begin()+5);
+  std::sort(threadOrder.begin()+5,threadOrder.begin()+7);
+  std::sort(threadAssignment.begin()+1,threadAssignment.begin()+5);
```
josemonsalve2: Coding standards: Spaces around + and - operators

tests/5.0/tasks/test_task_depend_iterator.cpp
```diff
@@ -0,0 +1,61 @@
...
+  for(int i = 0; i < 8; ++i)
+    threadOrderError = (threadOrder[i] != i) || threadOrderError;
+  for(int i = 0; i < 4; ++i)
```
josemonsalve2: There is a chance the test will run with less than threads since the specification does not enforce the number of threads to be equal to `num_threads()`, just never more than `num_threads()`. So I suggest you use `omp_get_num_threads()` to get the actual number of threads and use it for this for loop upper limit.

tests/5.0/tasks/test_task_depend_iterator.cpp
fabianmcg: I initially added OMPVV_REPORT_AND_RETURN , however it failed to compile on my computer due to the lack of support for offloading in my compiler. But I will added it.

tests/5.0/tasks/test_task_depend_iterator.cpp
fabianmcg: I know, that is why I set it as a warning because if the depend implementation is incorrect it may use only two threads even if launched with 4 threads. However since there is no specification guarantee may be remove that portion of the test?

josemonsalve2: I think it is fine to have this test, it's just a matter of adapting it to the actual number of threads during execution, and that should be it.

tests/5.0/tasks/test_task_depend_iterator.cpp
josemonsalve2: It may be because you are missing the OMPVV_TEST_OFFLOADING macro at the beginning. As I said, this is something we need to fix for sure.

tests/5.0/tasks/test_task_depend_iterator.cpp
fabianmcg: What I mean is that if the dependencies are not enforced correctly by the implementation then even when there are 4 active threads launched by OMP it may report only using two threads. I think since there are no specification guarantees it may be better to leave that part out.

tests/5.0/tasks/test_task_depend_iterator.cpp
spophale: Late to the party, but OMPVV_TEST_OFFLOADING might not be ideal for non-offloading cases as in the current setup it will report that tests are running on device when they are not.

spophale:
The iterator "it" is not declared. Also on Summit using xlc I see:
error: expected 'in', 'out' or 'inout' in OpenMP clause 'depend'
#pragma omp task depend(iterator(it = 0 : size), in : ptr[cols[pos + it]]) depend(out : ptr[i])
                        ^
1 error generated.
Error while processing test_task_depend_iterator.cpp

fabianmcg:
> The iterator "it" is not declared. Also on Summit using xlc I see:
> error: expected 'in', 'out' or 'inout' in OpenMP clause 'depend'
> #pragma omp task depend(iterator(it = 0 : size), in : ptr[cols[pos + it]]) depend(out : ptr[i])
> ^
> 1 error generated.
> Error while processing test_task_depend_iterator.cpp

I have seen this error in several compilers, it appears to be that most compilers (clang, pgi, gcc-8) haven't implemented the [iterator clause](https://www.openmp.org/spec-html/5.0/openmpsu22.html) for OpenMP 5.0

As the moment I only have been able to compile the code with g++ 9.2.1 on Debian testing.

spophale:
I was also wondering if it was an implementation issue or a problem with the program. Thanks for bringing it up.  

@fabianmcg we are changing some macros and adding new ones.
That might need more changes in the program.

tests/5.0/tasks/test_task_depend_iterator.cpp
```diff
@@ -0,0 +1,56 @@
...
+    {
+      for (int i = 0; i < 8; ++i) {
+        int pos = ptr[i], size = ptr[i + 1] - ptr[i];
+#pragma omp task depend(iterator(it = 0 : size), in : ptr[cols[pos + it]]) depend(out : ptr[i])
+        {
+#pragma omp critical
+          {
+            threadOrder.push_back(i);
+          } // end critical section
+          std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
josemonsalve2: we don't need this sleep here.

tests/5.0/tasks/test_task_depend_iterator.cpp
```diff
@@ -0,0 +1,56 @@
...
+  int ptr[] = {0, 4, 5, 6, 7, 8, 9, 10, 11}, cols[] = {1, 2, 3, 4, 5, 5, 6, 6, 7, 7, 8};
+  std::vector<int> threadOrder;
+  bool threadOrderError = false;
+#pragma omp parallel num_threads(4)
+  {
+#pragma omp single
+    {
+      for (int i = 0; i < 8; ++i) {
+        int pos = ptr[i], size = ptr[i + 1] - ptr[i];
+#pragma omp task depend(iterator(it = 0 : size), in : ptr[cols[pos + it]]) depend(out : ptr[i])
```
josemonsalve2: The sintax according to OpenMP 5.0 specification is:
```
iterator(iterators-definition)
```
where `iterators-definition` is:

```
iterator-specifier [,iterators-definition ]
```

And `iterator-specifier` is:
```
[ iterator-type ]  identifier=range-specification
```

And `iterator-type` is a type name.

It seems to me that you will need to define it with a typename. I will check it out, but it should be something like:

```
iterator(std::vector<int>::iterator  it = 0 : size)
```

In this case the type is: `std::vector<int>::iterator` , the identifier is `it` and the range is `0:size`

josemonsalve2: also in `ptr[cols[pos + it]]` the iterator must be dereference:
`ptr[cols[pos + *it ]]`

josemonsalve2: Uhmmm actually, the range has to also be an iterator object:


`range-specification` is of the form `begin:end[:step]`, where `begin` and `end` are **expressions for which their types can be converted to** `iterator-type` and `step` is an integral expression.
```

Comment (fabianmcg) [reply tests/5.0/tasks/test_task_depend_iterator.cpp]: As far I've seen and read, the iterator clause has no relation with C++ std iterators and OpenMP iterators are integer numbers defined by specification. If you check the OpenMP examples for 5.0, the clause comes without a type. That's why you don't have to de-reference the iterator.

File: tests/5.0/tasks/test_task_depend_iterator.cpp (hunk 5 above)
Comment (fabianmcg) [reply tests/5.0/tasks/test_task_depend_iterator.cpp]: That was to try to ensure the task execution on different threads. However since I removed the thread assignment check it can be removed.

File: tests/5.0/tasks/test_task_depend_iterator.cpp (hunk 6 above)
Comment (josemonsalve2) [reply tests/5.0/tasks/test_task_depend_iterator.cpp]: Actually, the iterator must be of type integer or pointer according to the restrictions:

```
The iterator-type must be an integral or pointer type.
```

I made this change and now it compiles and run with GCC 9.2

```
#pragma omp task depend(iterator(int it = 0 : size), in : ptr[cols[pos + it]]) depend(out : ptr[i])
```

Uhmm, I will bring this up with the spec people to see what they have to say

File: tests/5.0/tasks/test_task_depend_iterator.cpp [hunk 7]
```diff
@@ -0,0 +1,56 @@
...
+#include <iostream>
+#include <thread>
+#include <vector>
+#include "ompvv.h"
+
+int test_task_depend_iterator() {
+  int ptr[] = {0, 4, 5, 6, 7, 8, 9, 10, 11}, cols[] = {1, 2, 3, 4, 5, 5, 6, 6, 7, 7, 8};
+  std::vector<int> threadOrder;
+  bool threadOrderError = false;
+#pragma omp parallel num_threads(4)
```
josemonsalve2: @jhdavis8 shall we also have a `OMPVV_HOST_NUM_THREADS`
?

spophale: I don't see a reason why not. Although it is OK for tests to hard-code values as long as they are not unreasonable.

fabianmcg: The main reason of why it's hard coded to 4, is because the dependencies follows the graph:
```dot
digraph G {
   0 -> 1
   0 -> 2
   0 -> 3
   0 -> 4
   1 -> 5
   2 -> 5
   3 -> 6
   4 -> 6
   5 -> 7
   6 -> 7
}
````
Hence it could execute the spanning task from 0 all at the same time.

Comment (josemonsalve2) [reply tests/5.0/tasks/test_task_depend_iterator.cpp]: Having 4 threads does not guarantee this, and having more threads does not imply otherwise.
4 is fine, and not unreasonable, but we have received complains about hardcoding the number of threads and teams. This is more on the context of devices. 

For this particular case, it does not really make much of a difference, I was just wondering if it makes sense to think of a host OMPVV_NUM_THREADS for host for the sake of our infrastructure.

Comment (jhdavis8) [reply tests/5.0/tasks/test_task_depend_iterator.cpp:25]: @josemonsalve2 There is, as of recently, a OMPVV_NUM_THREADS_HOST macro available for these use. It was added with the device num macros.

File: tests/5.0/tasks/test_task_depend_iterator.cpp (hunk 5 above)
Comment (jhdavis8) [reply tests/5.0/tasks/test_task_depend_iterator.cpp]: Removed

File: tests/5.0/tasks/test_task_depend_iterator.cpp (hunk 7 above)
Comment (jhdavis8) [reply tests/5.0/tasks/test_task_depend_iterator.cpp]: Added the macro.

**Comment by jhdavis8** [comment]
Update: I have addressed any remaining changes on this test and I am waiting to check with Fabian that the errors I see were the same he saw. Also, @josemonsalve2, did you ever get a chance to talk to the spec people about the iterator type? For now I have left the type declaration off.

**Comment by jhdavis8** [comment]
Try with GCC 10?

**Comment by tob2** [comment]
> Try with GCC 10?

For what it is worth, with `g++-10 (SUSE Linux) 10.2.1 20200825 [revision c0746a1beb1ba073c7981eb09f55b3d993b32e5c]` and with the current GCC development version (GCC 11), it shows:

```
[OMPVV_INFO: test_task_depend_iterator.cpp:50] Test is running on host.
[OMPVV_INFO: test_task_depend_iterator.cpp:53] The value of errors is 0.
[OMPVV_RESULT: test_task_depend_iterator.cpp] Test passed on the host.
```

tests/5.0/tasks/test_task_depend_iterator.cpp
jhdavis8: From Deepak:

> The brackets indicate that the iterator-type is optional. The spec also says that if it is missing, the type is implicitly int for C/C++ and the default integer for Fortran.

> I think (iterator-type) would (be needed) if the expressions used in the range specification are pointer types

So this test should be valid.

jhdavis8 (approved):

jhdavis8:
> > Try with GCC 10?
> 
> For what it is worth, with `g++-10 (SUSE Linux) 10.2.1 20200825 [revision c0746a1beb1ba073c7981eb09f55b3d993b32e5c]` and with the current GCC development version (GCC 11), it shows:
> 
> ```
> [OMPVV_INFO: test_task_depend_iterator.cpp:50] Test is running on host.
> [OMPVV_INFO: test_task_depend_iterator.cpp:53] The value of errors is 0.
> [OMPVV_RESULT: test_task_depend_iterator.cpp] Test passed on the host.
> ```
> 
> Edit: And same result also with `g++-9 (SUSE Linux) 9.3.1 20200903 [revision 9790fa53b48f3a48e0f7a7ad65e2bbf3b206a7b0]`, however, that one is using the GCC 10 libgomp run-time library.

@tob2 Thanks for checking this. I believe this test is now ready to merge, once the approvals come in.

tmh97 (approved):
Good to go, passes on summit.

PR #85, opened 2020-02-07
Fix test/test enter exit data map global array.c to add test of delete attribute

jhdavis8:
Adds testing of delete clause, and addresses issue 10. The delete clause is tested as we proposed in a meeting last month: by mapping an array to the device, running a exit data map(delete) on the array, modifying the array on the host, and then running a target region with a map(to) on the array to see if the changes are seen by the device. The changes will not be seen by the device if the delete did not happen.

This updated test is passing all Summit compilers.

### DIFF
--- a/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_map_global_array.c
+++ b/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_map_global_array.c
@@ -1,45 +1,83 @@
+//===--- test_target_enter_exit_data_map_global_array.c ---------------------===//
+//
+// OpenMP API Version 4.5 Nov 2015
+//
+// This test is in two parts. First,
+//
+////===----------------------------------------------------------------------===//
+
 #include <stdio.h>
 #include <stdlib.h>
+#include "ompvv.h"
 #include <omp.h>
 
-// Test for OpenMP 4.5 target enter and target exit data with global arrays.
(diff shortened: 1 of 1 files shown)


tests/4.5/target_enter_exit_data/test_target_enter_exit_data_map_global_array.c
```diff
@@ -1,45 +1,88 @@
...
+    A[i] = N;
+  }
+
+#pragma omp target data map(tofrom: A) map(from: B)
+  {
+#pragma omp target exit data map(delete: A)
+    for (int i = 0; i < N; ++i) {
+      A[i] = 0;
+    }
+#pragma omp target map(to: A)   // if the delete does not work, this map will not happen.
```
tmh97: Is it not possible to map to A once the reference count becomes 0?

jhdavis8: Yup, unless you have the `always` clause. The map will only happen if the reference count is 1 or if the `always` clause is present.
Check out 4.5 spec page 217, lines 17-20.

tmh97: Thanks!

jhdavis8:
Test passed all summit compilers.

PR #86, opened 2020-02-09
New test/require unified shared memory

josemonsalve2:
This is a test for ` omp requires unified_shared_memory` It checks three different elements: 

1. Check if for a scalar value, we see the value reflected back and forth between host and device
2. Check if for an array allocated in the stack, we see the values reflected back and forth between host and device
3. Check if for an array allocated in the heap, we see the values reflected back and forth between host and device.

### DIFF
--- a/tests/5.0/requires/test_requires_unified_shared_memory.c
+++ b/tests/5.0/requires/test_requires_unified_shared_memory.c
@@ -74,6 +74,7 @@ int unified_shared_memory_stack() {
   for (int i = 0; i < N; i++) {
     OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 30);
     OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 30);
+    if (errors) break;
   }
   return errors;
 }
@@ -115,7 +116,11 @@ int unified_shared_memory_heap() {
   for (int i = 0; i < N; i++) {
     OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 30);
     OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 30);
+    if (errors) break;
   }
+
+  free(anArray);
+  free(anArrayCopy);
   return errors;
 }
 int main() {
--- a/tests/5.0/requires/test_requires_unified_shared_memory.c
+++ b/tests/5.0/requires/test_requires_unified_shared_memory.c
@@ -1,0 +1,131 @@
+//===---test_requires_unified_shared_memory.c -------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+//
+////===----------------------------------------------------------------------===//
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+#define N 1024
+
+#pragma omp requires unified_shared_memory
+
+int unified_shared_memory_scalar() {
+  OMPVV_INFOMSG("Unified shared memory testing - scalar value");
+  int errors = 0;
+  
+  int aVariable = 0;
+  // aVariable is only mapped to, with unified shared memory we should see the change reflected in the host.
+  // If no map is explicitely added, then scalars are firstprivate
+#pragma omp target map(to:aVariable)
+  {
+    aVariable += 10;
+  }
+
+  aVariable += 10;
+#pragma omp target map(alloc:aVariable)
+  {
+    aVariable += 10;
+  }
+
+  OMPVV_TEST_AND_SET_VERBOSE(errors, aVariable != 30);
+  return errors;
+}
+
+int unified_shared_memory_stack() {
+  OMPVV_INFOMSG("Unified shared memory testing - Array on stack");
+  int errors = 0;
+  
+  int anArray[N];
+  int anArrayCopy[N];
+
+  for (int i = 0; i < N; i++) {
+    anArray[i] = i;
+    anArrayCopy[i] = 0;
+  }
+  // If for any reason there is data mapping, we avoid the movement
+  // that may occur through the default mapping map(tofrom)
+#pragma omp target data map(alloc:anArray) 
+  {
+    // Modify again on the host
+    for (int i = 0; i < N; i++) {
+        anArray[i] += 10;
+    }
+#pragma omp target
+    {
+      for (int i = 0; i < N; i++) {
+        anArray[i] += 10;
+      }
+    }
+  }
+  // Modify again on the host
+  for (int i = 0; i < N; i++) {
+    anArray[i] += 10;
+  }
+#pragma omp target map(alloc:anArray)
+  {
+    for (int i = 0; i < N; i++) {
+      anArrayCopy[i] = anArray[i];
+    }
+  }
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 30);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 30);
+  }
+  return errors;
+}
+
+int unified_shared_memory_heap() {
+  OMPVV_INFOMSG("Unified shared memory testing - Array on heap");
+  int errors = 0;
+  
+  int *anArray;
+  int *anArrayCopy;
+
+  anArray = (int*)malloc(sizeof(int)*N);
+  anArrayCopy = (int*)malloc(sizeof(int)*N);
+
+  for (int i = 0; i < N; i++) {
+    anArray[i] = i;
+    anArrayCopy[i] = 0;
+  }
+  // Modiify in the device
+#pragma omp target
+  {
+    for (int i = 0; i < N; i++) {
+      anArray[i] += 10;
+    }
+  }
+  // Modify again on the host
+  for (int i = 0; i < N; i++) {
+    anArray[i] += 10;
+  }
+
+  // Get the value the device is seeing
+#pragma omp target 
+  {
+    for (int i = 0; i < N; i++) {
+      anArrayCopy[i] = anArray[i];
+    }
+  }
+
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 30);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 30);
+  }
+  return errors;
+}
+int main() {
+  int isOffloading;
+  OMPVV_TEST_AND_SET_OFFLOADING(isOffloading);
+  OMPVV_WARNING_IF(!isOffloading, "With no offloading, unified shared memory is guaranteed due to host execution");
+  int errors = 0;
+  OMPVV_TEST_AND_SET_VERBOSE(errors, unified_shared_memory_scalar());
+  OMPVV_TEST_AND_SET_VERBOSE(errors, unified_shared_memory_stack());
+  OMPVV_TEST_AND_SET_VERBOSE(errors, unified_shared_memory_heap());
+
+  OMPVV_REPORT_AND_RETURN(errors);
+}
--- a/tests/5.0/requires/test_requires_unified_shared_memory.c
+++ b/tests/5.0/requires/test_requires_unified_shared_memory.c
@@ -114,8 +114,8 @@ int unified_shared_memory_heap() {
   }
 
   for (int i = 0; i < N; i++) {
-    OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 30);
-    OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 30);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 20);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 20);
     if (errors) break;
   }
 
--- a/tests/5.0/requires/test_requires_unified_shared_memory.c
+++ b/tests/5.0/requires/test_requires_unified_shared_memory.c
@@ -1,7 +1,9 @@
 //===---test_requires_unified_shared_memory.c -------------------------------===//
 //
 // OpenMP API Version 5.0 Nov 2018
-//
+// 
+// This test just uses requries unified_shared_memory in a simple target region
+// and it is intended to show that this would not break anything in the compiler
 //
 ////===----------------------------------------------------------------------===//
 #include <omp.h>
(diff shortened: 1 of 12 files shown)


josemonsalve2:
Currently, this test does not pass in any compiler, 

I tried clang on UD system, but the current architecture it targets is `sm_30`, resulting in a compilation error: 
```
test_requires_unified_shared_memory.c:14:22: error: Target architecture does not support unified addressing
```

I tried it in summit with GCC 9.1.0 on summit, and `requires` has not been implemented. Resulting in 
```
test_requires_unified_shared_memory.c:14:22: sorry, unimplemented: 'unified_shared_memory' clause on 'requires' directive not supported yet
```

I Tried it on clang/9.1.0 as available in summit, and it resulted in the runtime error:
```
test_requires_unified_shared_memory.c.o: FAIL. exit code: 1
test_requires_unified_shared_memory.c.o:
[OMPVV_ERROR: test_requires_unified_shared_memory.c:34]  Condition aVariable != 30 failed
[OMPVV_ERROR: test_requires_unified_shared_memory.c:131]  Condition unified_shared_memory_scalar() failed
[OMPVV_ERROR: test_requires_unified_shared_memory.c:75]  Condition anArray[i] != i + 30 failed
[OMPVV_ERROR: test_requires_unified_shared_memory.c:76]  Condition anArrayCopy[i] != i + 30 failed
[OMPVV_ERROR: test_requires_unified_shared_memory.c:132]  Condition unified_shared_memory_stack() failed
Libomptarget fatal error 1: failure of target construct while offloading is mandatory
[OMPVV_INFO: test_requires_unified_shared_memory.c:128] Test is running on device.
[OMPVV_INFO: test_requires_unified_shared_memory.c:17] Unified shared memory testing - scalar value
[OMPVV_INFO: test_requires_unified_shared_memory.c:39] Unified shared memory testing - Array on stack
[OMPVV_INFO: test_requires_unified_shared_memory.c:83] Unified shared memory testing - Array on heap
```

And finally xlc, unimplemented as well:
```
test_requires_unified_shared_memory.c:14:13: error: expected an OpenMP directive
```

tests/5.0/requires/test_requires_unified_shared_memory.c
```diff
@@ -0,0 +1,136 @@
...
+#pragma omp target 
+  {
+    for (int i = 0; i < N; i++) {
+      anArrayCopy[i] = anArray[i];
+    }
+  }
+
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 30);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 30);
```
jdoerfert: Why 30? Don't you only add 10 twice?

josemonsalve2: Ah! Yes, copy and paste mistake

tests/5.0/requires/test_requires_unified_shared_memory.c
```diff
@@ -0,0 +1,136 @@
...
+  int aVariable = 0;
+  // aVariable is only mapped to, with unified shared memory we should see the change reflected in the host.
+  // If no map is explicitely added, then scalars are firstprivate
+#pragma omp target map(to:aVariable)
+  {
+    aVariable += 10;
+  }
+
+  aVariable += 10;
+#pragma omp target map(alloc:aVariable)
```
jdoerfert: I'm not sure about the alloc here. What do you think should happen? (map is not necessarily a noop for scalars)

tests/5.0/requires/test_requires_unified_shared_memory.c
jdoerfert: Does this test case, the heap one, fail as well with clang?

tests/5.0/requires/test_requires_unified_shared_memory.c
josemonsalve2: It serves two purposes: 
1. It avoids the default behavior of `firstprivate` scalar variables, otherwise, even with shared memory it will be privatized
2. If an implementation is mistakenly not using shared memory, I expect this to cause an allocation that changes the  memory location of aVariable

tests/5.0/requires/test_requires_unified_shared_memory.c
josemonsalve2: Yes, It fails with:
Libomptarget fatal error 1: failure of target construct while offloading is mandatory

josemonsalve2: This is summit 9.0.0. 

I was under the impression that this was same trunk, but this version output makes me doubt that. 
```
 clang --version
clang version 9.0.0 (https://github.com/clang-ykt/clang 5e1ea7eba87adbe43fed5746bcbba41d7cff1bf2) (https://github.com/clang-ykt/llvm fa0da850cf08eaffb84cb1480a27efa63c63fd14)
Target: powerpc64le-unknown-linux-gnu
Thread model: posix
InstalledDir: /sw/summit/llvm/9.0.0/9.0.0.patch001.cuda-10.1.168/bin
```

jdoerfert: Can we try trunk?

josemonsalve2: I would need to re-compile it on our UD system for sm_50 or higher I guess. It only seems to have sm_30, and it fails with 
```
Target architecture does not support unified addressing
```

josemonsalve2: Uhmm, on a second thought, we have K40's. They support managed memory, but they don't seem to have a page migration engine. i am not sure if you are relying on this for llvm ?

tests/5.0/requires/test_requires_unified_shared_memory.c
trws: This code is invalid anyway.  All `map` clauses are required to have correct semantics even in a `unified_shared_memory` program, as the system is allowed to use them for optimization, so it's an invalid test.

trws: Actually both of these target regions are invalid tests.  The main things you get from this you can actually test are:

* can I access a host heap variable from the device
* can I access a host stack variable from the device
* can I access a host global variable from the device and is it *the same* variable as seen on the device
* can I access a target allocated variable from the host (think omp_target_alloc)

It also includes the requirements of `unified_address` so using `is_device_ptr` should work on any host pointer, and a zero-length map of an unmapped list-item should result in the original address being passed to the device.

tests/5.0/requires/test_requires_unified_shared_memory.c
```diff
@@ -0,0 +1,136 @@
...
+  int *anArrayCopy;
+
+  anArray = (int*)malloc(sizeof(int)*N);
+  anArrayCopy = (int*)malloc(sizeof(int)*N);
+
+  for (int i = 0; i < N; i++) {
+    anArray[i] = i;
+    anArrayCopy[i] = 0;
+  }
+  // Modiify in the device
```
tmh97: Quick spelling swap Modiify -> Modify

tests/5.0/requires/test_requires_unified_shared_memory.c
josemonsalve2: @trws Thanks for your input, really valuable. 

From what you are saying, if we use a pointer and clause like this:

```
int *aPtr = (int*) malloc(sizeof(int)*N);
#pragma omp target 
   for (int i =0; i < N; i ++) {
      aPtr[i] = i;
   }
```
It should work because the default map is zero length array?

trws:
It should work certainly, as to why it’s more complicated than that unfortunately.

If you have unified_shared memory, the target of the pointer is accessible but is not *present*, meaning it was not mapped it is just directly accessible without need to map. The zero-length array section map will actually fail as a result. Under other circumstances that would mean the pointer would be set to NULL, but if you specify unified_address or unified_shared_memory, the original pointer value is preserved. Since the pointer has the original value and the target is accessible the example works.

We actually were just having a discussion about this on the spec, and are planning to clarify a few things in this area, but this is the behavior you should get.

On Feb 17, 2020, at 10:41 AM, Jose Manuel Monsalve Diaz <notifications@github.com> wrote:

﻿

@josemonsalve2 commented on this pull request.

________________________________

In tests/5.0/requires/test_requires_unified_shared_memory.c<https://github.com/SOLLVE/sollve_vv/pull/86#discussion_r380326038>:

> +#pragma omp requires unified_shared_memory
+
+int unified_shared_memory_scalar() {
+  OMPVV_INFOMSG("Unified shared memory testing - scalar value");
+  int errors = 0;
+
+  int aVariable = 0;
+  // aVariable is only mapped to, with unified shared memory we should see the change reflected in the host.
+  // If no map is explicitely added, then scalars are firstprivate
+#pragma omp target map(to:aVariable)
+  {
+    aVariable += 10;
+  }
+
+  aVariable += 10;
+#pragma omp target map(alloc:aVariable)


@trws<https://github.com/trws> Thanks for your input, really valuable.

From what you are saying, if we use a pointer and clause like this:

int *aPtr = (int*) malloc(sizeof(int)*N);
#pragma omp target
   for (int i =0; i < N; i ++) {
      aPtr[i] = i;
   }


It should work because the default map is zero length array?

josemonsalve2:
I have run these and here are the current tests that are failing on clang 

```
Reported errors(5):
  test_requires_unified_shared_memory_heap.c on clang version 9.0.0 CORAL (runtime) 
  test_requires_unified_shared_memory_omp_target_alloc.c on clang version 9.0.0 CORAL (runtime) 
  test_requires_unified_shared_memory_omp_target_alloc_is_device_ptr.c on clang version 9.0.0 CORAL (runtime) 
  test_requires_unified_shared_memory_stack.c on clang version 9.0.0 CORAL (runtime) 
  test_requires_unified_shared_memory_static.c on clang version 9.0.0 CORAL (runtime) 

  .c pass rate: 7/12 (58%) [ 0 build failures ]
```

More detailed information:
https://crpl.cis.udel.edu/ompvvsollve/result_report/results.html?result_report=a70f2ee46

josemonsalve2:
If `is_device_ptr` or `map(pointer)` is not specified, the test fails no matter what. 

Also all the test fail for device allocated memory accessed from the host (think `omp_target_alloc`)

tests/5.0/requires/test_requires_unified_shared_memory_heap.c
```diff
@@ -0,0 +1,71 @@
...
+      anArray[i] += 10;
+    }
+  }
+  // Modify again on the host
+  for (int i = 0; i < N; i++) {
+    anArray[i] += 10;
+  }
+
+  // Get the value the device is seeing
+#pragma omp target
```
jhdavis8: double check mapping rules for this. Is the reference count one when this target is encountered b/c anArray is already in the data env?

tests/5.0/requires/test_requires_unified_shared_memory_heap_is_device_ptr.c
```diff
@@ -0,0 +1,71 @@
...
+#pragma omp requires unified_shared_memory
+
+int unified_shared_memory_heap() {
+  OMPVV_INFOMSG("Unified shared memory testing - Array on heap");
+  int errors = 0;
+  
+  int *anArray;
+  int anArrayCopy[N];
+
+  anArray = (int*)malloc(sizeof(int)*N);
```
spophale: Don't we need to use omp device allocation ?

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc_is_device_ptr.c
```diff
@@ -0,0 +1,78 @@
...
+  int errors = 0;
+  
+  int *anArray = (int *) omp_target_alloc(sizeof(int)*N, omp_get_default_device());
+  int anArrayCopy[N];
+
+  // Initialization
+#pragma omp target is_device_ptr(anArray)
+  {
+    for (int i = 0; i < N; i++) {
+      anArray[i] = 0;
```
jhdavis8: Should this set anArray to i instead of 0, since the error checks for i + 20.

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc.c
```diff
@@ -0,0 +1,78 @@
...
+  // Get the value the device is seeing
+#pragma omp target
+  {
+    for (int i = 0; i < N; i++) {
+      anArrayCopy[i] = anArray[i];
+    }
+  }
+
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 20);
```
spophale: Either initialize to "i" or value should be 20.

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc_is_device_ptr.c
```diff
@@ -0,0 +1,78 @@
+//===---test_requires_unified_shared_memory.c -------------------------------===//
```
jhdavis8: Top of file names are incorrect for a few of these

tests/5.0/requires/test_requires_unified_shared_memory_heap_is_device_ptr.c
```diff
@@ -0,0 +1,71 @@
...
+
+  for (int i = 0; i < N; i++) {
+    anArray[i] = i;
+    anArrayCopy[i] = 0;
+  }
+  // Modify in the device
+#pragma omp target is_device_ptr(anArray)
```
jhdavis8: Need to check if this is permitted by spec? Can a host-malloced pointer be used in is_device_ptr?

trws: If it's the host device, then sure.  The requirement is that it's a valid host-handle to data for the device in question, for the host that should generally be a regular host pointer (exceptions might be opencl or sycl backends with certain properties)

tests/5.0/requires/test_requires_unified_shared_memory_heap_is_device_ptr.c
spophale: So the answer is yes, if the unified memory requirement is satisfied and working correctly the heap allocation should be accessible as well as "see" the same value.

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc.c
josemonsalve2: Done

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc.c
```diff
@@ -0,0 +1,78 @@
...
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 20);
+    if (errors) break;
+  }
+
+  free(anArray);
```
spophale, line 66: Needs to be freed using "omp_target_free", else the behavior is undefined.

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc.c
```diff
@@ -0,0 +1,78 @@
...
+
+int unified_shared_memory_target_alloc() {
+  OMPVV_INFOMSG("Unified shared memory testing - Memory allocated in the device");
+  int errors = 0;
+  
+  int *anArray = (int *) omp_target_alloc(sizeof(int)*N, omp_get_default_device());
+  int anArrayCopy[N];
+
+  // Initialization
+#pragma omp target
```
spophale, line 29: I think we need to use is_device_ptr to use the existing device pointer.

spophale, line 29: Also, depending on the need, we may want to use the device number returned by "omp_get_default_device", using the device clause on the target.

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc.c
```diff
@@ -0,0 +1,78 @@
...
+    for (int i = 0; i < N; i++) {
+      anArray[i] = i;
+    }
+  }
+  for (int i = 0; i < N; i++) {
+    anArrayCopy[i] = 0;
+  }
+
+  // Modify in the device
+#pragma omp target
```
spophale, line 40: See previous comments on is_device_ptr and device clause.

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc_is_device_ptr.c
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArray[i] != i + 20);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, anArrayCopy[i] != i + 20);
+    if (errors) break;
+  }
+
+  free(anArray);
```
spophale, line 66: Needs to be freed using "omp_target_free", else the behavior is undefined.

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc.c
josemonsalve2: fixed

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc.c
josemonsalve2: After reviewing during meeting, this should be correct according to the specs

josemonsalve2: After reviewing during meeting, this should be correct according to the specs.

if there is a compiler that changes the default device, it is a bug

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc.c
josemonsalve2: After reviewing during meeting, this should be correct according to the specs

tests/5.0/requires/test_requires_unified_shared_memory_omp_target_alloc_is_device_ptr.c
josemonsalve2: Sollved (Pun intended)

Issue #87, opened 2020-02-11
errors variable is uninitialized in reduction_sub test


shivaramaarao:
source file :
tests/4.5/target_teams_distribute/test_target_teams_distribute_reduction_sub.F90

At line no: 30
 30     INTEGER:: x, y, errors, host_result, result
 31
the value errors is not initialized to 0

this would result in random pass/fail of the test
@line 50:
OMPVV_TEST_AND_SET_VERBOSE(errors, result .ne. host_result)
the above line sets errors as
errors = errors + ...

spophale:
There are two "error" variables with different scopes. I believe the one on line 30 can be safely removed as there exists another on line 19 which is correctly initialized on 20.

spophale:
Thanks @shivaramaarao. The fix is in #88 .

PR #88, opened 2020-02-11
Bug fix for test_target_teams_distribute_reduction_sub.F90

spophale:
Addressing Issue #87 reported. Needs to be tested.

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_reduction_sub.F90
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_reduction_sub.F90
@@ -27,7 +27,7 @@ PROGRAM test_target_teams_distribute_device
 CONTAINS
   INTEGER FUNCTION test_sub()
     INTEGER,DIMENSION(N):: a
-    INTEGER:: x, y, errors, host_result, result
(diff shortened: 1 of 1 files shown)


jhdavis8:
I don't think we can remove this error var since it is referenced on lines 50 and 53 to track errors in the test_sub function

jhdavis8:
@jhdavis8 ah nevermind, I see it is coming from the outer scope.

spophale:
Tested with xlf on summit.

josemonsalve2:
Errors is only needed within the function, I feel it might be better to remove the outer one and leave the inner one, initialized

jhdavis8:
> Errors is only needed within the function, I feel it might be better to remove the outer one and leave the inner one, initialized

I agree with this

spophale:
There needs to be consistency across tests. Does not matter where.

jhdavis8:
> There needs to be consistency across tests. Does not matter where.

Looking at other Fortran tests, many of them have a useless/redundant `INTEGER:: errors` in the outer context that could be removed. I think the best solution is to make a PR (I can do it) to remove all of these since it is clearer if the errors var is defined in the context is used in.

PR #89, opened 2020-02-11
Fix Macro Usage and Formatting Issues in fix_test/test_target_enter_e…

tmh97:
Code style changes, removing unneeded braces and print statement surrounding 2 pragma statements, and macro usage changes. As per issue #10


tmh97:
Tests functioned properly with gcc and clang

Issue #79, opened 2020-01-28
Changing behavior of OMPVV_REPORT  and OMPVV_REPORT_AND_RETURN


spophale:
**Is your feature request related to a problem? Please describe.**
Both check for OpenMP offloading support. Some tests don't do offloading and still get reported as executed on device.

**Describe the solution you'd like**
OMPVV_TEST_OFFLOADING should test offloading support and set global state. If a test does not offload then OMPVV_TEST_OFFLOADING should not be used.

Reporting macros should only look at the flag to decide where the test was executed.

jhdavis8:
Is this issue ready to close? Fixed the behavior with PR #91

PR #91, opened 2020-02-12
Change the header file to address issue #79

jhdavis8:
This is for issue #79. Some tests do not do anything on the device, but the existing header file reports that they "ran on the device" since it directly checks if device offloading is availible. Since this check is redundant when the test also calls OMPVV_TEST_OFFLOADING, in this update to the header file, OMPVV_REPORT now checks a global state variable to see if OMPVV_TEST_OFFLOADING was called. If it was not called, the test reports that it ran on the host at completion.

I am open to changing the message for this case since it would be inaccurate if the test *does* do things on the device but forgets to call OMPVV_TEST_OFFLOADING, perhaps not mentioning device/host at all since it would technically be unknown.

### DIFF
--- a/ompvv/ompvv.h
+++ b/ompvv/ompvv.h
@@ -80,7 +80,7 @@ _Pragma("omp target map (from: _ompvv_isOffloadingOn)") \
 #define OMPVV_REPORT(err) { \
   OMPVV_INFOMSG("The value of " #err " is %d.", err); \
   if (_ompvv_isOffloadingOn == -1) \
-    printf("[OMPVV_RESULT: %s] Test %s on the host.\n", __FILENAME__, (err == 0)? "passed":"failed"); \
+    printf("[OMPVV_RESULT: %s] Test %s.\n", __FILENAME__, (err == 0)? "passed":"failed"); \
   else \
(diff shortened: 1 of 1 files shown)


ompvv/ompvv.h
```diff
@@ -77,9 +78,11 @@ _Pragma("omp target map (from: _ompvv_isOffloadingOn)") \
 }
 // Macro for reporting results
 #define OMPVV_REPORT(err) { \
-  OMPVV_TEST_OFFLOADING_PROBE \
   OMPVV_INFOMSG("The value of " #err " is %d.", err); \
-  printf("[OMPVV_RESULT: %s] Test %s on the %s.\n", __FILENAME__, (err == 0)? "passed":"failed", (_ompvv_isOffloadingOn)? "device" : "host"); \
+  if (_ompvv_isOffloadingOn == -1) \
+    printf("[OMPVV_RESULT: %s] Test %s on the host.\n", __FILENAME__, (err == 0)? "passed":"failed"); \
```
jhdavis8: Change message to just say test passed/failed.

josemonsalve2: remove "on the host"

spophale:
@tmh97 @josemonsalve2 
Is there a reason to block this ?

josemonsalve2:
@jhdavis8 remember to add one for `OMPVV_INIT` or something?

PR #92, opened 2020-02-12
Fix test/test enter exit data map malloced array.c to add delete testing and fix issue #10

jhdavis8:
This is to address issue #10 and add testing of the delete clause. The approach is exactly the same as used for pul request #85. 

The test passes all summit compilers.


jhdavis8:
Update: I added a `[:N]` after every array name in the map clauses and now it is passing. Not sure about any spec reasons for this.

PR #93, opened 2020-02-13
Modifying test_target_map_pointer.default.c to use headerfile, as per…

tmh97:
This test checks that the default behavior of map clause is tofrom when no map-type is specified for pointer p.

Fixed macro usage - > issue #10 

Passes on Fatnode for all relevant GCC and Clang versions.


tests/4.5/target/test_target_map_pointer_default.c
```diff
@@ -9,45 +9,51 @@
...
+  int i;
  
   for (i = 0; i < N; i++) 
     compute_array[i] = 0;
+  
   p = &compute_array[0];
 
-// p is mapped as a pointer to compute_array
+// p is default mapped tofrom as a pointer to compute_array
 #pragma omp target data map(tofrom: compute_array)
```
tmh97: Remove line 32 for default mapping, must keep this when testing implicit mapping

tmh97:
Added additional tests for default pointer behavior, passes all GCC and Clang versions

tmh97:
Split the file into two separate tests, re-ran both and they pass on relevant Clang and GCC versions.

tests/4.5/target/test_target_map_pointer_default.c
```diff
@@ -1,53 +1,58 @@
-//===--- test_target_map_pointer.c - test default behavior pointer mapping --===//
+//===--- test_target_map_pointer_default.c -------------------------------------===//
```
josemonsalve2: Rename this file to `test_target_map_pointer_no_map_type_modifier`

josemonsalve2:
After changing the file name it should be good to go

josemonsalve2:
If it runs, and there are no bugs, then go ahead and merge it.

PR #95, opened 2020-02-21
Fixing macro usage and adding additional test function

tmh97:
The original test has become test_scalar_to(), this passes on all relevant GCC and Clang Versions. Attempting to add additional test function test_scalar_from(). The latter is still a work in progress


tmh97:
Changed name and retesting with several versions of Clang and GCC, passes all on device.

PR #100, opened 2020-02-24
Add Fortran version of target teams distribute device test

jhdavis8:
This is a new version of the PR which doesn't include old changes caused by a bad git branch. See PR #64 for the original comments and description. 

At present the Fortran test fails gfortran with this error: `libgomp: GOMP_target_enter_exit_data unhandled kind 0x05`. It passes xlf.

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_device.c
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_device.c
@@ -25,9 +25,9 @@ int main() {
   int isOffloading = 0;
   OMPVV_TEST_AND_SET_OFFLOADING(isOffloading);
 
-  int a[ARRAY_SIZE];
-  int b[ARRAY_SIZE];
   int num_devices = omp_get_num_devices();
+  int a[num_devices][ARRAY_SIZE];
+  int b[ARRAY_SIZE];
   int num_teams[num_devices];
   int errors[num_devices];
   int sum_errors = 0;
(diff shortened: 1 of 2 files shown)


josemonsalve2:
@spophale @jhdavis8 This is one test where unified shared memory on 4.5 could result in incorrect values. This is the example I was looking for during our call. 

If unified shared memory is used, and the map is not performed, then devices will override eachother's memory. Resulting in a failure

jhdavis8:
Finally got a good reproducer written and found out someone else has already reported this bug... https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91762. Once everyone approves this can be merged.

tests/4.5/target_teams_distribute/test_target_teams_distribute_device.c
```diff
@@ -61,7 +63,7 @@ int main() {
     for (int x = 0; x < ARRAY_SIZE; ++x) {
       OMPVV_TEST_AND_SET_VERBOSE(errors[dev], a[x] != 1 + dev + b[x]);
       if (a[x] != 1 + dev + b[x]) {
-	break;
+        break;
       }
     }
   }
```
jhdavis8: Split up the array into sections for each device to work on to be mapped back non-overlapping.

jhdavis8: This change has been made and is passing all summit compilers.

jhdavis8:
Passes all summit compilers save for gfortran due to an already-reported bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=91762)

PR #101, opened 2020-02-25
QMCPACK timing test for parallel offload.

spophale:
This test has been picked from the Redmine Issue https://openmp-ecp.ornl.gov/redmine/issues/45

### DIFF
--- a/tests/4.5/application_kernels/qmcpack_parallel_offload.c
+++ b/tests/4.5/application_kernels/qmcpack_parallel_offload.c
@@ -3,8 +3,10 @@
 // OpenMP API Version 4.5 Nov 2015
 // 
 // This test checks for multiple parallel offloading kernels. 
-// Kernel1 does parallel offload without specifying a device number while 
-// kernel2 tries to offload to specific devices.
+// Kernel1 does parallel offload without specifying a device number. 
+// Kernel2 tries to offload to specific devices.
+// Kernel3 does parallel offload with nowait without specifying a device number. 
+// Kernel4 tries to offload to specific devices with nowait.
 // Adapted from https://reviews.llvm.org/D74145 to match V&V coding and
 // reporting style. 
 //
@@ -19,11 +21,10 @@
 
 #include "ompvv.h"
 
+const int N = 1024, num_teams = 8, num_th = 4;
 
 void qmcpack_kernel1(int ndev) {
-
-  const int N = 1024, num_teams = 8, num_th = 4;
-  OMPVV_INFOMSG("qmcpack_multiple_threads_offloading");
+  OMPVV_INFOMSG("qmcpack parallel offloading");
 
   //Spawn 1 thread per device
 #pragma omp parallel for num_threads(ndev)
@@ -36,10 +37,9 @@ void qmcpack_kernel1(int ndev) {
     }
   }
 }
-void qmcpack_kernel2(int ndev) {
 
-  const int N = 1024, num_teams = 8, num_th = 4;
-  OMPVV_INFOMSG("qmcpack_multiple_threads_offloading");
+void qmcpack_kernel2(int ndev) {
+  OMPVV_INFOMSG("qmcpack parrallel offloading to different devices");
 
   //Spawn 1 thread per device
 #pragma omp parallel for num_threads(ndev)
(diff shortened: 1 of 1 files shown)


spophale:
Oscar and QMCPACK team needs to look at this test to check if it covers all their requirements before we put it in our repo.

tests/4.5/application_kernels/qmcpack_parallel_offload.c
```diff
@@ -0,0 +1,87 @@
...
+
+  const clock_t duration1 = (clock() - start) * 1000 / CLOCKS_PER_SEC / N;
+  
+  if(num_dev > 1){
+    start = clock();
+    for (int i = 0; i < N; ++i) {
+      qmcpack_kernel2(num_dev);
+    }
+    const clock_t duration2 = (clock() - start) * 1000 / CLOCKS_PER_SEC / N;
+    printf("Avg time for parallel offloading to multiple devices: %ld ms\n",duration2);
```
josemonsalve2: I suggest we use OMPVV_INFOMSG

tests/4.5/application_kernels/qmcpack_parallel_offload.c
```diff
@@ -0,0 +1,87 @@
...
+      }
+    }
+  }
+}
+
+int main(int argc, char *argv[]) {
+  int i;
+  const int N = 1000;
+  int num_dev = omp_get_num_devices();
+  //printf("No. of devices available %d.\n", num_dev);
```
josemonsalve2: Remove this comment

tests/4.5/application_kernels/qmcpack_parallel_offload.c
spophale: Will do that after the test is finalized.

spophale: Actually, I don't know if it appropriate as I want the timings to be printed without additional flags.

josemonsalve2: The idea of the TEST_VERBOSE flag was to be able to obtain a simple pass fail report and no other output. If that's the case, then we either want to re-evaluate this design in the infrastructure, or provide a new flag. I don't think it is a good idea to have `printf`s with no other information.

tests/4.5/application_kernels/qmcpack_parallel_offload.c
spophale: Done.

tests/4.5/application_kernels/qmcpack_parallel_offload.c
spophale: This test is a little different from our usual conformance tests. If we see more of these then maybe we should think of a new macro.

josemonsalve2: I agree this is a different to our conformance tests. That's why I am conflicted about adding information that is output in this test (and only in this test) regarding performance, when our focus has always been V&V. If we want to consider performance, we should consider how to change the infrastructure for performance.

sunitachandra: V nice to see an app. test code! I like that we are printing timings!! Probably better to stick with the macros/ infrastructure design/software practice by turning on TEST_VERBOSE and using OMPVV_INFOMSG - that way we are strictly following our rules that we have set and then encourage anyone else to do the same

spophale: INFOMSGs were designed to give additional information for tests, I fear that the timings will get buried in all other messages. That said, current format is for easy testing during discussion phase where they may want to take it for a test drive. I am all in to design a macro that better fits timing reporting.

josemonsalve2: Yes. This is definitely needed. So far we had considered giving timing information for the duration of the whole test, but we have not discussed what happens when we want to give intermediate timing results.

spophale:
Feedback: Add meaningful work inside of the target regions to be safe from dead code elimination.

tests/4.5/application_kernels/qmcpack_parallel_offload.c
```diff
@@ -0,0 +1,135 @@
...
+    for (size_t i = 0; i < N; ++i) {
+      for (size_t j = 0; j < N/2 ; ++j) {
+        int tmp_val = N *i*j;
+      }
+    }
+  }
+}
+
+void qmcpack_kernel4(int ndev) {
+  OMPVV_INFOMSG("qmcpack parrallel offloading to different devices with nowait");
```
jhdavis8: Quick spelling fix, parrallel -> parallel.

spophale:
I am closing this PR as this is the only one with timing. If we were to go in the benchmarking direction, more ground up changes would be required to the V&V.

PR #103, opened 2020-02-25
Modifying test_target_private.c to use headerfile (Issue #10)

tmh97:
Making style changes and including macro usage. Still need to add description and do some more testing, currently passing on GCC and Clang.

### DIFF
--- a/tests/4.5/target/test_target_private.c
+++ b/tests/4.5/target/test_target_private.c
@@ -4,20 +4,27 @@
 //
 //Testing private clause with target directive
 ////===----------------------------------------------------------------------===//
+
 #include <omp.h>
 #include <stdio.h>
+#include "ompvv.h"
 
 #define N 10
 
 int main() {
+
   int compute_array[4][N];
   int errors = 0;
-  int i,j, isHost = -1;
-  
- 
-  for (i=0; i<4; i++) 
-    for (j=0; j<N; j++) 
+  int i, j;
+
+  //Check for offloading
+  int is_offloading;
+  OMPVV_TEST_AND_SET_OFFLOADING(is_offloading);
+
+  for (i = 0; i < 4; i++) 
+    for (j = 0; j < N; j++) 
       compute_array[i][j] = 0;
+
 
   omp_set_num_threads(4);
 #pragma omp parallel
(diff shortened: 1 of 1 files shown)


tests/4.5/target/test_target_private.c
```diff
@@ -4,53 +4,53 @@
 //
 //Testing private clause with target directive
```
josemonsalve2: Improve the description.

tests/4.5/target/test_target_private.c
```diff
@@ -4,53 +4,53 @@
...
-  } // End target
-}//end-parallel
+  } // end target
+}//end parallel
 
-  for (i=0; i<4; i++){ 
-    for (j=0; j<N; j++){
-      if(compute_array[i][j] != 100){
-        errors += 1;
+  for (i = 0; i < 4; i++){
```
josemonsalve2: The actual number of threads may be less than 4. So we would need to get `omp_get_num_threads` in the parallel region, and use that number in this verification code.

tests/4.5/target/test_target_private.c
```diff
@@ -4,53 +4,53 @@
...
 
-  for (i=0; i<4; i++){ 
-    for (j=0; j<N; j++){
-      if(compute_array[i][j] != 100){
-        errors += 1;
+  for (i = 0; i < 4; i++){ 
+    for (j = 0; j < N; j++){
+      OMPVV_TEST_AND_SET_VERBOSE(errors, (compute_array[i][j] != 100));
+      if (compute_array[i][j] != 100) 
         printf("compute_array[%d][%d] = %d\n",i,j,compute_array[i][j]);
```
josemonsalve2: Use `OMPVV_ERROR_IF(compute_array[i][j] != 100, "compute_array[%d][%d] = %d\n",i,j,compute_array[i][j])` instead

tmh97: Taking care of these recommendations now, thank you!

tests/4.5/target/test_target_private.c
tmh97: Is it valid to replace 44 with `for (i = 0; i < omp_get_thread_num(); i++)`

josemonsalve2: No because it is already outside of the parallel region, so the number of threads is always 1. You have to use a variable to probe the number of threads in the parallel region. 

In general, `omp_get_num_threads()` will give you the number of threads in the current execution context where it is called.

tmh97:
@josemonsalve2 I updated header, added a probe for the real number of threads, and added OMPVV_ERROR_IF macro at the end instead of traditional printf setup. Let me know how these changes are looking, thanks again for looking this over.

tmh97:
Testing again before merging, passes on gcc and clang.

PR #104, opened 2020-02-25
Testing invocation of math functions inside target region.

spophale:
QMCPACK team suggested this test as it is important for them that math functions are supported in offloaded kernels.


tests/4.5/application_kernels/qmcpack_target_math.c
```diff
@@ -0,0 +1,46 @@
+//===--- qmcpack_target_math.c --- math lib invocation inside target---------===//
+//
+// OpenMP API Version 4.5 Nov 2015
+//
+//
```
josemonsalve2: Add description

josemonsalve2:
I don't think we should have math.h tests to be part of the specification tests. We cannot claim that an implementation is not compliant with the specification if it does not run math.h functions. 

At some point we discussed separating this sort of tests that are not "specification" tests but "application specific tests".

tests/4.5/application_kernels/qmcpack_target_math.c
```diff
@@ -0,0 +1,46 @@
...
+  OMPVV_INFOMSG("test_math_lib_inside_target");
+
+  double array[N];
+  int errors = 0;
+  
+  // Array initialization
+  for (int i = 0; i < N; ++i) {
+    array[i] = 0.99;
+  }
+
```
ye-luo: Let us make this test closer to the real code and more challenging.
Please a dummy line
`int c99_zero = FP_ZERO;`
This should choke clang in the CUDA pass.

spophale: FP_ZERO is for floating point. Is there some reason you want to assign it to int ?

ye-luo: FP_ZERO is one of the return value of fpclassify, the return type is int.
https://www.gnu.org/software/libc/manual/html_node/Floating-Point-Classes.html

tmh97:
@spophale is this PR ready for merge?

spophale:
I want to separate the two tests. Then merge.

ye-luo:
> I want to separate the two tests. Then merge.

Maybe it is better to merge this one in a timely fashion and then make a new PR to split the test.

PR #107, opened 2020-02-27
Modifying test_target_update_if.c to use headerfile (Issue #10)

tmh97:
Test is ready for review, passes with clang and gcc on the device. I noticed that test fails on the host when NO_OFFLOADING is enabled for GCC compilation, however, clang passes fine on the host even with NO_OFFLOADING. Need to decide if this is a concern or not considering it is a 'target' test.

### DIFF
--- a/tests/4.5/target_update/test_target_update_if.c
+++ b/tests/4.5/target_update/test_target_update_if.c
@@ -15,8 +15,10 @@
 
 #include <omp.h>
 #include <stdio.h>
+#include "ompvv.h"
 
 #define N 100
+
 int a[N];
 int b[N];
 int c[N];
@@ -39,17 +41,15 @@ int init_b(){
   
 // Test for OpenMP 4.5 target update  with if
 int main() {
-  int errors[2]={0,0}, i = 0, isHost = -1, isOffloading = 0, change_flag=0;
+  int errors[2]={0,0}, i = 0, report_errors = 0, change_flag = 0;
 
   for (i = 0; i < N; i++) {
     a[i] = 10;
   }
 
   // We test for offloading
-#pragma omp target map(from: isOffloading)
-  {
-    isOffloading = !omp_is_initial_device();
-  }
+  int is_offloading;
+  OMPVV_TEST_AND_SET_OFFLOADING(is_offloading); 
 
 for(count=0; count< 4;count++){
   // a and b array initialization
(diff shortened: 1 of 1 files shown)


tests/4.5/target_update/test_target_update_if.c
```diff
@@ -39,29 +41,26 @@ int init_b(){
...
   // a and b array initialization
   for (i = 0; i < N; i++) {
     b[i] = 2; 
     c[i] = 0;
   }
-#pragma omp target data map(to: a[:N], b[:N]) map(tofrom: c) map(tofrom: isHost)
+#pragma omp target data map(to: a[:N], b[:N]) map(tofrom: c)
 {
-  #pragma omp target //map(to: a[:N], b[:N], isHost)
+  #pragma omp target //map(to: a[:N], b[:N])
```
josemonsalve2: remove comment

tests/4.5/target_update/test_target_update_if.c
```diff
@@ -39,29 +41,26 @@ int init_b(){
...
 
   // We test for offloading
-#pragma omp target map(from: isOffloading)
-  {
-    isOffloading = !omp_is_initial_device();
-  }
+  int is_offloading;
+  OMPVV_TEST_AND_SET_OFFLOADING(is_offloading); 
 
 for(count=0; count< 4;count++){
```
josemonsalve2: spaces between `<`,  `=`  and  `;`

tests/4.5/target_update/test_target_update_if.c
```diff
@@ -39,29 +41,26 @@ int init_b(){
...
   // a and b array initialization
```
josemonsalve2: b and c are initialized

tests/4.5/target_update/test_target_update_if.c
```diff
@@ -71,7 +70,7 @@ for(count=0; count< 4;count++){
   change_flag = init_b();
   #pragma omp target update if (change_flag) to(b[:N]) //update b=4 for all odd iterations
```
josemonsalve2: Is this comment of b=4 accurate?

tests/4.5/target_update/test_target_update_if.c
```diff
@@ -71,7 +70,7 @@ for(count=0; count< 4;count++){
...
 
-  #pragma omp target //map(from: a,b,c) map(tofrom: isHost)//mapping from to counter default tofrom mapping on b
+  #pragma omp target //map(from: a,b,c) //mapping from to counter default tofrom mapping on b
```
josemonsalve2: remove the map(from) and potentially the comment. It's ok if you have a comment here explaining what's the purpose, but we need to fix this wording I believe.

tests/4.5/target_update/test_target_update_if.c
```diff
@@ -71,7 +70,7 @@ for(count=0; count< 4;count++){
...
   {
         int j = 0;
         for (j = 0; j < N; j++) {
```
josemonsalve2: Line 83: Spaces of comment

tests/4.5/target_update/test_target_update_if.c
tmh97: Good catch, I do not believe this is accurate. Will look at this closer and fix.

tmh97: @josemonsalve2 looked into this, b=4 is correct value during "odd" iterations. This is good to go.

PR #108, opened 2020-02-27
Modifying test_target_update_from.c to use headerfile (Issue #10)

tmh97:
Made simple macro additions, added new description, changed name. Test passes on GCC and Clang for all relevant versions. Similar to PR #107 the test fails on host with gcc, but passes with clang. Is this expected for 'target_update' test?

### DIFF
--- a/tests/4.5/target_update/test_target_update_from.c
+++ b/tests/4.5/target_update/test_target_update_from.c
@@ -14,6 +14,7 @@
 #include "ompvv.h"
 
 #define N 100
+
 int a[N];
 int b[N];
 int c[N];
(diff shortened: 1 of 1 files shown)


tmh97:
Also removed update_b() function as it was not called in program

tests/4.5/target_update/test_target_update_from.c
```diff
@@ -58,16 +56,9 @@ int main() {
 }// end target-data
```
jhdavis8: Indentation width is too long on the lines above here

josemonsalve2: Also after linee 48, we should check for the value of b is correct, on the host side.

tmh97:
Added indentation fixes, as well as an extra check at the end to ensure host value is updated.

tests/4.5/target_update/test_target_update_from.c
```diff
@@ -1,73 +1,66 @@
...
-        for (j = 0; j < N; j++) {
-          b[j] = (a[j] + b[j]);//b=12 
-        }
+    int j = 0;
+    for (j = 0; j < N; j++) {
+      b[j] = (a[j] + b[j]);//b=12 
+    }
   } // end target
 
   #pragma omp target update from(b[:N]) //update b=12 on host
```
spophale: I don't see the value of modifying and testing array c. We could very well end device and date env at 47 and check value of b.

tmh97: I agree, should I go ahead and remove target section where c is dealt with (lines 48-53), as well as remove lines 59-60 where c is checked?

tmh97:
Changed a few things around, now the test checks for correct values of array b on the host immediately after the update clause. Let me know how this looks, passes GCC and Clang.

PR #111, opened 2020-03-10
Modifying test_target_enter_data_malloced.. to use macros

tmh97:
Added macro usage, adjusted formatting issues, checks for offloading, tested and passed on GCC and Clang. Issue #10 *Still need to finish header*.

### DIFF
--- a/tests/4.5/target_enter_data/test_target_enter_data_malloced_array.c
+++ b/tests/4.5/target_enter_data/test_target_enter_data_malloced_array.c
@@ -17,7 +17,6 @@
 #include <omp.h>
 #include "ompvv.h"
 
-// Test for OpenMP 4.5 target enter and target exit data with allocated arrays.
 
 // global variables
 int n=10, B[10];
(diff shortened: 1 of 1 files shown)


tests/4.5/target_enter_data/test_target_enter_data_malloced_array.c
```diff
@@ -1,51 +1,66 @@
...
 
-void init(int **A){
+void init(int **A) {
  int i;
- *A = (int *)malloc(n*sizeof(int));
- if(NULL == *A) exit(-1);
- x=*A;
- for(i=0; i<n; i++){
+ *A = (int *) malloc(n*sizeof(int));
+ if (NULL == *A) exit(-1);
```
josemonsalve2: We should consider an OMPVV_ERROR here as well as the exit. Or overpass the whole test and let the REPORT say that it failed.

tests/4.5/target_enter_data/test_target_enter_data_malloced_array.c
```diff
@@ -1,51 +1,66 @@
...
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <omp.h>
+#include "ompvv.h"
 
 // Test for OpenMP 4.5 target enter and target exit data with allocated arrays.
 
+// global variables
 int n=10, B[10];
```
josemonsalve2: Any reason why n is not a `#define`?

tmh97: good catch, will swap out.

tests/4.5/target_enter_data/test_target_enter_data_malloced_array.c
tmh97: Going to go with an OMPVV_ERROR message indicating it has failed

tests/4.5/target_enter_data/test_target_enter_data_malloced_array.c
tmh97: Seems it has to be a named variable, throwing issues when I try to use #define

PR #112, opened 2020-03-11
Modifying test_target_enter_exit_data_depend.c to use headerfile

tmh97:
Making style changes, including macro usage, currently passes on Clang and GCC. Still need to add header information. Issue #10


tmh97:
Consider nowait on line 102

tmh97:
Split up potentially

PR #113, opened 2020-03-11
Modifiying test_target_enter_exit_data_devices.c to use headerfile

tmh97:
Making style changes, adding macros, passes on clang and GCC.

### DIFF
--- a/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_devices.c
+++ b/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_devices.c
@@ -1,3 +1,19 @@
+//===--- test_target_enter_exit_data_devices.c---------------------------------------===//
+//
+//  OpenMP API Version 4.5 Nov 2015
+//
+//  This test checks for data mapping on multiple devices when using the target
+//  data map directive. The test uses omp_get_num_devices to confirm the number 
+//  of total devices available. There are two separate functions, and both make sure 
+//  that data mapping is happening on each available device. The first function does 
+//  this through the use of omp_set_default_device, while the second function explicity 
+//  uses the device() clause to control which device is being utilized.
+//
+//  Since OpenMP 4.5 does not have an API call to obtain the current device, 
+//  this test does not guarantee that the execution devices are different. 
+//
+////===------------------------------------------------------------------------------===/
+
 #include <assert.h>
 #include <omp.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


tmh97:
is target data map on line 72 redundant?

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_devices.c
```diff
@@ -17,29 +34,23 @@ int test_set_default_dev() {
   printf("initial device: %d\n", omp_get_initial_device());
   printf("default device: %d\n", def_dev);
```
josemonsalve2: Change all these printf's to OMPVV_INFOMSG

josemonsalve2: Change these to `OMPVV_INFOMSG`

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_devices.c
```diff
@@ -17,29 +34,23 @@ int test_set_default_dev() {
...
     omp_set_default_device(dev);
-    // unstructured mapping
-    {
+
 #pragma omp target enter data map(alloc: h_matrix[dev][0 : N])
-        printf(""); // forcing the compiler to not moving out of the scope
-    }
-    // operation
+    
     // assert(dev == omp_get_default_device() && "message");
```
josemonsalve2: remove this comment

josemonsalve2:
I did not see the map on line 72. 


However, this test suffers from the same problem that @jhdavis8 's test has. If we decide not to use the TEST_SHARED_ENV option, it is not possible to make sure this test is compliant. But if we don't use this scheme we have no way we can fully guarantee that target regions are running on different devices

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_devices.c
```diff
@@ -1,45 +1,55 @@
...
 
-  printf("test_set_default_dev\n");
+  OMPVV_INFOMSG("test_set_default_dev");
 
   // Get number of devices
   int num_dev = omp_get_num_devices();
-  printf("num_devices: %d\n", num_dev);
+  OMPVV_INFOMSG("num_devices: %d", num_dev);
 
   int def_dev = omp_get_default_device();
```
spophale: Although I don't find a relationship between num_devices and default device no, I wonder if the device no is between 1 and num_dev.

tmh97: Check what value default is ->

tmh97: On Fatnode -> num_dev = 2, default_dev = 0, and initial device = -10

spophale: how are you getting initial device ?

tmh97: omp_get_initial_device()

PR #115, opened 2020-03-19
Modifying test_target_data_map_array_sections.c to use macros

tmh97:
Tested and passing on clang and GCC


tests/4.5/target_data/test_target_data_map_array_sections.c
```diff
@@ -1,6 +1,20 @@
-#include <omp.h>
```
jhdavis8: Might be good to keep this include, just to have access to the runtime functions.

PR #117, opened 2020-03-22
Modifying test_target_enter_exit_data_if.c to use headerfile

tmh97:
Currently passing on GCC and Clang. May need changes to test, specifically size_threshold strategy.

### DIFF
--- a/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_if.c
+++ b/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_if.c
@@ -1,4 +1,4 @@
-//===---- test_target_enter_exit_data_if.c - check the if clause of target data ------===//
+//===---- test_target_enter_exit_data_if.c --------------------------------===//
 // 
 // OpenMP API Version 4.5 Nov 2015
 // 
@@ -20,6 +20,7 @@
 
 #include <omp.h>
 #include <stdio.h>
+#include "ompvv.h"
 
 #define SIZE_THRESHOLD 512
 
@@ -28,7 +29,7 @@ int main() {
   int a[1024];
   int b[1024];
   int c[1024];
-  int size, i = 0, errors[2] = {0,0}, isHost = -1, isOffloading = 0;
+  int size, i = 0, errors[2] = {0,0}, isHost = -1;
 
   // a and b array initialization
   for (i = 0; i < 1024; i++) {
@@ -37,10 +38,13 @@ int main() {
   }
 
   // We test for offloading
-#pragma omp target map(from: isOffloading)
-  {
-    isOffloading = !omp_is_initial_device();
-  }
+  int isOffloading; 
+  OMPVV_TEST_AND_SET_OFFLOADING(isOffloading);
+ 
+  if (!isOffloading)
+  OMPVV_WARNING("It is not possible to test conditional data transfers "
+                 "if the environment is shared or offloading is off. Not testing "
+                 "anything"); 
 
   // check multiple sizes. 
   for (size = 256; size <= 1024; size += 256) {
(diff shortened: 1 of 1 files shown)


tests/4.5/target_enter_exit_data/test_target_enter_exit_data_if.c
```diff
@@ -20,6 +20,7 @@
 
 #include <omp.h>
 #include <stdio.h>
+#include "ompvv.h"
 
 #define SIZE_THRESHOLD 512
```
josemonsalve2, line 26: comment below this line is incorrect.

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_if.c
```diff
@@ -37,10 +38,13 @@ int main() {
   }
 
   // We test for offloading
-#pragma omp target map(from: isOffloading)
-  {
-    isOffloading = !omp_is_initial_device();
-  }
+  int isOffloading;
```
josemonsalve2, line 41: Move this to the beginning as it was before?

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_if.c
```diff
@@ -37,10 +38,13 @@ int main() {
...
   // We test for offloading
-#pragma omp target map(from: isOffloading)
-  {
-    isOffloading = !omp_is_initial_device();
-  }
+  int isOffloading; 
+  OMPVV_TEST_AND_SET_OFFLOADING(isOffloading);
+ 
+  if (!isOffloading)
+  OMPVV_WARNING("It is not possible to test conditional data transfers "
```
josemonsalve2, line 45: use OMPVV_WARNING_IF(!isOffloading). And double check if the shared env is also true here.

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_if.c
```diff
@@ -28,7 +29,7 @@ int main() {
   int a[1024];
   int b[1024];
   int c[1024];
-  int size, i = 0, errors[2] = {0,0}, isHost = -1, isOffloading = 0;
```
josemonsalve2, line 31: No need to create isHost in here. I suggest you init in the target region.

josemonsalve2: This also means you don't need to do the map(tofrom: ishost)

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_if.c
```diff
@@ -37,10 +38,13 @@ int main() {
...
+  int isOffloading; 
+  OMPVV_TEST_AND_SET_OFFLOADING(isOffloading);
+ 
+  if (!isOffloading)
+  OMPVV_WARNING("It is not possible to test conditional data transfers "
+                 "if the environment is shared or offloading is off. Not testing "
+                 "anything"); 
 
   // check multiple sizes. 
   for (size = 256; size <= 1024; size += 256) {
```
josemonsalve2, line 50: I'm not so sure about this test now. What we want to check if tarrget enter exit data if, not target if. But I can't think of a way that we could do this without getting into shared data environment discussions.

The test should be checking if the if clause in the enter and exit data directives have any effect in the semantics.

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_if.c
```diff
@@ -82,17 +86,17 @@ int main() {
...
-              printf("Test failed on host and device with offloading %s.\n", (isOffloading ? "enabled" : "disabled"));
+  if (!errors[0] && !errors[1]) {
+    OMPVV_INFOMSG("Test passed with offloading %s", (isOffloading ? "enabled" : "disabled"));
+  } else if (errors[0]==0 && errors[1]!=0) {
+      OMPVV_ERROR("Test failed on host with offloading %s.", (isOffloading ? "enabled" : "disabled"));
+    } else if (errors[0]!=0 && errors[1]==0) {
+        OMPVV_ERROR("Test failed on device with offloading %s.", (isOffloading ? "enabled" : "disabled"));
+      } else if (errors[0]!=0 && errors[1]!=0) {
+              OMPVV_ERROR("Test failed on host and device with offloading %s.", (isOffloading ? "enabled" : "disabled"));
+        }
```
josemonsalve2, line 99: Some of this is already done in OMPVV_REPORT_AND_RETURN.

PR #120, opened 2020-04-04
QMCPACK target static lib test and add OMPVV static lib

jhdavis8:
In this PR:

1. `qmcpack_target_static_lib.c`: A test of the ability to have a target region in a static library which checks that the scalar value modified on the device is correctly mapped back.

2. `libompvv.c` and `libompvv.h`: Static library for the whole test suite including the offload test function used in the QMCPACK test above.

3. Changes to `Makefile`: Adds linking of libompvv.a. Documents in the help section the NUM_THEADS/TEAMS options that were missing.

4. Changes to `.gitignore`: Ignore .a static library files which should not be tracked.

These changes are tested and passing on all summit compilers. Right now the `libompvv.o` file I store in `/bin` which I think is correct, but all the other `libompvv.*` files including `libompvv.a` are stored in `/ompvv`. I am open to changing this by moving these files to a new `/lib` directory if we think that is better practice.


Makefile
```diff
@@ -452,6 +476,10 @@ help:
 	@echo "  MODULE_LOAD=1             Before compiling or running, module load is called"
 	@echo "  ADD_BATCH_SCHED=1         Add the jsrun command before the execution of the running script to send it to a compute node"
 	@echo "  NO_OFFLOADING=1           Turn off offloading"
+	@echo "  LINK_OMPVV_LIB=1          Link the OMPVV static library (found in sollve_vv/ompvv)"
```
josemonsalve2, line 479: What happens if we just always link it?

jhdavis8: I think this would work, since you have to include the header file anyways if you want to use it. I'll make this change.

jhdavis8:
Tested and passing all summit compilers.

PR #121, opened 2020-04-05
Add tulip.def, AMD compiler flags for Cray CCE, and DEVICE_TYPE Makefile option

jhdavis8:
In this PR:

1. Add `tulip.def` so the infrastructure will automatically load the needed modules to compile on Tulip when `SYSTEM=tulip MODULE_LOAD=1` is provided.

2. Modify `make.def` to add Cray compiler flags for AMD devices, and an option called `DEVICE_TYPE` which changes the flags appropriately for Nvidia and AMD targets. I have not yet configured these flags for the Cray Fortran compiler, `ftn`.

3. Modify `Makefile` to document the new `DEVICE_TYPE` flag.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -449,6 +449,7 @@ help:
 	@echo "  LOG_ALL=1                 Enables dump of the make process output, errors, and binary execution outputs into logs.txt"
 	@echo "  SYSTEM=sys_name           Includes the definitions for the requires modules and batch schedulers in the different systems"
 	@echo "                            This definitions must be in sys/SYSTEM.def. (Do not include the .def extension)"
+	@echo "  DEVICE_TYPE=dev_name      Specifies device type being used, either nvidia or amd, to change compiler flags as needed."
 	@echo "  MODULE_LOAD=1             Before compiling or running, module load is called"
 	@echo "  ADD_BATCH_SCHED=1         Add the jsrun command before the execution of the running script to send it to a compute node"
 	@echo "  NO_OFFLOADING=1           Turn off offloading"
--- a/sys/make/make.def
+++ b/sys/make/make.def
@@ -67,11 +67,20 @@ CC?=none
 CPP?=none
 FC?=none
 
-# CRAY compiler
-ifeq ($(CC), cc)
-  CFLAGS += -fopenmp -fopenmp-targets=nvptx64 -Xopenmp-target -march=sm_70
-  CLINK = cc
-  CLINKFLAGS += -fopenmp -fopenmp-targets=nvptx64 -Xopenmp-target -march=sm_70
+ifeq ($(DEVICE_TYPE), amd)
+  # CRAY compiler (AMD)
+  ifeq ($(CC), cc)
+    CFLAGS += -fopenmp -fopenmp-targets=amdgcn-amd-amdhsa -Xopenmp-target=amdgcn-amd-amdhsa -march=gfx906
+    CLINK = cc
+    CLINKFLAGS += -fopenmp -fopenmp-targets=amdgcn-amd-amdhsa -Xopenmp-target=amdgcn-amd-amdhsa -march=gfx906
+  endif
+else
+  # CRAY compiler (NVIDIA)
+  ifeq ($(CC), cc)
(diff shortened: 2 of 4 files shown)


josemonsalve2:
Check what I did for AOMP. I am thinking that we need to determine this differently for all the compilers. This is also the case for GCC and Clang. Although we gotta wait for GCC 10 for AMD support, and for AMD to push the clang support, they will eventually need a mechanism to do this.

What I suggest is that we try to decouple the device discovery from the flag selection in each compiler, by having some common logic executing first figuring out if it is AMD, what AMD arch it is, or if it is NVIDIA, what sm it is?

jhdavis8:
Added Fortran flags, tested and working on tulip

Makefile
```diff
@@ -449,6 +449,7 @@ help:
 	@echo "  LOG_ALL=1                 Enables dump of the make process output, errors, and binary execution outputs into logs.txt"
 	@echo "  SYSTEM=sys_name           Includes the definitions for the requires modules and batch schedulers in the different systems"
 	@echo "                            This definitions must be in sys/SYSTEM.def. (Do not include the .def extension)"
+	@echo "  DEVICE_TYPE=dev_name      Specifies device type being used, either nvidia or amd, to change compiler flags as needed."
```
josemonsalve2, line 452: I think we shouldn't add a new flag for this. AOMP already shows a way to discover AMD and nvidia-smi may be another option. However, this could be added in a different PR

sys/make/make.def
```diff
@@ -67,11 +67,21 @@ CC?=none
...
-  CLINKFLAGS += -fopenmp -fopenmp-targets=nvptx64 -Xopenmp-target -march=sm_70
+# CRAY compilers
+ifeq ($(DEVICE_TYPE), amd)
+  # CRAY compiler (AMD)
+  ifeq ($(CC), cc)
+    CFLAGS += -fopenmp -fopenmp-targets=amdgcn-amd-amdhsa -Xopenmp-target=amdgcn-amd-amdhsa -march=gfx906
+    CLINK = cc
+    CLINKFLAGS += -fopenmp -fopenmp-targets=amdgcn-amd-amdhsa -Xopenmp-target=amdgcn-amd-amdhsa -march=gfx906
+  endif
+else
```
josemonsalve2, line 78: Probably else if and compare with nvidia, for the sake of future accelerators (e.g. Aurora by NEC)

sys/make/make.def
```diff
@@ -67,11 +67,21 @@ CC?=none
...
+  # CRAY compiler (NVIDIA)
+  ifeq ($(CC), cc)
+    CFLAGS += -fopenmp -fopenmp-targets=nvptx64 -Xopenmp-target -march=sm_70
+    CLINK = cc
+    CLINKFLAGS += -fopenmp -fopenmp-targets=nvptx64 -Xopenmp-target -march=sm_70
```
josemonsalve2, line 83: is the march absolutely necessary? wouldn't the compiler chose the default value (e.g. in LLVM when you build there is a flag to select the default one). This could help when running in other systems.

sys/make/make.def
```diff
@@ -67,11 +67,21 @@ CC?=none
...
-# CRAY compiler
-ifeq ($(CC), cc)
-  CFLAGS += -fopenmp -fopenmp-targets=nvptx64 -Xopenmp-target -march=sm_70
-  CLINK = cc
-  CLINKFLAGS += -fopenmp -fopenmp-targets=nvptx64 -Xopenmp-target -march=sm_70
+# CRAY compilers
+ifeq ($(DEVICE_TYPE), amd)
+  # CRAY compiler (AMD)
+  ifeq ($(CC), cc)
+    CFLAGS += -fopenmp -fopenmp-targets=amdgcn-amd-amdhsa -Xopenmp-target=amdgcn-amd-amdhsa -march=gfx906
```
josemonsalve2, line 74: Check out how we can discover the AMD Architecture for AOMP.

jhdavis8:
Going to merge this so we can test tulip. Llthough there are changes that would improve this, I would like to address these in a new PR.

Issue #129, opened 2020-04-22
test_target_teams_distribute_default_firstprivate.F90 fails when run


k-yeung:
tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90 fails both tests when compiled with GFortran (current git trunk) and run with an AMD GCN accelerator target.

```
[OMPVV_INFOMSG test_target_teams_distribute_default_firstprivate.F90:21] Test is running on device
[OMPVV_ERROR test_target_teams_distribute_default_firstprivate.F90:24]  Condition test_firstprivate_private() .ne. 0 failed
[OMPVV_ERROR test_target_teams_distribute_default_firstprivate.F90:25]  Condition test_firstprivate_first() .ne. 0 failed
[OMPVV_INFOMSG ompvv.F90:244] The value of errors is 2
[OMPVV_RESULT test_target_teams_distribute_default_firstprivate.F90] Test failed on the device.
```

If a write statement is added to print the value of d(x) being compared, they will all be zeroes.

The problem is that the 'default(firstprivate)' clause is being applied to all the variables (including d), so the result of the computation does not make it out of the enclosing construct.

From the OpenMP 4.5 specification:

2.15.1.1: Variables with implicitly determined data-sharing attributes are those that are referenced in a given construct, do not have predetermined data-sharing attributes, and are not listed in a data-sharing attribute clause on the construct.

2.15.3.1: The default clause explicitly determines the data-sharing attributes of variables that are referenced in a parallel, teams, or task generating construct and would otherwise be implicitly determined...

I believe this behaviour is correct according to the spec, as listing the variables in the 'map' clause does not count as being 'listed in a data-sharing attribute clause'.

spophale:
Thanks @k-yeung for reporting it. Yes, I see what you mean by differentiating between data movement vs. data sharing attribute.

jhdavis8:
See PR #131

jhdavis8:
PR #131 has been merged.

PR #131, opened 2020-04-22
Add fixed version of default_firstprivate test

jhdavis8:
Addressing issue #129. My proposed fix is to add data sharing attributes for the arrays that have been improperly made firstprivate. However, the "first" part of the test is still failing on gfortran, but I am hesitant to file a bug report until we are sure because they have already ruled a bug report from me as invalid for this test (due to the issue this PR is fixing). See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=94290.

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_default_firstprivate.F90
@@ -100,7 +100,7 @@ CONTAINS
 
     !$omp target data map(from: d(1:N), num_teams(1:N)) &
     !$omp& map(to: a(1:N), b(1:N), c(1:N))
-    !$omp target teams distribute default(firstprivate) &
+    !$omp target teams distribute default(firstprivate) shared(d) &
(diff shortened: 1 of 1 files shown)


tob2:
There is in any case an issue in the second function `test_firstprivate_first`. One has:
```
    !$omp target teams distribute default(firstprivate) map(alloc:d) …
```
which is the same as
```
!$omp target map(alloc:d)
! d is not initialized here
!$omp teams distribute default(firstprivate)
do …
   d = …  ! d is firstprivate here as instructed
end do …
! team distributed ends here;
! d is still uninitialized 
```
i think you want to add `shared(d)` to the `omp target teams` …

tob2:
Any progress? With my additional change – adding `shared(x)`, it does work for me when testing. (Besides making sense semantically.)

jhdavis8:
@tob2 Sorry I missed your first comment! I will correct this today

jhdavis8:
Fixed, passes Summit gfortran and xlf. Thanks @tob2 !

tob2:
What's actually the status of this pull request? There are two approvals – since a week. it is also marked as `ready`. (For what it is worth: the PR also looks fine to me.)

Only the third requested reviewer (@josemonsalve2) has not commented/approved it, yet.

jhdavis8:
It looks ready, I'll go ahead and merge.

PR #132, opened 2020-04-23
Auto-detect GCC and clang versions

mjklemm:


mjklemm:
This PR is about adding some auto-detection capabilities for the compiler versions to the test suite.  I have made the changes for GCC and clang as an example.  Please let me know if this interesting to you.  I will then make the modifications for the remaining compilers, too.

tob2:
Just a comment as a bystander.

For GCC it prints here:
`9.3.1 20200406 [revision 6db837a5288ee3ca5ec504fbd5a765817e556ac2]`
is this intended or should it be `…*\).*/\1/g` with a tailing `.*` – i.e. printing `9.3.1`?

In terms of POSIX, it should be `head -n 1` not `head -1` as the `-n` is required by POSIX. (But I assume it also works w/o -n nearly everywhere.)

mjklemm:
> Just a comment as a bystander.
> 
> For GCC it prints here:
> `9.3.1 20200406 [revision 6db837a5288ee3ca5ec504fbd5a765817e556ac2]`
> is this intended or should it be `…*\).*/\1/g` with a tailing `.*` – i.e. printing `9.3.1`?

Hm, is this the case for (nightly) development builds of GCC?  If so, then it might be a good idea to match against that too and include it.  I will have to check clang what they do.

> In terms of POSIX, it should be `head -n 1` not `head -1` as the `-n` is required by POSIX. (But I assume it also works w/o -n nearly everywhere.)

Good catch!  I have fixed that.

tob2:
> > `9.3.1 20200406 [revision 6db837a5288ee3ca5ec504fbd5a765817e556ac2]`

> Hm, is this the case for (nightly) development builds of GCC?

That's the normal (open)SUSE distribution build (`/usr/bin/gcc`). The date string should be always present, I think, and the one in the bracket appears when running `./contrib/gcc_update` before building GCC (which is seemingly what SUSE did). But Linux distributions in general simply pick a current branch version and not a release to get the newest fixes. — The development version is `10.0.0` (and will be branched relatively soonish).

spophale:
Thanks @mjklemm and @tob2 .

jhdavis8:
Not to bring up an old pull request... but it looks like (on Summit at least) these changes have broken our compiler version recording for the web results, since the appropriate modules are not loaded before evaluation. I believe it should be fixable with our loadModules custom function. Will create a PR when it is implemented.

PR #136, opened 2020-05-01
Another possible solution.

spophale:
I like this better because 1. parallel on host seems more reliable 2. private is only on target (and not on parallel as well.)


tmh97:
Looks good to me

PR #137, opened 2020-05-02
New test/team/test team on host.c

spophale:
Tested on xlc by adding target before teams construct. Does not compile otherwise.

### DIFF
--- a/tests/5.0/teams/test_teams.c
+++ b/tests/5.0/teams/test_teams.c
@@ -27,7 +27,7 @@ int main() {
     num_threads[x] = -99;
   }
 
-#pragma omp teams num_teams(3) thread_limit(N)
+#pragma omp teams num_teams(N) thread_limit(N)
 {
     num_teams[omp_get_team_num()] = omp_get_num_teams();
     num_threads[omp_get_team_num()]= omp_get_num_threads();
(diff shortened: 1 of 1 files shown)


tests/5.0/teams/test_teams.c
```diff
@@ -0,0 +1,60 @@
...
+
+
+  for (int x = 0; x < N; ++x) {
+    num_teams[x] = -99;
+    num_threads[x] = -99;
+  }
+
+#pragma omp teams num_teams(N) thread_limit(N)
+{
+    num_teams[omp_get_team_num()] = omp_get_num_teams();
```
tmh97, line 32: Is return value of 0 the expected behavior of omp_get_team_num()?

tests/5.0/teams/test_teams.c
```diff
@@ -0,0 +1,60 @@
...
+    num_threads[omp_get_team_num()]= omp_get_num_threads();
+
+}
```
jhdavis8, line 35: This is minor but I think this code block should be indented by one level

spophale:
Ping @jhdavis8 & @tmh97

tmh97:
This is ready to go

PR #142, opened 2020-05-06
New test/team/test_team_distribute_default_none.c

spophale:


tests/5.0/teams/test_teams_distribute_default_none.c
```diff
@@ -0,0 +1,76 @@
...
+  OMPVV_WARNING("Test only uses default(none) clause and does not guarantee that the default(none) is enforced.");
+
+  for (int x = 0; x < N; ++x) {
+    a[x] = 1;
+    b[x] = x;
+    c[x] = 2*x;
+    d[x] = 0;
+  }
+
+#pragma omp teams distribute num_teams(4) default(none) shared(a, b, c, d, num_teams) private(privatized)
```
jhdavis8, line 38: Wondering if this is compliant...
4.5: "If specified, a teams construct must be contained within a target construct. That target construct must not contain any statements or directives outside of the teams construct."

5.0: "A teams region can only be strictly nested within the implicit parallel region or a target region. If a teams construct is nested within a target construct, that target construct must contain no statements, declarations or directives outside of the teams construct."

spophale: I don't understand the comment. We are not using the target construct. IF it was nested then the 5.0 part would apply. This is a standalone teams construct.

spophale: @jhdavis8 @tmh97 review complete ?

jhdavis8: Ah I see, I think I was mistinterpreting the spec to say standalone teams was not permitted. Will approve.

spophale:
Ping @tmh97

PR #145, opened 2020-05-13
Add 5.0 test for requires dynamic allocators

jhdavis8:
This is a test for the requires dynamic allocators clause. With the clause given, the test should be able to use OpenMP allocator routines inside of a target region. This test checks that this is possible using code based on an OpenMP 5.0 example of allocator usage.

Checking is done on the device since the allocator is both created and destroyed ont he device.

Right now no Summit compiler will compile this, since I the feature is not supported by any of them. Therefore I am less certain this code is correct than usual. Any feedback is much appreciated.


spophale:
Did you try executing it on Summit ?

jhdavis8:
Yes, it won't compile on any of the available compilers. GCC gives a message saying the feature is unimplemented and Clang/XLC give errors for all the memory management calls inside target.

tests/5.0/requires/test_requires_dynamic_allocators.c
```diff
@@ -0,0 +1,64 @@
...
+#define N 1024
+
+int test_requires() {
+  int errors = 0;
+
+#pragma omp target map(from: errors)
+  {
+    int* x;
+    omp_memspace_t x_memspace = omp_default_mem_space;
+    omp_alloctrait_t x_traits[1] = {omp_atk_alignment, 64};
```
jhdavis8, line 30: Remove this for now. Don't test alignment.

jhdavis8: I looked into this a bit further to see exactly where this came from. Allocator traits like alignment are explained on page 153, but the name `omp_atk_alignment` comes from page 406. It is a member of the allocator keys enum. `omp_alloctrait_t` is the type `omp_init_allocator` expects as its third argument.

tests/5.0/requires/test_requires_dynamic_allocators.c
```diff
@@ -0,0 +1,64 @@
...
+    omp_allocator_handle_t x_alloc = omp_init_allocator(x_memspace, 1, x_traits);
+
+    x = (int *) omp_alloc(N*sizeof(int), x_alloc);
+
+    // Error if x is not 64-byte aligned (address should be divisible by 64)
+    OMPVV_TEST_AND_SET_VERBOSE(errors, ((intptr_t)(x)%64 != 0));
```
jhdavis8, line 36: Based on summit results it is possible Clang and XLC do not define the intptr_t type. I might look into a different way to test this.

jhdavis8:
Will not compile on the Cori build of Clang 11.

PR #149, opened 2020-05-19
Minor fixes to detect GCC and clang versions

tob2:
Followup to Pull Req. #132
	* sys/make/make.def (C_VERSION, CXX_VERSION, F_VERSION): Use
	'-n 1' instead of '-1' for 'head' for better POSIX compatibility;
	strip-off date and git version for GCC.

### DIFF
--- a/sys/make/make.def
+++ b/sys/make/make.def
@@ -81,7 +81,7 @@ ifeq ($(CC), gcc)
   CFLAGS += -O3 -std=c99 -fopenmp $(COFFLOADING)
   CLINK = gcc
   CLINKFLAGS += -O3 -fopenmp $(COFFLOADING)
-  C_VERSION = echo "$(shell gcc --version | head -1 | sed 's/.* \([0-9]*\.[0-9]*\.[0-9]*\)/\1/g')"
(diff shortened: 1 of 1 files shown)


tob2:
As mentioned in in #132:
* POSIX's head only supports `head [−n number]` (and the documentation to the GNU tools also only documents the `-n` variant)
* GCC version might not only have `x.y.z` version numbers but more. Hence, everything which comes after the third [0-9]* is now stripped off.

I have here (all distribution GCC builds):
```
* openSUSE tumbleweed:
gcc-7 (SUSE Linux) 7.5.0
gcc-9 (SUSE Linux) 9.3.1 20200406 [revision 6db837a5288ee3ca5ec504fbd5a765817e556ac2]
gcc-10 (SUSE Linux) 10.1.1 20200507 [revision dd38686d9c810cecbaa80bb82ed91caaa58ad635]
* Debian stretch:
gcc (Debian 6.3.0-18+deb9u1) 6.3.0 20170516
* Ubuntu 14.04.4 LTS/Trusty:
gcc (Ubuntu 4.8.4-2ubuntu1~14.04.3) 4.8.4
```
With the current version (before my patch), *_VERSION will include the date stamp or date + git version, which works but is rather lengthy.

Issue #148, opened 2020-05-18
OMPVVLIB in the C tests build step fails with GCC 9.2.0 


mcytowski:
**Describe the bug**
We are building and running the benchmark on Linux system with GCC 9.2.0 and V100. Compilation of most of the C tests is successful, however during runtime all of the C tests fail with the following error:
`libgomp: Cannot map target functions or variables (expected X-1, have X)`

**Test that it applies to**
All C tests.

**To Reproduce**
Clone the current repository, compile and run with GCC 9.2.0.

**Expected behavior**
Removing $OMPVVLIB from the build rules for C files in Makefile seems to be solving this issue:
`-       -$(QUIET)$(call loadModules,$(C_COMPILER_MODULE)) $(CCOMPILE) $(VERBOSE_MODE) $(DTHREADS) $(DTEAMS) $(HTHREADS) $< -o $(BINDIR)/$(notdir $@) $(OMPVVLIB) $(if $(LOG),$(RECORD)$(notdir $(@:.o=.log))\`
`+       -$(QUIET)$(call loadModules,$(C_COMPILER_MODULE)) $(CCOMPILE) $(VERBOSE_MODE) $(DTHREADS) $(DTEAMS) $(HTHREADS) $< -o $(BINDIR)/$(notdir $@)  $(if $(LOG),$(RECORD)$(notdir $(@:.o=.log))\`

**Compiler**
GCC 9.2.0

**Accelerator hardware**
V100

spophale:
Are you using our Makefile ?

I did a quick test on Summit (also V100) with gcc 9.1.0, I don't see the same behavior.
Make command : 
make CC=gcc SOURCES=a.c all

Some tests pass, some fail but none report the runtime error you see. There is a possibility that this could be specific to 9.2.0.

josemonsalve2:
Swaroop, @tmh97 and I also noticed the same issue with our UD systems. It happened after we started linking the static library to all the tests. This is certainly a bug of some sort in the compiler. But in the meantime we should provide a way to disable the static lib. 

@mcytowski I was thinking this could be an issue with the newlib version that we are using, but I am not sure. It only seems to happen with C

spophale:
What is the version of your gcc @josemonsalve2 @tmh97 ?
I will verify if my repo is updated.

josemonsalve2:
7.1 7.2 7.3 8.1 8.2 9.1 9.2 
[...]

PR #153, opened 2020-05-27
Add LINK_OMPVV_LIB flag to enable static lib and test of static lib

jhdavis8:
By adding LINK_OMPVV_LIB=1 when using the Makefile, linking the static library can now be disabled. This is a quick patch for issue #148.

This will also remove the QMCPACK target static lib test from the list of sources when not using the SOURCES= flag.

Tested and working on Summit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -138,7 +138,11 @@ endif
 
 ifeq "$(SOURCES)" ""
 # Getting all the source files in the project
+ifdef LINK_OMPVV_LIB
 SOURCES_C := $(shell find $(CURDIR)/tests/$(OMP_VERSION) -name *.c)
+else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -238,8 +242,12 @@ endif
...
 ##################################################
 
-OMPVVLIB=-L$(CURDIR)/ompvv -lompvv
-OMPVVLIB_DEP=$(CURDIR)/ompvv/libompvv.a
+OMPVVLIB=
+OMPVVLIB_DEP=
+ifdef LINK_OMPVV_LIB
+	OMPVVLIB=-L$(CURDIR)/ompvv -lompvv
+	OMPVVLIB_DEP=$(CURDIR)/ompvv/libompvv.a
+endif
```
jhdavis8, line 250: try linking only to the test that uses it

jhdavis8: The library is now only linked to the test that uses it. Changes are all tested and working.

PR #154, opened 2020-05-27
Add 5.0 test for mapping order

tmh97:
Tested on Obiwan with aomp, passes without delete clause. Need to decide whether to rename this test test_target_mapping_to_before_delete.c or to make this a very large test. Also, is /target appropriate directory?

### DIFF
--- a/tests/5.0/target/test_target_mapping_order.c
+++ b/tests/5.0/target/test_target_mapping_order.c
@@ -1,0 +1,72 @@
+//===--- test_target_mapping_order.c---------------------------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// The description of the map clause was modified to clarify the mapping order when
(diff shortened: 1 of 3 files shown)


tests/5.0/target/test_target_mapping_order.c
```diff
@@ -0,0 +1,98 @@
...
+  } member; 
+
+  member.var = 1;
+  
+  for (i = 0; i < N; i++) { 
+    a[i] = i;
+    member.b[i] = i;
+  }
+
+#pragma omp target map (alloc: scalar, a, member) (to: scalar, a, member)
```
tmh97, line 79: add extra map clause

tmh97:
Tested on summit w/ newest xlc : error: variable already marked as mapped in current construct
#pragma omp target map (tofrom: scalar, c, member) map (from: x, y, z) map (delete: scalar, c, member)

PR #155, opened 2020-05-27
Add test of nested functions with declare target

jhdavis8:
This is a simple test of using declare target on a host function that calls another host function. According to 5.0 specification, the inner function, while not explicitly declared on the target, should be treated as if it were. 

The functions I test just increment the passed-in value and return it. They are tested in a simple target region and the result is mapped out and checked.

This test is passing on Summit gcc but not clang or xlc.

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_nested.c
+++ b/tests/5.0/declare_target/test_declare_target_nested.c
@@ -2,7 +2,11 @@
 //
 // OpenMP API Version 5.0 Nov 2018
 //
-//
+// Test of declare target on a host function that calls another host
+// function. According to 5.0 specification, the inner function, while not
+// explicitly declared on the target, should be treated as if it was. The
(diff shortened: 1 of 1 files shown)


tests/5.0/declare_target/test_declare_target_nested.c
```diff
@@ -0,0 +1,51 @@
...
+// declared functions increment the passed-in value and return it. They are
+// tested in a simple target region and the result is mapped out and checked.
+//
+//===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
```
jhdavis8, line 17: Try with functions in header file or classes?

tmh97:
Still test on summit

jhdavis8:
Will not compile on Cori Clang 11 build.

PR #157, opened 2020-05-28
Mutiple fixes

simoatze:
Hi guys,

a couple of small fixes:

- Two of the tests declare the "errors" variable within the CONTAINS scope but they are never initialized.
- Used wrong index variable within for-loop
- Inner loop variable should be privatized as it is shared within the parallel region.

Let me know if you agree.

Thanks!
Simone

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_reduction_bitand.F90
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_reduction_bitand.F90
@@ -17,8 +17,6 @@ PROGRAM test_target_teams_distribute_device
   USE ompvv_lib
   USE omp_lib
   implicit none
-  INTEGER :: errors
-  errors = 0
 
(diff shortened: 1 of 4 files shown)


jhdavis8:
It is likely the 5.0 versions of the reduction tests have the same bug. I will check later today.

spophale:
Thanks @simoatze !

tests/4.5/target_teams_distribute/test_target_teams_distribute_reduction_bitand.F90
```diff
@@ -33,6 +33,7 @@ INTEGER FUNCTION test_bitand()
     LOGICAL:: tested_true, tested_false
     REAL(8):: false_margin
     result = 0
+    errors = 0
```
spophale, line 36: "errors" from main program not used. Remove 20-21.

PR #158, opened 2020-06-03
Adding 5.0 tests for use_device_ptr and use_device_addr.

spophale:


tests/5.0/target_data/test_target_data_use_device_ptr.c
```diff
@@ -0,0 +1,56 @@
...
+
+  for (int i = 0; i < N; ++i)
+    array_host[i] = i;
+  
+  OMPVV_TEST_OFFLOADING;
+
+#pragma omp target data map(tofrom: array_device[0:N]) use_device_ptr(array_device)
+    {
+#pragma omp target is_device_ptr(array_device) map(tofrom: array_host[0:N])
+      {
```
tob2, line 38: While it is somewhat useless (outside a test case) to have use_device_ptr – and directly pass it to is_device_ptr, that seems to be fine. Still, one could place a `int *ptr_array_device = array_device;` in `target data` and pass this to `target` as `is_device_ptr`.

tests/5.0/target_data/test_target_data_use_device_addr.c
```diff
@@ -0,0 +1,58 @@
...
+
+
+  for(int i = 0; i < N; i++)
+    host_array[i] = i;
+  
+  OMPVV_TEST_OFFLOADING;
+
+#pragma omp target data map(to: device_data) use_device_addr(device_data)
+    {
+#pragma omp target map(tofrom: host_data, host_array)
```
tob2, line 34: This testcase does not really use "use_device_addr" – in the "target data" section, device_data (with value 0) has the address of the device copy – but it is not used.
In the "target" section, gets an implicit "firstprivate(device_data)" (a device_data is a scalar) and that firstprivatized variable (with value 0) is then used inside target.

I think you need to do something along the lines of:
```
device_data = 42;
#pragma omp target data map(to: device_data) use_device_addr(device_data)
{
  intptr_t ptr_addr = (intptr_t) &device_data;

  #pragma omp target map(to: device_data) map(tofrom:errors)
  {
      // has implicit firstprivate(ptr_addr); map needed to override implicit firstprivate(device_data)
    if ((intptr_t) &device_data != ptr_addr) ++errors;
  }

  #pragma omp target  map(tofrom:errors) // implicit: firstprivate(ptr_addr); device_data still there due to "target data"
  {
    if (*(int*) ptr_addr != 42) ++errors;
  }
```

tests/5.0/target_data/test_target_data_use_device_ptr.c
spophale: I have made changes to this test to verify ptr address is used.

tests/5.0/target_data/test_target_data_use_device_addr.c
```diff
@@ -0,0 +1,51 @@
...
+  int device_data = 14, host_data=0;
+
+  OMPVV_TEST_OFFLOADING;
+
+#pragma omp target data map(to: device_data) use_device_addr(device_data)
+    {
+        int *dev_ptr;
+        dev_ptr = &device_data;
+#pragma omp target map(to:device_data) map(tofrom: errors) map(from: host_data)
+      {
```
tob2, line 31: I think you need a `is_device_ptr(dev_ptr)` here.

Additionally, I fear the compiler misunderstands the `map(to:device_data)` by trying to map the device-address `device_data` instead of the host-address one. At least when trying to compile it with GCC it fails for `&device_data != dev_ptr` and `device_addr` has an odd address.

I think you need something like:

```
int *dev_ptr;
// ↓ map data – could be either 'target enter/exit data' as here or another 'target data' encompassing everything
#pragma omp target enter data map(to:device_data)
#pragma omp target data use_device_addr(device_data)
{
         dev_ptr = &device_data;
}
// ↓ this target is outside the 'use_device_addr' (!); additionally 'is_device_ptr' has been added.
#pragma omp target is_device_ptr(dev_ptr) map(to:device_data) map(tofrom: errors) map(from: host_data)
      {
           if(&device_data != dev_ptr)
…

// At the very end: be kind and clean up  ('release:' would also do)
#pragma omp target exit data map(delete: device_data)
```

tests/5.0/target_data/test_target_data_use_device_addr.c
```diff
@@ -0,0 +1,51 @@
...
+          if(&device_data != dev_ptr)
+            errors++;
+        
+      } // end target
+
+#pragma omp target map(from: host_data)
```
tob2, line 37: Here, I think another `is_device_ptr(dev_ptr)` has to be added.

tests/5.0/target_data/test_target_data_use_device_ptr.c
tob2: Looks good to me.

tests/5.0/target_data/test_target_data_use_device_addr.c
spophale: Agree.

tests/5.0/target_data/test_target_data_use_device_addr.c
spophale: It will be interesting to find out which address it sees for 'device_addr'

PR #159, opened 2020-06-03
Add test for declare variant directive

jhdavis8:
Partially derived from the 5.0 examples document, this test checks the declare variant clause (with the match(context-selector-specification)). It creates two variants of a simple base function for use in target and parallel regions, and tests that the variants are used correctly.

This test does not compile on any Summit compiler due to lack of support for the declare variant feature.


jhdavis8:
Will not compiler with Cori Clang 11 build.

PR #160, opened 2020-06-03
New 5.0 test for nested declare target

tmh97:
Based around 5.0 examples (4.10.1 and 4.10.3). Still needs proper testing on summit


tests/5.0/declare_target/test_declare_target_nested.c
```diff
@@ -0,0 +1,74 @@
...
+
+int main () {
+
+  int errors = 0;
+  
+  //initalize arrays on host
+  for (i = 0; i < N; i++) {
+    a[i] = i;
+    b[i] = 2i;
+    c[i] = 3i;
```
jhdavis8, line 67: Think this should be 2*i and 3*i

jhdavis8:
We will have to rename my test since these tests have the same name.

tmh97:
Tested on summit with newest cuda/xlc, error: unexpected OpenMP directive '#pragma omp end declare target' #pragma omp end declare target
                        ^

PR #164, opened 2020-06-17
Adding test for atomic construct.

spophale:


jhdavis8:
Is this test ready to merge?

spophale:
Yes, just tested with gcc and llvm. Works as expected.

PR #165, opened 2020-06-17
Imperfect loop collapse

spophale:
Adding test adapted from the examples and contributed by LLNL.


spophale:
I was able to compile and run with llvm 11 on Summit.

Issue #168, opened 2020-06-18
Exit call in Fortran tests causes random exit codes with one version of the XLF compiler


colleeneb:
Mentioning an issue I ran into with an XL compiler [1] for the record.

In ompvv.F90, there is a use of "exit". Exit is a gfortran extension, not part of fortran standard and not always portable. E.g, this apparently doesn't work properly on some version of XL Fortran [1], like the one I have :). But you guys run on Summit and it's ok there, though?

Basically when I run the Fortran tests with the XL compiler [1], I get incorrect fail/pass rates, since the "exit" call is returning random exit codes. An example is below. note it passes on the device but has totally random exit codes. If I rerun, I get different exit codes for the test, too.

```
 running: bin/declare_target_module.F90.FOR.run 
/home/bertoni/tmp/xl/sollve_vv/sys/scripts/run_test.sh bin/declare_target_module.F90.o 1 2>&1 | tee -a logs/declare_target_module.F90.log && echo "PASS" > _ompvv_temp_result_.exitstatus.1307 || echo "FAIL" > _ompvv_temp_result_.exitstatus.1307
declare_target_module.F90.o: FAIL. exit code: 200
declare_target_module.F90.o:                                                                                                                                                                                                                                        
[OMPVV_INFOMSG declare_target_module.F90:40] Test is running on device                                                                                                                                                                                              
[OMPVV_INFOMSG ompvv.F90:251] Test ran with no errors                                                                                                                                                                                                               
[OMPVV_RESULT declare_target_module.F90] Test passed on the device.
``` 

Because of this, I'd suggest replacing `#define OMPVV_REPORT_AND_RETURN()` in `ompvv.F90` with portable 
```
#define OMPVV_REPORT_AND_RETURN() if( report_and_set_errors(__FILENAME__) .ne. 0) error stop EXIT_FAILURE
```
Or if you only call exit() when there's an error, then you can avoid this issue, too.
```
#define OMPVV_REPORT_AND_RETURN() if( report_and_set_errors(__FILENAME__) .ne. 0) call exit(EXIT_FAILURE)
```

However, I haven't had this issue with any other compiler, so maybe it's not worth changing. I still wanted to report it just in case, though.

[1]

```
$ xlf -qversion
IBM XL Fortran for Linux, V16.1.1 (Community Edition)
Version: 16.01.0001.0003
```

colleeneb:
To try to be super clear, this is the issue:
```
[...]

PR #175, opened 2020-07-15
 Adding in STOP to handle exit codes for Fortran tests 

colleeneb:
This contains a proposed solution to Issue #168, since it has been bothering me when running with IBM compilers. (In general, I think using exit codes to see if something passed or failed is really useful.)

The PR changes  `ompvv/ompvv.F90` and `tests/4.5/offloading_success.F90 ` to use `stop [number]` when the test fails. As mentioned in Issue #168, `call exit()` isn't part of the Fortran standard, and IBM doesn't seem to have an extension for it. So with IBM compilers, the tests will pass but have weird exit codes. Using `stop` is a solution.
One difference between the original code and this is that here if the run is successful, it exits by itself, (not `call exit(0)`).  

`tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_bitand.F90 ` I changed since the I think the syntax was wrong. Left as is, we get a compile-time error like:
```
"/home/bertoni/tmp/sollve_vv/tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_bitand.F90", line 83.5: 1515-019 (S) Syntax is incorrect.
"/home/bertoni/tmp/sollve_vv/tests/5.0/target_teams_distribute/test_target_teams_distribute_reduction_bitand.F90", line 83.68: 1513-061 (S) Actual argument attributes do not match those specified by an accessible explicit interface. 
```
I changed it to match `test_target_teams_distribute_reduction_bitor.F90`, feel free to adjust if this is wrong. Or I can submit it as another PR.

Alternatively, if you know anyone on the Fortran standard committee, please lobby them to add `call exit(error_code)` to the standard! ;)

### DIFF
--- a/tests/4.5/offloading_success.F90
+++ b/tests/4.5/offloading_success.F90
@@ -20,6 +20,6 @@ implicit none
     print*, "Target region executed on the device"
   END IF
 
-  CALL EXIT(errors)
+  if( errors .ne. 0 ) stop 1
 
 END PROGRAM offloading_success
--- a/ompvv/ompvv.F90
+++ b/ompvv/ompvv.F90
@@ -85,9 +85,9 @@
 #define OMPVV_SET_ERRORS(err) call set_errors(err)
 
 ! Macro for correct exit code
-#define OMPVV_RETURN(err) CALL EXIT(MERGE(EXIT_SUCCESS, EXIT_FAILURE, err == 0))
+#define OMPVV_RETURN(err) if( err .ne. 0) stop EXIT_FAILURE
 
-#define OMPVV_REPORT_AND_RETURN() CALL EXIT(MERGE(EXIT_SUCCESS, EXIT_FAILURE,  report_and_set_errors(__FILENAME__) == 0))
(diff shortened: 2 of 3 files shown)


tests/4.5/offloading_success.F90
```diff
@@ -20,6 +20,6 @@ PROGRAM offloading_success
     print*, "Target region executed on the device"
   END IF
 
-  CALL EXIT(errors)
+  if( errors .ne. 0 ) stop 1
```
spophale, line 23: why not use error stop here ?

colleeneb: Error stop should be fine, too. I can change to that if you want. 

From the 2018 Fortran standard (https://j3-fortran.org/doc/year/18/18-007r1.pdf), if `stop` is given a constant number, it should return that number as the exit code. `Error stop` should do the same. The difference seems to be only when `stop` or `error stop` are used without arguments.

jhdavis8:
Looks good! Thanks @colleeneb !

Issue #185, opened 2020-07-31
test_target_mapping_before_delete.c: Expected-value check is wrong


tob2:
Looking at `tests/5.0/target/test_target_mapping_before_delete.c`, a colleague pointed out that the following does not make sense. **Note the  '20' vs. '30' in:**

```c
int to_from_before_delete(){
…
  int scalar = 30;
…
#pragma omp target map (tofrom: scalar, c, member) map (from: x, y, z) map (delete: scalar, c, member)
…
    x = scalar;
…
  OMPVV_TEST_AND_SET_VERBOSE(errors, x != 20);
```

PR #186, opened 2020-08-03
test_target_mapping_before_delete.c: Fix expected value

tob2:
Issue #185
* tests/5.0/target/test_target_mapping_before_delete.c
  (to_from_before_delete): Check with the same value the 'x' value
  as the the one has been used for 'scalar' before 'x = scalar'.


tob2:
Untested – but still obvious: Line 64 is " int scalar = 30;" and 'x' and 'scalar' only appear in assignments – hence, 'x' must be 30 and not 20.

PR #192, opened 2020-09-10
Makefile: Make find more robust, silence git error

tob2:
* Makefile (SOURCE_C, SOURCE_CPP): Quote -name arg of find.
	* sys/make/make.def (OMPVV_IS_GIT): Redirect errors to /dev/null.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,11 +139,11 @@ endif
 ifeq "$(SOURCES)" ""
 # Getting all the source files in the project
 ifdef LINK_OMPVV_LIB
-SOURCES_C := $(shell find $(CURDIR)/tests/$(OMP_VERSION) -name *.c)
+SOURCES_C := $(shell find $(CURDIR)/tests/$(OMP_VERSION) -name "*.c")
 else
-SOURCES_C := $(shell find $(CURDIR)/tests/$(OMP_VERSION) ! -name qmcpack_target_static_lib.c -name *.c)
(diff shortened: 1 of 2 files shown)


tob2:
@jhdavis8, @tmh97, @spophale – please review.

Background to the two changes:
- When not quoting `find`'s `-name` argument, `*.c` or `*.cpp` files from the main directory get picked up. Those can exist either as leftovers from manual testing/playing around – or as here when doing remote testing; in our case, the wrapper copies the files onto a remote host, which happens to be the into the current directory which is mounted on both systems.
_In any case, a subsequent run will fail as the `*.c`/`*.cpp` name gets expanded and 'find' fails. (As here for a VERSION=4.5 run followed by VERSION=5.0 run.)_ — Note: For Fortran, the argument is already properly quoted.

- The other issue shows up in the log as: `fatal: Not a git repository (or any parent up to mount point ...)` / 
`Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).` – which looks ugly.
Thus, redirecting/silencing the stderr output for the `OMPVV_IS_GIT` check makes sense.

tob2:
First patch PING – @jhdavis8, @tmh97, @spophale – please review. Thanks!

spophale (approved):

tmh97 (approved):

Issue #187, opened 2020-08-03
test_target_data_use_device_addr: nested 'map' inside 'data target use_device_addr'


tob2:
`tests/5.0/target_data/test_target_data_use_device_addr.c`: Invalid nesting of `map(to: device_data)` in `target data use_device_addr(device_data)`?

Follow-up to PR #158 – and as already remarked there, I believe that  `map(to: device_data)` inside a `omp target data` block with `use_device_addr(device_data)` is invalid.

At least I expect it to fail (unless `required unified_shared_memory` [implicitly or explicitly]) as the compiler will use the device address inside `target data` due to `use_device_addr` and then try to use the device address as host address for mapping, which will fail. – _For sure, with GCC, it **does fail** when tested with AMD GCN offloading._ 

Namely, the current code has in [`tests/5.0/target_data/test_target_data_use_device_addr.c`](https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.0/target_data/test_target_data_use_device_addr.c#L26)
```C
...
#pragma omp target data map(to: device_data) use_device_addr(device_data)
    {
        int *dev_ptr;
        dev_ptr = &device_data;
#pragma omp target map(to:device_data) map(tofrom: errors) map(from: host_data) is_device_ptr(dev_ptr)
      {
          if(&device_data != dev_ptr)
            errors++;
...
```

**Suggested diffs.**

I am not sure whether I like (A) or (B) better. Once could also modify the testcase such that both variants are used.

**(A)** Avoiding 'map(to: device_data)' with use_device_addr.
```diff
--- a/tests/5.0/target_data/test_target_data_use_device_addr.c
+++ b/tests/5.0/target_data/test_target_data_use_device_addr.c
@@ -23,10 +23,13 @@ int main() {
 
   OMPVV_TEST_OFFLOADING;
 
-#pragma omp target data map(to: device_data) use_device_addr(device_data)
+#pragma omp target data map(to: device_data)
     {
-        int *dev_ptr;
+      int *dev_ptr;
+#pragma omp target data use_device_addr(device_data)
+      {
         dev_ptr = &device_data;
+      }
 #pragma omp target map(to:device_data) map(tofrom: errors) map(from: host_data) is_device_ptr(dev_ptr)
       {
           if(&device_data != dev_ptr)
```

**(B)** Or – as 'device_data' is already mapped. (Even better, but will be TR9/OpenMP 5.1: `has_device_addr` instead of `is_device_ptr`.)

[...]

PR #193, opened 2020-09-10
Issue 187: test_target_data_use_device_addr: nested 'map' inside 'dat…

tob2:
…a target use_device_addr'

Issue #187

Unless 'requires unified_shared_memory' is fulfilled, a variable (or list-item)
which was in a use_device_addr of a data block cannot be directly used in an
enclosed target section; it would work with OpenMP TR9's/5.1's has_device_addr,
but not with 5.0. Hence, use a separate 'target data' for use_device_addr.

* tests/5.0/target_data/test_target_data_use_device_addr.c:
	Move 'use_device_data' usage into a separate 'omp target data'.

### DIFF
--- a/tests/5.0/target_data/test_target_data_use_device_addr.c
+++ b/tests/5.0/target_data/test_target_data_use_device_addr.c
@@ -23,10 +23,13 @@ int main() {
 
   OMPVV_TEST_OFFLOADING;
 
-#pragma omp target data map(to: device_data) use_device_addr(device_data)
+#pragma omp target data map(to: device_data)
     {
         int *dev_ptr;
-        dev_ptr = &device_data;
+#pragma omp target data use_device_addr(device_data)
+      {
+          dev_ptr = &device_data;
+      }
 #pragma omp target map(to:device_data) map(tofrom: errors) map(from: host_data) is_device_ptr(dev_ptr)
       {
           if(&device_data != dev_ptr)


jhdavis8 (approved):

spophale:
Pg 162 lines 24-27 for target data say:
"If one or more of the use_device_ptr or use_device_addr clauses and one or more map clauses are present on the same construct, the address conversions of use_device_addr and use_device_ptr clauses will occur as if performed after all variables are mapped according to those map clauses."

tob2:
> Pg 162 lines 24-27 for target data say:
> "If one or more of the use_device_ptr or use_device_addr clauses and one or more map clauses are present on the same construct, the address conversions of use_device_addr and use_device_ptr clauses will occur as if performed after all variables are mapped according to those map clauses."

I concur, but it does not help with the problem – which is not with the `omp target data` but with the enclosed `omp target`. In any case, the following works perfectly well:
```C
#pragma omp target data map(a1,p1) use_device_addr(a1,a2) use_device_ptr(p1,p2) map(a2,p2)
 {
  foo (a1, a2, p1, p2);

  #pragma omp target is_device_ptr(p1,p2)
    {  /* use p1 and p2 */ ... }
 }
```
Namely, a1/a2/p1/p2 are first mapped (`tofrom`) and then for a1/a2, the following applies:

_"If a list item that appears in a `use_device_addr` clause has corresponding storage in the device data environment, references to the list item in the associated structured block are converted into references to the corresponding list item. If the list item is not a mapped list item, it is assumed to be accessible on the target device. Inside the structured block, the list item has a device address and its storage may not be accessible from the host device. [...]"_

And for p1/p2, the following applies:

_"If a list item that appears in a use_device_ptr clause is a pointer to an object that is mapped to the device data environment, references to the list item in the associated structured block are converted into references to a device pointer that is local to the structured block and that refers to the device address of the corresponding object. [...]"_

And the `is_device_ptr` tells the compiler that p1/p2 don't need to be mapped but already contain a device address (_"The is_device_ptr clause indicates that its list items are device pointers. Inside the target construct, each list item is privatized and the new list item is initialized to the device address to which the original list item refers. [...]"_)

The problem is that your code does:
```C
#pragma omp target data map(a1) use_device_addr(a1,a2) map(a2)
 {
  #pragma omp target  /* what to put here? */ 
    {  /* uses a1, a2  */ ... }
 }
```

Namely, which clause should be at the `target` construct? Let's try:
* `map` – that's not what we want - as inside the "associated structured block" the references are converted "to references to the corresponding list item" – hence, "The map clause specifies how an original list item is mapped from the current task’s data
environment to a corresponding list item in the device data environment of the device" does not make sense as we no longer reference the list item in the "task's data environment".
* `is_device_ptr` – well,  "device pointers" sound fine and we nearly there but for the testcase under discusstion, the following is not fulfilled: "A list item that appears in an is_device_ptr clause must have a type of pointer, array,
reference to pointer or reference to array." (For C++, for C without "reference to ...".) 

The following sounds very nice:
* `has_device_addr` – _"The has_device_addr clause indicates that its list items already have device addresses and
therefore they may be directly accessed from a target device. [...]"_

However, the big caveat is that this is not even in the current Technical Report (TR) 8 but only in the draft of TR 9. Assuming, the code should comply with OpenMP 5.0 and not TR9/OpenMP 5.1, that's not a solution, either.

Thus, I ended up with the version I have – alternatively, one could use is_device_ptr, but then one needs a pointer (array or reference) and as the test case is labelled "test_target_data_use_device_addr", I thought that the current solution makes more sense. — Better suggestions are welcome, however.

tob2:
@spophale – did the comment address your question? If not, what is still unclear?
@tmh97 – *review ping*

spophale (approved):

PR #197, opened 2020-09-15
F90 5.0 tests for teams

tmh97:
Test with xlf.

### DIFF
--- a/tests/5.0/teams/test_teams.F90
+++ b/tests/5.0/teams/test_teams.F90
@@ -14,6 +14,7 @@
 #define K 2
 
 PROGRAM test_teams
+ 
   USE iso_fortran_env
   USE ompvv_lib
   USE omp_lib
@@ -26,7 +27,7 @@ PROGRAM test_teams
   OMPVV_REPORT_AND_RETURN()
 
 CONTAINS
-  INTEGER FUNCTION teams()
+    INTEGER FUNCTION teams()
     CHARACTER(len=300):: infoMsg
     INTEGER,DIMENSION(N):: num_teams, num_threads, a, b
     INTEGER,DIMENSION(K):: errors
@@ -59,16 +60,16 @@ is inconsistent with the thread limit that has been set.")
 below one.")
 
     DO x = 2, num_teams(1)
-      IF (num_teams(x) .ne. num_teams(x-1))
+      IF (num_teams(x) .ne. num_teams(x-1)) THEN
         errors(1) = errors(1) + 1
       END IF
-      IF (num_threads(x) .ne. num_threads(x-1))
+      IF (num_threads(x) .ne. num_threads(x-1)) THEN
         errors(2) = errors(2) + 1
       END IF
     END DO
 
-  teams = errors(1) + errors(2)
-  END FUNCTION teams
+    teams = errors(1) + errors(2)
+    END FUNCTION teams
 END PROGRAM test_teams 
 
 
--- a/tests/5.0/teams/test_teams.F90
+++ b/tests/5.0/teams/test_teams.F90
@@ -49,17 +49,13 @@ CONTAINS
     num_threads(omp_get_team_num()) = omp_get_num_threads()
     !$omp end teams 
     
-    OMPVV_WARNING_IF(num_teams(1) .eq. 1, "Test operated with one team,
-num_teams requested is inconsistent with this result")
+    OMPVV_WARNING_IF(num_teams(1) .eq. 1, "Test operated with one team, num_teams requested is inconsistent with this result")
 
-    OMPVV_ERROR_IF(num_teams(1) .le. 1, "omp_get_num_teams() reported a
-value less than one.")
+    OMPVV_ERROR_IF(num_teams(1) .le. 1, "omp_get_num_teams() reported a value less than one.")
 
-    OMPVV_WARNING_IF(num_threads(1) .eq. 1, "Team 0 reported only 1
-thread. This is inconsistent with the thread limit that has been set.")
+    OMPVV_WARNING_IF(num_threads(1) .eq. 1, "Team 0 reported only 1 thread. This is inconsistent with the thread limit that has been set.")
 
-    OMPVV_ERROR_IF(num_threads(1) .le. 1, "omp_get_num_threads() reported a value
-below one.")
+    OMPVV_ERROR_IF(num_threads(1) .le. 1, "omp_get_num_threads() reported a value below one.")
 
     DO x = 2, num_teams(1)
       IF (num_teams(x) .ne. num_teams(x-1)) THEN
--- a/tests/5.0/teams/test_teams.F90
+++ b/tests/5.0/teams/test_teams.F90
@@ -1,0 +1,67 @@
+!===--- test_teams.F90 -----------------------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test uses the teams directive on host and verifies the 
+! requested number of teams with requested number of threads were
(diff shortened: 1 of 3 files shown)


tmh97:
1. test_team_default_shared.F90 -> Passes with no errors on device (should be running on host instead of device)
2. test_teams.F90 -> Unexpected END statement Error at lines ~ 72/73
3. test_teams_distribute.F90 -> SEG FAULTS with gcc 9, Runs with gcc 10 but fails on device.

spophale:
We need to dig a bit.

tmh97:
> We need to dig a bit. Does the commit fix all the bugs?
@spophale Those issues I commented about a few hours ago still need to be solved. The commit fixed some other syntax issues.

tests/5.0/teams/test_teams.F90
```diff
@@ -0,0 +1,75 @@
...
+    INTEGER:: x
+    errors(1) = 0
+    errors(2) = 0
+
+    DO x = 1, N
+      num_teams(x) = -99
+      num_threads(x) = -99
+    END DO
+
+    !$omp teams num_teams(N) thread_limit(N)
```
jhdavis8, line 43: Same thing, can use OMPVV_NUM_TEAMS_DEVICE and OMPVV_NUM_THREADS_HOST here. Altho because it is a standalone teams we may need a OMPVV_NUM_TEAMS_HOST macro? I can add this if we think it is needed.

jhdavis8 (changes requested):

tmh97:
**update**

1.    test_team_default_shared.F90 -> Passes with no errors on device
2.    test_teams.F90 -> Passes on device, issue with thread count
3.    test_teams_distribute_default_none.F90 -> SEG FAULTS with gcc 9, Runs with gcc 10 but fails on device. -> Runs on device, fails due to errors=2

tests/5.0/teams/test_teams_distribute_default_none.F90
```diff
@@ -0,0 +1,97 @@
...
+CONTAINS
+  INTEGER FUNCTION default_none1()
+    INTEGER :: a(N), b(N), c(N), d(N)
+    INTEGER :: privatized, x, y, errors, num_teams, share
+    errors = 0
+    share = 0
+
+    DO x = 1, N
+       a(x) = 1
+       b(x) = x
```
tob2, line 43: Here a local variable in 'default_none1' is set to `b = [(x, x=0,N)]` — which is fine but ... (continued below)

tests/5.0/teams/test_teams_distribute_default_none.F90
```diff
@@ -0,0 +1,97 @@
...
+
+    DO x = 1, N
+       OMPVV_TEST_AND_SET(errors, (d(x) .ne. (1 + x)*2*x))
+    END DO
+
+    default_none1 = errors
+  END FUNCTION default_none1
+
+  INTEGER FUNCTION default_none2()
+     INTEGER :: b(N)
```
tob2, line 72: ... (continued from comment above to default_none1) ... and here a completely separated variable `b(N)`, which a local variable is allocatabled, probably on the stack with some random value. And .... (continued below in the same function)

tests/5.0/teams/test_teams_distribute_default_none.F90
```diff
@@ -0,0 +1,97 @@
...
+     INTEGER :: share, x, num_teams, errors
+     errors = 0
+     share = 0
+
+     !$omp teams distribute num_teams(4) default(none) shared(share,b,num_teams)
+     DO x = 1, N
+        !$omp atomic update 
+        share = share + b(x)
```
tob2, line 80: ... and here `b(x)` is used. This atomic update does: `share += uninitialized value` but somehow assumes that the local variable of `b(1:N)` of `default_none2` has something to do with the initialization of `b(1:N)` of `default_none1`.

Depending on how the variables are put on the stack, this might work by chance, but it is invalid according to the Fortran standard and as the run-time test shows it (usually) also does not work in practice.

tob2 (changes requested):
As commented inline – the value set to a local variable `b` in one subroutine does not affect the value in the other subroutine (relates to `test_teams_distribute_default_none.F90`).

tests/5.0/teams/test_teams_distribute_default_none.F90
tmh97: @tob2 thanks for catching this, will address shortly.

tmh97:
test_team_default_shared.F90 and test_teams_distribute_none.F90 are both ready and pass on device.
test_teams.F90 failing with omp_get_num_threads() reporting value below one.

tests/5.0/teams/test_teams.F90
```diff
@@ -0,0 +1,77 @@
...
+#define K 2
+
+PROGRAM test_teams
+ 
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
+  OMPVV_TEST_OFFLOADING
```
tmh97, line 23: remove check for offload

tmh97:
Testing with IBM compiler

- test_teams.F90 -> `/sw/summit/xl/16.1.1-5/xlf/16.1.1/bin/.orig/xlf_r: 1501-230 (S) Internal compiler error; please contact your Service Representative. For more information visit:`
- test_teams_distribute_default_none.F90 -> `line 48.9: 1515-179 (S) The TEAMS construct must immediately follow a TARGET directive.
`

tmh97:
Testing with GCC 10.2.0 (newest on summit)

1. test_team_default_shared.F90 -> Passes on device
2. test_teams -> Team 0 reports only 1 thread, test failed on the device (last meeting determined it is comp. bug)
3. test_teams_distribute_default_none.F90 -> Passes on device

tmh97:
Testing with XL (xlf) 16.1.1-8 (newest on summit)

1. test_team_default_shared.F90 -> Passes on device
2. test_teams.F90 -> `/sw/summit/xl/16.1.1-8/xlf/16.1.1/bin/.orig/xlf_r: 1501-230 (S) Internal compiler error; please contact your Service Representative.`
3. test_teams_distribute_default_none.F90  -> error with summit ! cannot locate library for gfortran (even tho its xlf)?

tests/5.0/teams/test_teams.F90
```diff
@@ -0,0 +1,73 @@
...
+
+  OMPVV_TEST_VERBOSE(teams() .ne. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+
+CONTAINS
+    INTEGER FUNCTION teams()
+    CHARACTER(len=300):: infoMsg
+    INTEGER,DIMENSION(OMPVV_NUM_TEAMS_DEVICE):: num_teams 
+    INTEGER,DIMENSION(OMPVV_NUM_THREADS_DEVICE):: num_threads
```
tob2, line 33: This variable goes from lower bound `1` to `8` but below, the code has:
  `num_threads(omp_get_team_num())`
The problem is that `omp_get_team_num()` returns values from `0` to `omp_get_num_threads()-1`.

Hence, either the variable has to be defined as `DIMENSION( 0 : OMPVV_NUM_THREADS_DEVICE - 1)` or it has to be used as  `omp_get_team_num() + 1`.

tmh97: Thanks Tobias! Resolved this issue with most recent commit, decided to go with`omp_get_team_num() + 1`

jhdavis8 (approved):

spophale (approved):

PR #198, opened 2020-09-15
New F90 5.0 tests for target

tmh97:


tmh97:
Tested with gfortran on summit (gcc 9.1.0 - gcc 10). Resolved obvious syntax issues, noticing remaining errors stating "Unexpected END statement". This error is persistent for both files changed in this commit. Considering additional in additional checks with a structure to make test stronger

tob2:
> noticing remaining errors stating "Unexpected END statement".

Namely, "!$omp target" requires (in Fortran) an "!$omp end target" – which is missing.

The only thing which does not require "end ..."  are loops ("omp ... do", "omp ... simd",  "omp ... parallel do simd" etc.). (And, of course, stand-alone directives like 'omp target update'.)

_I am not sure whether I mix this up with OpenACC (where it is definitely planned), but I vaguely recall some discussion regarding omitting the "!$omp end ..." if there is a BLOCK ... END BLOCK after the directive, mimicing C's '{' ... '}'._

tests/5.0/target/test_target_mapping_before_alloc.F90
```diff
@@ -0,0 +1,50 @@
...
+    INTEGER:: x, scalar, errors
+    errors = 0
+    scalar = 80
+
+    DO x = 1, N
+      a(x) = x
+    END DO
+
+    !$omp target map(alloc:scalar, a) map(to: scalar, a) 
+    IF (scalar.ne.80 .or. a(1).ne.2) THEN
```
spophale, line 44: alloc before to == alloc . Values will be undefined.

PR #199, opened 2020-09-15
F90 5.0 test for declare target

tmh97:


spophale (approved):

tmh97:
Tested on summit with gcc (9.1.0), noticing errors regarding "Unexpected END statement" at lines 52 and 53" as well as an "Unclassifiable statement" at line 32

tmh97:
Tested on summit with GCC 9.1.0, throwing error `Error: Unexpected !$OMP DECLARE TARGET statement at 29`. Test passes on host with no errors

tests/5.0/declare_target/test_declare_target_nested_functions.F90
```diff
@@ -0,0 +1,60 @@
...
+    USE iso_fortran_env
+    USE ompvv_lib
+    USE omp_lib
+    implicit none
+    INTEGER :: errors
+    errors = 0 
+
+    OMPVV_TEST_OFFLOADING
+
+    !$omp declare target to(outer_fn)
```
tob2, line 29: This location is invalid. The OpenMP spec has:

> * If the directive does not have any clause, has a device_type clause, or has clauses that
contain a function name, it must appear in the specification part of a subroutine subprogram,
function subprogram or interface body.

The specification part of `program` ends with the assignment `errors = 1`.

tob2: Actually, from that wording, it is not permitted in a `program` as this is neither a `subroutine` nor a `function` ...

tmh97:
"Passes" on device with gcc 10 right now regardless of whether the declare target or target map exist, need to look into this.

tmh97:
May need to rework this test and the original .c test to ensure that the test **fails** if the important functions run on the host, even when the OMPVV_TEST_OFFLOADING macro reports it was able to run code on the device initially.

tests/5.0/declare_target/test_declare_target_nested_functions.F90
```diff
@@ -0,0 +1,60 @@
...
+
+    OMPVV_TEST_VERBOSE(test_declared_functions() .ne. 0)
+
+    OMPVV_REPORT_AND_RETURN()
+
+CONTAINS  
+    INTEGER FUNCTION inner_fn(a)
+      INTEGER:: a
+      inner_fn = 1 + a
```
tmh97, line 38: if initial device, trigger error

tests/5.0/declare_target/test_declare_target_nested_functions.F90
jhdavis8: Thanks for you comment @tob2. What version of spec are you looking at? I see the following under declare target section for Version 5.0 (pg. 184, lines 7-10):

> • Any declare target directive with clauses must appear in a specification part of a subroutine subprogram, function subprogram, program or module.
> • Any declare target directive without clauses must appear in a specification part of a subroutine subprogram, function subprogram or interface body to which it applies.

By this I think it should work as long as it as moved to the specification part of the program, since it is a declare target with clauses. This is also consistent with 5.0 examples.

tob2: > Thanks for you comment @tob2. What version of spec are you looking at? I see the following under declare target section for Version 5.0 (pg. 184, lines 7-10):

I did look at https://www.openmp.org/wp-content/uploads/openmp-TR9.pdf – however, this wording is already in TR8.

> > • Any declare target directive with clauses must appear in a specification part of a subroutine subprogram, function subprogram, program or module.

...

> By this I think it should work as long as it as moved to the specification part of the program, since it is a declare target with clauses. This is also consistent with 5.0 examples.

I concur that the 5.0.1 example `declare_target.2.f90` at https://www.openmp.org/wp-content/uploads/openmp-examples-5-0-1.pdf#page=190 shows a similar use. – And that the wording of 5.0 permits it.

In any case, it has to be in a specification part – whether it can be in the one of `program` or has to be in the one of `outer_fn`.

_Side remark: without further restrictions, OpenMP 5.0, TR8+TR9 permit code which is likely not intended to work (e.g. `declare target` for a use-associated variable). Hence, without a lengthy list of restrictions, it only makes sense to have a `declare target` for a procedure in the specification part of that procedure (or in the specification part an interface block for that procedure)._

_Hence, I think, it would be best to leave it out completely for this testcase. This should have no effect on the testcase, which relies already on auto-`declare target` of `inner_fn` such that the compiler can also auto-declare `outer_fn` without the directive._

tmh97:
Addressed the issue regarding placement of the `declare target to` clause and the test ran fine. Attempted to introduce calls to omp_is_initial_device within the outer and inner functions and ran into problems.

tests/5.0/declare_target/test_declare_target_nested_functions.F90
```diff
@@ -0,0 +1,69 @@
...
+
+    OMPVV_TEST_VERBOSE(test_declared_functions() .ne. 0)
+
+    OMPVV_REPORT_AND_RETURN()
+
+CONTAINS  
+    INTEGER FUNCTION inner_fn(a)
+      INTEGER:: a
+      
+      OMPVV_ERROR_IF(omp_is_initial_device(), "Target region executed on host&
```
spophale, line 40: The long message is causing xlf to throw an error. Temp solution would be to shorten the msg.

tests/5.0/declare_target/test_declare_target_nested_functions.F90
```diff
@@ -0,0 +1,69 @@
...
+      & instead of device, test fails on the host")
+     
+       inner_fn = 1 + a
+    END FUNCTION inner_fn
+
+    INTEGER FUNCTION outer_fn(a)
+      INTEGER:: a
+      
+      OMPVV_ERROR_IF(omp_is_initial_device(), "Target region executed on host&
```
spophale, line 49: The long message is causing xlf to throw an error. Temp solution would be to shorten the msg. 
We should look into the MACRO.

spophale:
After resolving the macro issue, I get the following with xlf:
"
"test_declare_target_nested_functions.F90", line 26.29: 1514-480 (S) The entity outer_fn that appears in the DECLARE TARGET directive must be declared in the same scoping unit.
"

tmh97:
> After resolving the macro issue, I get the following with xlf:
> "
> "test_declare_target_nested_functions.F90", line 26.29: 1514-480 (S) The entity outer_fn that appears in the DECLARE TARGET directive must be declared in the same scoping unit.
> "

Noticing similar error with GCC, I removed the macros entirely and just testing with omp_is_inital_device and still noticed issues.

tmh97:
Potentially remove check for offloading and only check for is_initial_device right before the computation occurs

tmh97:
Attempted to remove macro for offloading check and move call to omp_is_initial_device() directly before computations occurs -> still noticing same errors related to having omp_is_initial_device inside a target region

tmh97:
test with llvm (check summit.def file), perform one last lookover before emailing about compiler bug.

tmh97:
compilation error for xlf 16.01 : `The entity outer_fn that appears in the DECLARE TARGET directive must be declared in the same scoping unit.
`

spophale:
@tmh97 are you ready to merge this ?

PR #201, opened 2020-09-15
test_declare_target_nested.c globals must be declared for target

### LINKED ISSUE #200: test_declare_target_nested.c globals must be declared for target
The `test_declare_target_nested.c` test declares global variables `int i, error` after the `end declare target` pragma. However, these variables are subsequently used from `int test_target()`, which is called from inside of a `target` region. In order to use these variables from the target device, they should be declared for the target first. Furthermore, since the `errors` variable is used by both the host and the target,  its value must be copied back to the host before being returned.

I also noticed that `int test_target()` is called from a target region (and therefore may be offloaded from the host), so the comment
> //confirm updated values on host

is not accurate since this check happens on the target.

jacobdweightman:
The `test_declare_target_nested.c` test declares global variables `int i, error` after the `end declare target` pragma. However, these variables are subsequently used from `int test_target()`, which is called from inside of a `target` region. In order to use these variables from the target device, they should be declared for the target first. Furthermore, since the `errors` variable is used by both the host and the target, its value must be copied back to the host before being returned.

I also noticed that `int test_target()` is called from a target region (and therefore may be offloaded from the host), so the comment

    //confirm updated values on host

is not accurate since this check happens on the target.

Fixes #200

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_nested.c
+++ b/tests/5.0/declare_target/test_declare_target_nested.c
@@ -52,7 +52,13 @@ int test_wrapper() { //wrapper for declare target function
   {
     test_target();
   }
-  #pragma omp target update from(errors)
+  #pragma omp target update from(errors, a, b, c)
+  
+  for (i = 0; i < N; i++) { //check array values on host
+    if ( a[i] != 5 || b[i] != 10 || c[i] != 15) {
+      errors++;  
+    } 
+  }
   return errors;
 }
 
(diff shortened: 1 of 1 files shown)


tob2:
@jhdavis8 — ping
@tmh97 — ping
For what it is worth, skimming the patch, it also looks fine to me.

spophale (approved):

tmh97:
Line 74 `#pragma omp target update to(a,b,c)` may not be needed as variables included within the declare target and end declare target are implicity mapped to the device data environment

tmh97:
Passes with llvm 11 on host, llvm 9 and 10 throws error `Libomptarget fatal error 1: failure of target construct while offloading is mandatory`

tmh97 (approved):

PR #202, opened 2020-09-16
New feature/target thread team macros (for issue #22)

jhdavis8:
This PR addresses issue #22 for test_target_private and test_target_firstprivate. The main fix is adding the OMPVV_NUM_THREADS_HOST where needed so the user can select these values through the Makefile, but some whitespace cleanup is also present.


spophale (approved):
Don't mind moving to depend on the header file if the thread requirements are reasonable and do not cause incorrect results across all tests.

spophale:
@tmh97 please review.

tmh97 (approved):

Issue #203, opened 2020-09-16
target_enter_data/test_target_enter_data_if.F90 has no matching if() clause


vzakhari:
**Describe the bug**

There is no matching `if(s > THRESHOLD)` clause on `omp target exit data` here: https://github.com/SOLLVE/sollve_vv/blob/master/tests/4.5/target_enter_data/test_target_enter_data_if.F90#L102

The corresponding C test is correct.

**Test that it applies to**
target_enter_data/test_target_enter_data_if.F90

**To Reproduce**

**Expected behavior**

**Compiler**

**Accelerator hardware**

PR #207, opened 2020-09-28
Adding one-line fix to address problems described in issue 203

tmh97:
Adding if clause onto the target exit data directive, required to match the functionality of the original .c test.
Test passes on device with gcc 9 on summit


jhdavis8:
Issue #203 (linking for reference)

jhdavis8 (approved):

spophale (approved):

PR #210, opened 2020-09-30
Add 5.0 test of simd inside atomic for C

jhdavis8:
This test simply checks that the simd construct can be used to avoid a race condition inside a parallel for simd construct on the device, a 5.0 feature.


jhdavis8:
Currently passes GCC only. Still needs a Fortran version, and we should also test atomic+simd in distribute constructs and target constructs.

spophale (approved):

tmh97 (approved):

PR #221, opened 2020-10-16
test_target_map_*_arrays.F90: Update comments

tob2:
* tests/4.5/target/test_target_map_program_arrays.F90: Update comments
  regarding copy out and assumptions for those tests.
* tests/4.5/target/test_target_map_subroutines_arrays.F90: Likewise.

In each program, two comments about `This value should not be copied back` were wrong (copy'n'paste from first function; first: `to:`, others: `from:` and `tofrom:`). Additionally, the code has the assumption that it does not run on the host nor runs on a shared-memory system — and that unitialized memory never happens to be 999.

I think that's fine, but I makes sense to document this in the source code. (`omp_is_initial_device` could be used to check whether it is run on the host, i.e. it is known to use shared memory, but for unified-shared memory, I have no idea how this could be checked – except trying it, but that's what the test does.)

### DIFF
--- a/tests/4.5/target/test_target_map_program_arrays.F90
+++ b/tests/4.5/target/test_target_map_program_arrays.F90
@@ -51,6 +51,8 @@
               helper_array_3d = array_3d
 
               ! This value should not be copied back
+              ! Assumes that that the target region is neither
+              ! executed on the host nor with unified-shared memory
               array_1d(:) = 0
               array_2d(:,:) = 0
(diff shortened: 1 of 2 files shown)


tob2:
@spophale , @jhdavis8, @tmh97 
Slightly boring as only a comment update, still would be nice if you could review it. Thanks.

tmh97 (approved):

jhdavis8 (approved):

PR #222, opened 2020-10-21
5.0 allocate allocators test

jhdavis8:
This is a test of a 5.0 feature, `allocate` directive with `allocator` clause. It uses a 64-byte aligned `omp_allocator` type with the allocate directive to allocate an array, which is then checked for the byte alignment with a `parallel for simd` construct and the `aligned` clause.

This test passes with Clang 11 on Summit.

I also added a small change to the gitignore so it will ignore emacs buffers (filenames ending in ~).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ ompvv/*.a
 results.json
 results.csv
 recent_online_report_tags
+*~
--- a/tests/5.0/allocate/test_allocate_allocator.c
+++ b/tests/5.0/allocate/test_allocate_allocator.c
@@ -19,7 +19,7 @@
 
 #define N 1024
 
-int test_requires() {
+int test_allocate_allocator() {
   int errors = 0;
 
   int* x;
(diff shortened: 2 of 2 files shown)


spophale (approved):

tmh97 (approved):
Looks good to go. I notice the header describes something regarding "memory can be written to in the target region" but it looks like there is not target section in this test, maybe this should be a "parallel region" instead?

tob2:
I know that this has been merged but I do not understand the testcase:
```C
  int* x;
...
#pragma omp allocate(x) allocator(x_alloc)
  x = (int *) malloc(N*sizeof(int));
...
  free(x);
```
In my understanding there are two ways to do OMP allocations:

* By calling `omp_alloc` and later `omp_free`, which both take a (in C++: optional) `omp_allocator_handle_t`
* By using the `omp allocate`

At least in my understanding of the spec, `omp allococate`'s list items have to be declared in the same scoping unit, are allocated by the `omp allocate` directive and automatically deallocated at the end of the scoping unit.

Here, it looks as if  `x` is allocated twice: once via `omp allocate` and then again via `malloc`/`free`, which do not use the OpenMP allocator.

Unless I miss something, I propose to split this into two parts: 
* One test for `omp_alloc` / `omp_free` replacing `malloc` and `free`
* Another test for `omp allocate(x[N])`

PR #225, opened 2020-10-28
LSMS Application Kernel

spophale:
Adding lsms kernel that tests collapse of 4 loops. Currently does not work with llvm 11 or 12. The kernel does not use any 5.0 features.

The first 3 and the last 3 loops collapse just fine and produce correct results on host as well as target. Still investigating if there is some incompatibility between loop1 and loop4.

### DIFF
--- a/tests/4.5/application_kernels/lsms_triangular_packing.cpp
+++ b/tests/4.5/application_kernels/lsms_triangular_packing.cpp
@@ -1,0 +1,144 @@
+/*
+ * Copyright (C) 2014, The University of Texas at Austin
+ * Copyright (C) 2014-2015, Michael Lehn
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *  - Redistributions of source code must retain the above copyright
+ *    notice, this list of conditions and the following disclaimer.
+ *  - Redistributions in binary form must reproduce the above copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *  - Neither the name of The University of Texas at Austin nor the names
+ *    of its contributors may be used to endorse or promote products
+ *    derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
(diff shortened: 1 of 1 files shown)


tests/4.5/application_kernels/lsms_triangular_packing.cpp
```diff
@@ -0,0 +1,144 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ */
+
+#include <complex>
+#include <cstdio>
+#include <ompvv.h>
+
```
tob2, line 34: Somewhere an `#include <omp.h>` is missing. (Either in this file or in `ompvv/ompvv.h`, which uses `omp_is_initial_device` – I would put it here.)

tests/4.5/application_kernels/lsms_triangular_packing.cpp
```diff
@@ -0,0 +1,144 @@
...
+         IndexType   incRowL,
+         IndexType   incColL,
+         Buffer      *p)
+{
+    OMPVV_INFOMSG("app_kernel_lsms_triangular_packing");
+    IndexType mp = (mc+MR-1) / MR;
+
+// Collapsing the whole loop nest here causes errors
+#pragma omp target teams distribute parallel for collapse(4)
+    for (IndexType j=0; j<mp; ++j) {
```
tob2, line 80: In GCC 11, it fails with:
```
lsms_triangular_packing.cpp:80:5: error: initializer expression refers to iteration variable 'j'
   80 |     for (IndexType j=0; j<mp; ++j) {
      |     ^~~
```
Which is in a function with the comment:
```c
/* Helper function called via walk_tree, to diagnose uses
   of associated loop IVs inside of lb, b and incr expressions
   of OpenMP loops.  */
```

tests/4.5/application_kernels/lsms_triangular_packing.cpp
spophale: Thanks @tob2 ! This code is not ready, I am still analyzing.

tests/4.5/application_kernels/lsms_triangular_packing.cpp
```diff
@@ -0,0 +1,144 @@
...
+// Collapsing here is fine
+//#pragma omp target teams distribute parallel for collapse(3)
+        for (IndexType j0=0; j0<MR; ++j0) {
+            for (IndexType i=j; i<mp; ++i) {
```
tob2, line 84: I note that for **GCC 11** using the commented pragma instead of the one above, I get:
* [OMPVV_RESULT: lsms_triangular_packing.cpp] Test **passed on the device.**
* [OMPVV_RESULT: lsms_triangular_packing.cpp] Test **failed on the host.** (with error_sum = 34543.)
* With OpenMP disabled, error_sum = 0.

I have not tried to understand whether that makes sense or not.

spophale: @tob2 Yes, it is strange. I will shortly push another PR - test_target_teams_distribute_parallel_for_collapse.c which is trying to extract the behavior of this kernel while simplifying the logic.
It passes with xl.

tob2: test_target_teams_distribute_parallel_for_collapse.c: that's the merged pull request #226. (It passes with GCC 11 also on the host.)

spophale: Any more thoughts on this kernel ?

spophale:
TODO: Move test from 4.5 to 5.0.
Non-rectangular loop support added in 5.0. Needs to be tested with 5.0 compilers (gcc 10.2 and newer LLVM).

tmh97:
Passed on device w/ latest llvm trunk!

PR #226, opened 2020-10-29
Adding new test for collapse.

spophale:
I am trying to extract out the OpenMP parts of the LSMS kernel. 
Currently, the code works with xlc on device, but gives a seg-fault on host.


tob2:
I *think* the testcase is fine.

Testing with GCC 11 also passes (no OpenMP = serial, OpenMP on the host and OpenMP on the device). _Albeit using the thread sanitizer, when executing with host fallback, TSAN reports a data race: [PR97649](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=97649)._

tmh97 (approved):

jhdavis8 (approved):
Looks good!

PR #227, opened 2020-10-29
test_task_detach.c: Fix function + type names

tob2:
* tests/5.0/task/test_task_detach.c (test_task_detach): Use correct
  function and type names.
Namely:
* omp_fufill_event misses an -l- in "ful"(l) (→ omp_fulfill_event)
* omp_event_t: This was renamed before the final OpenMP 5.0 spec to omp_event_handle_t

See, e.g., https://www.openmp.org/spec-html/5.0/openmpsu162.html — while the pre-OpenMP 5.0 TR7 still had the previous type name: https://www.openmp.org//wp-content/uploads/openmp-TR7.pdf#page=408

### DIFF
--- a/tests/5.0/task/test_task_detach.c
+++ b/tests/5.0/task/test_task_detach.c
@@ -4,10 +4,10 @@
 //
 // This is a test of the 5.0 task detach clause. The clause should cause the
 // task construct to detach and not complete, causing the following taskwait
-// to block thread0 until thread1 fufills the event the task is detached to.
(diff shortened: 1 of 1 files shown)


tob2:
@tmh97, @jhdavis8, @spophale – please review. This is the simple/obvious part; Issue #228 additionally mentions a race condition, which I did not look at (yet).

jhdavis8 (approved):

spophale (approved):

PR #229, opened 2020-11-18
Adding new F90 5.0 test for target_data use_device_addr

tmh97:
New F90 5.0 test. Both xlf and gfortran throw compilation error on line 45 where we test `!$omp target data use_device_addr(device_data)`


tmh97:
`tmh97/sollve_vv/tests/5.0/target_data/test_target_data_use_device_addr.F90", line 45.14: 1514-876 (S) The TARGET DATA construct must have at least one MAP or USE_DEVICE_PTR clause.
"/autofs/nccs-svm1_home1/tmh97/sollve_vv/tests/5.0/target_data/test_target_data_use_device_addr.F90", line 45.28: 1515-022 (S) Syntax Error: Extra token " use_device_addr " was found. The token is ignored.
"/autofs/nccs-svm1_home1/tmh97/sollve_vv/tests/5.0/target_data/test_target_data_use_device_addr.F90", line 45.43: 1515-022 (S) Syntax Error: Extra token " ( " was found. The token is ignored.
"/autofs/nccs-svm1_home1/tmh97/sollve_vv/tests/5.0/target_data/test_target_data_use_device_addr.F90", line 45.44: 1515-022 (S) Syntax Error: Extra token " device_data " was found. The token is ignored.
"/autofs/nccs-svm1_home1/tmh97/sollve_vv/tests/5.0/target_data/test_target_data_use_device_addr.F90", line 45.55: 1515-022 (S) Syntax Error: Extra token " ) " was found. The token is ignored.`

tmh97:
**GCC compilation errors**

-    45 |          !$omp target data use_device_addr(device_data)
      |                                                       1
Error: TARGET DATA must contain at least one MAP clause at (1)

**Still runs on device successfully**

tmh97:
same error is persistent on .c version of the test 
`/test_target_data_use_device_addr.c:29:9: error: expected at least one 'map' or 'use_device_ptr' clause for '#pragma omp target data'`

tob2:
> same error is persistent on .c version of the test

That C file compiles here with GCC 10 (and GCC mainline alias 11); however, it does fail with GCC 9 with the same message.

> **GCC compilation errors**
> `45 | !$omp target data use_device_addr(device_data) | 1 Error: TARGET DATA must contain at least one MAP clause at (1)`
> **Still runs on device successfully**

There is something odd: It should not produce a binary and, hence, should not run if there is an error! Otherwise, I also get that error for GCC 10 (and no binary).

With GCC mainline/11 gfortran, there is no error for `use_device_addr` but it crashes (internal compile error) for `is_device_ptr`; I need to check what goes wrong there.

(Side remark: is_device_ptr and use_device_ptr is very confusing for Fortran in OpenMP 4.5 and 5.0; OpenMP 5.1 clarifies a lot and adds has_device_addr, but that's out of scope.)

tests/5.0/target_data/test_target_data_use_device_addr.F90
```diff
@@ -0,0 +1,67 @@
...
+
+
+CONTAINS
+   INTEGER FUNCTION test(x)
+      REAL, INTENT(INOUT) :: x
+
+      !$omp target data map(to: device_data)
+
+         !$omp target data use_device_addr(device_data)
+            x = device_data
```
tob2, line 46: I do not understand that line:  `x` is a variable on the host; `device_data` is a references to the "corresponding storage in the device data environment ."

But as we are on the host, `device_data`'s device data cannot be accessed. (Exception: unified-shared memory.)

I think one could do: `dev_ptr = c_loc(device_data)`  to obtain the device address but not much more.

tests/5.0/target_data/test_target_data_use_device_addr.F90
```diff
@@ -0,0 +1,67 @@
...
+         !$omp end target data
+
+         !$omp target map(to:device_data) map(tofrom: errors) map(from:host_data) is_device_ptr(x)
```
tob2, line 49: With this `is_device_ptr` I got completely stuck when studying OpenMP 5.0 – as it is completely unclear what `is_device_ptr` is supposed to mean and what is permitted and not also does not make any sense for Fortran.

OpenMP 5.1 solved all my issues and has:

> The `is_device_ptr`clause indicates that its list items are device pointers if they are of type `C_PTR`. [...] For any list item in an `is_device_ptr`clause that is not of type `C_PTR`, the behavior is as if the list item appeared in a has_device_addr clause.

However, even with this reading it does not make sense. `x` is called from host context and is not mapped to the device (i.e. there is no "corresponding storage in the device data environment"). It's value it taken from the bogus `x = device_data` assignment – but that does not turn it into a "corresponding storage in the device data environment".

tob2 (changes requested):
Given the issues I have, I think the code is not ready.

I have to admit that the pointless restrictions of OpenMP 5.0 makes it hard to write a sensible and working testcase.

Side remark: When I wrote a testcase, I use `type(c_ptr), value` to pass the value and then use c_f_pointer + a procedure call to handle the restriction, carefully avoiding the dereferrencing of the pointer on the host side. Maybe my testcase, [GCC's use_device_addr-1.f90](https://gcc.gnu.org/git/?p=gcc.git;a=blob;f=libgomp/testsuite/libgomp.fortran/use_device_addr-1.f90), gives some inspiration (-1 uses double precision, -2 uses real; and -3 and -4 use array-descriptor variables with double precision/real). –  _Just to avoid license issues, I hereby license it (also) under the BSD-3._

tests/5.0/target_data/test_target_data_use_device_addr.F90
tmh97: Thank you for the feedback! I Will take a look at the test-case you referenced above and fix this.

jhdavis8:
I've created an updated version of this test, so I'm going to close this PR and create a new one dedicated to the new form of the test. Will add everyone here as reviewers.

Issue #228, opened 2020-10-29
tests/5.0/task/test_task_detach.c: Order problem, race problem


tob2:
As noted by a colleague, the testcase [tests/5.0/task/test_task_detach.c](https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.0/task/test_task_detach.c) has some issues.
* Some function/type name issues, see Pull Request #227

And an order/race-condition issue:
* "`flag_event` is set on entry to the task in thread 0, and the event is fulfilled in thread 1, but as both threads run concurrently, there is no guarantee that `flag_event` will be set in thread 0 by the time the `omp_fulfill_event` is reached in thread 1 (i.e. a **race condition**).
* _"Also, with the relaxed OpenMP memory model, I believe that even if `flag_event` is initialized in thread 0 first, thread 1 might still not see it without an explicit memory barrier._

* "If there is **only one thread,** then omp_fulfill_event will be called **before** the event has been initialized, which is also wrong."

For reference, the current code is:

```C
#pragma omp parallel num_threads(2) default(shared)
  {
    if (omp_get_thread_num() == 1 || omp_get_num_threads() < 2) {
      num_threads = omp_get_num_threads();
      x++;
      omp_fufill_event(flag_event);
    }
    if (omp_get_thread_num() == 0) {
#pragma omp task detach(flag_event)
      {
        y++;
      }
#pragma omp taskwait
      record_x = x;
      record_y = y;
    }
  }
```

PR #230, opened 2020-11-18
Fixing data race and order issues in task detach test

jhdavis8:
This is an attempt to fix the data race in test_task_detach.c mentioned in issue #228. I changed the test structure substantially. The detached task now has its event fulfilled through a function call in another task. The event-fulfilling function is called "test_callback" (although this name is a bit misleading) as the test is inspired by the examples shown in [this presentation](https://openmpcon.org/wp-content/uploads/2018_Session2_Klemm.pdf). The order is checked by a writes to variables in the first two tasks which are checked in the third task. Data races on x between the second and third task are prevented using flush directives and by ordering the write to x before the event_fulfill.

The test passes Clang 11 only. XL and GCC do not seem to have the features implemented in the versions on Summit.

### DIFF
--- a/tests/5.0/task/test_task_detach.c
+++ b/tests/5.0/task/test_task_detach.c
@@ -59,9 +59,9 @@ int test_task_detach() {
   OMPVV_WARNING_IF(num_threads == 1, "Test ran with one thread, so the results are not conclusive");
 
   OMPVV_TEST_AND_SET_VERBOSE(errors, record_x != 1);
-  OMPVV_ERROR_IF(record_x == 0, "Depend did not wait for associated event to be fulfilled");
-  OMPVV_ERROR_IF(record_y == 0, "Depend did not wait for task body to execute");
-  OMPVV_ERROR_IF(record_x == -1 || record_y == -1, "Recording variables were not set correctly")
+  OMPVV_ERROR_IF(record_x == 0, "Dependent task preceded callback, so detach did not work correctly.");
+  OMPVV_ERROR_IF(record_y == 0, "Dependent task preceded detached task body, so depend did not work correctly.");
+  OMPVV_ERROR_IF(record_x == -1 || record_y == -1, "Recording variables were not set correctly.")
 
     return errors;
 }
--- a/tests/5.0/task/test_task_detach.c
+++ b/tests/5.0/task/test_task_detach.c
@@ -2,14 +2,18 @@
 //
 // OpenMP API Version 5.0 Nov 2018
 //
-// This is a test of the 5.0 task detach clause. The clause should cause the
-// task construct to detach and not complete, causing the following taskwait
-// to block thread0 until thread1 fulfills the event the task is detached to.
-// Order of events should be: code before omp_fulfill_event() ~~ code in task
-// --> code after taskwait. This order is checked by recording the value of
-// variables set before the call to omp_fulfill_event() and in the task.
+// This is a test of the 5.0 task detach clause. A task with the detach
+// should not complete until its associated block and the event in the
+// detach clause is fulfilled with the omp_fulfill_event clause. This test
+// confirms that the task will not complete until these conditions are
+// met by placing a taskwait after the detached task. Variables are set
+// inside the task body and in the function that fulfills the event, and
+// their values are checked after the taskwait to ensure they are set as
+// expected. This test is based on the example of task detach presented
+// by Michael Klemm at the 2018 OpenMPCon.
 //
 ////===----------------------------------------------------------------------===//
+
 #include <assert.h>
 #include <omp.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


jhdavis8:
@tob2 Any thoughts on this? I believe the race condition is fixed.

spophale (approved):

jhdavis8:
We are debating making this test a bit more rigorous. Right now, if the compiler effectively ignores the detach clause, it will pass the test.

tob2 (approved):
I did not spot any problem (except for the style/spacing issue `( event)` (space before `e`) in `test_callback`). – Thus, LGTM.

_A colleague, working on the GCC implementation, noted that it “passes with my implementation” but commented: “Unfortunately, the test doesn't do anything particularly exciting now – it just sets off a single task in a single thread, that fulfills the completion event itself and finishes. It doesn't even fail if detach is simply ignored!”_ [Note: single thread = single thread actively running at a time.]

Side remark: I have the impression that GCC often creates less threads that it could – not necessarily less than it has to – albeit not for this testcase. And I also spotted the following in the OpenMP spec:

> Note – Since the initial value of the _dyn-var_ ICV is implementation defined, programs that depend on a specific number of threads for correct execution should explicitly disable dynamic adjustment of the number of threads.

Namely, `if (dyn-var = true)` → `1 ≤ number of threads ≤ min(ThreadsAvailable, ThreadsRequested)` while for `if (dyn-var = false) and (ThreadsRequested ≤ ThreadsAvailable)` → `number of threads = ThreadsRequested`.

→ for both: https://www.openmp.org/spec-html/5.0/openmpsu35.html

It might make sense to go through the tests that set `num_threads` and check whether adding `omp_set_dynamic(0);` / `call omp_set_dynamic(.false.)` makes sense.

jhdavis8:
Yes, it's not a particularly interesting test at the moment. I'm thinking we can try to add a third task between the two which calls omp_fulfill_event, and use depend clauses to make sure the third task follows the first rather than a taskwait.

tests/5.0/task/test_task_detach.c
```diff
@@ -18,42 +22,47 @@
...
-        y++;
-      }
-#pragma omp taskwait
+#pragma omp task
+    {
+      x = test_callback(flag_event);
+    }
+#pragma omp task depend(inout: y)
+    {
       record_x = x;
```
tob2, line 49: Isn't there a potential race condition for `x`? `y` gets updated and this `record_* =` task waits for it as does `test_callback`, but the `x = ` assignment could happen in parallel with `record_x = ...`. Or do I miss something?

jhdavis8: Hmm I see the concern. I guess those changes would bring us back to the original problem of making the test not really use the detach feature.

jhdavis8: Although my understanding is that the first task, because of its detach clause, will not be considered complete for the purposes of depend until both its associated block is complete and the event is fulfilled. Wouldn't that mean the third task will not start execution until the callback is executed and y is updated? Although I could see an issue with the third task starting before the write to x in the second task is completed.

Here's the 5.0 spec on the detach clause (p. 137, third paragraph):

> If a detach clause is present on a task construct a new event allow-completion-event is created.
> The allow-completion-event is connected to the completion of the associated task region. The
> original event-handle will be updated to represent the allow-completion-event event before the task
> data environment is created. The event-handle will be considered as if it was specified on a
> firstprivate clause. The use of a variable in a detach clause expression of a task
> construct causes an implicit reference to the variable in all enclosing constructs.

What if we change the body of the second task to the the following:

x = 1;
#pragma omp flush
test_callback(flag_event);

This way x is updated before the event is fulfilled. We'd also need to add a flush before record_x in the third task.

jhdavis8: I made this update and it passes Clang, should appear on this PR now. @tob2 do you think this is correct/race-free?

tests/5.0/task/test_task_detach.c
```diff
@@ -18,42 +22,50 @@
...
   OMPVV_WARNING_IF(num_threads == 1, "Test ran with one thread, so the results are not conclusive");
 
-  OMPVV_TEST_AND_SET_VERBOSE(errors, record_x != 1 || record_y != 1);
-  OMPVV_ERROR_IF(record_x == 0, "Taskwait did not wait for associated event to be fulfilled");
-  OMPVV_ERROR_IF(record_y == 0, "Taskwait did not wait for task body to execute");
-  OMPVV_ERROR_IF(record_x == -1 || record_y == -1, "Recording variables were not set in the parallel region as expected.")
+  OMPVV_TEST_AND_SET_VERBOSE(errors, record_x != 1);
+  OMPVV_ERROR_IF(record_x == 0, "Depend did not wait for associated event to be fulfilled");
+  OMPVV_ERROR_IF(record_y == 0, "Depend did not wait for task body to execute");
+  OMPVV_ERROR_IF(record_x == -1 || record_y == -1, "Recording variables were not set correctly")
```
jhdavis8, line 64: Change messages

tests/5.0/task/test_task_detach.c
tob2: Tested with GCC by a colleague; it works with -O0 but not with -O3.

My built does not support it, but when I look at the dump (-fdump-tree-optimized -O0) of a slightly massaged code, I think that is because of a race condition with
```C
#pragma omp task
    {   x = 1; ... }
#pragma omp task depend(inout: y)
    { ...; record_x = x; ... }
```
which GCC turns into something like the following (pseudo code)
```C
args2.x = &x;  // pass by reference - as the value is modified: 'x = 1'
__builtin_GOMP_task (test_task_detach._omp_fn.2, &args2);
args3.x = x; // pass by value - as the value is only read: 'record_x = x'
__builtin_GOMP_task (test_task_detach._omp_fn.3, &args3);
```
And when the task-calling function returns before the task thread has set `x`, it fails – as doing the assignment is quick and due to pass-by-value, the 'omp flush' does not help.

I think GCC's optimization is fine, but I did not (re)check the spec – if so, the race condition is already in the testcase and not only caused by the compiler.

tob2: Re-reading this patch, I think the code does *not*  have a race condition and your patch is *fine*.

Thus: *LGTM*

tob2 (approved):
LGTM

tests/5.0/task/test_task_detach.c
spophale: I will wait for the msgs to be updated before approving the request. Also, I don't particularly like clubbing depend and detach together in 64. It is more informative if it is split up.

jhdavis8: Updated the messages and split the message for unset recording variables on line 64.

spophale (approved):

tmh97 (approved):

PR #237, opened 2021-01-05
Add 5.0 test of loop construct with order(concurrent)

jhdavis8:
This test checks the loop directive with the order(concurrent) clause. According to 5.0 spec, the order(concurrent) clause is assumed to be present even when it is not, so this test also effectively checks the standalone loop directive. The test uses a similar approach to other worksharing loop directive tests we created for 4.5.

Right now on Summit, GCC passes this test while Clang and XL fail due to lack of implementation.

### DIFF
--- a/tests/5.0/loop/test_loop_order_concurrent.c
+++ b/tests/5.0/loop/test_loop_order_concurrent.c
@@ -23,6 +23,7 @@
 int test_loop_order_concurrent() {
   OMPVV_INFOMSG("test_loop_order_concurrent");
   int errors = 0;
+  int total_wait_errors = 0;
   int x[N];
   int y[N];
   int z[N];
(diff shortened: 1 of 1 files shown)


jhdavis8:
I updated this test to check that the threads do not proceed out of the loop region until all other threads have finished. This requirement only exists for loop directives bound to parallel regions.

spophale:
Are you planning another test for explicit order(concurrent) ? I would suggest adding another loop with explicit order (concurrent) in there and cross validate.

jhdavis8:
Huh, I must have removed the order(concurrent) clause when I was testing it. I intended for this to be the test of the explicit clause. I will add a function that tests loop standalone so both cases are covered.

tmh97 (approved):

spophale (approved):

PR #238, opened 2021-01-06
5.0 test of loop construct with bind clause

jhdavis8:
This tests the loop construct with the bind clause with each of its three possible bindings, team, parallel and thread. This test uses a similar array computation and value check to the previous loop construct test. Right now, gcc usually passes this test but will sometimes fail due to a data race on the array in the `bind(thread)` test. I am not sure if this is a compiler issue or a spec interpretation issue, so I made this PR to get feedback on the test.

Also, 5.0 states that when `bind(teams)` is used, threads can continue execution after executing the loop region, while they must wait for other threads when `bind(thread)` or `bind(parallel)` is used. I am thinking about adding a check that the threads wait for other threads before continuing in the `bind(thread)` and `bind(parallel)` tests.

### DIFF
--- a/tests/5.0/loop/test_loop_bind.c
+++ b/tests/5.0/loop/test_loop_bind.c
@@ -16,10 +16,46 @@
 #include <stdlib.h>
 #include "ompvv.h"
 
-#define N 1024
+#define N 32
 
-int test_loop_bind() {
-  OMPVV_INFOMSG("test_loop_bind");
+int test_loop_bind_teams() {
+  OMPVV_INFOMSG("test_loop_bind_teams");
+  int errors = 0;
+  int x[N];
+  int y[N];
+  int z[N];
+  int num_teams = -1;
+
+  for (int i = 0; i < N; i++) {
+    x[i] = 1;
+    y[i] = i;
+    z[i] = 2*i;
+  }
+
+#pragma omp teams num_teams(OMPVV_NUM_TEAMS_DEVICE) thread_limit(OMPVV_NUM_THREADS_HOST)
+  {
+#pragma omp loop bind(teams)
+    for (int i = 0; i < N; i++) {
+      x[i] += y[i]*z[i];
+    }
+    if (omp_get_team_num() == 0) {
+      num_teams = omp_get_num_teams();
+    }
+  }
+
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, x[i] != 1 + (y[i]*z[i]));
+  }
+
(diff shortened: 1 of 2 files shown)


tests/5.0/loop/test_loop_bind.c
```diff
@@ -0,0 +1,151 @@
+//===--- test_loop_bind.c ---------------------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test checks the loop directive with the order(concurrent) clause. The
+// order(concurrent) clause is assumed to be present if it is not present, so
+// this test covers the standalone loop directive as well. The test creates a
+// parallel region with a loop construct nested within, and performs simple
+// operations on an int array which are then checked for correctness. Note
+// that the number of
```
jhdavis8, line 10: Header comment needs updated

tests/5.0/loop/test_loop_bind.c
```diff
@@ -0,0 +1,151 @@
...
+    z[i] = 2*i;
+  }
+
+#pragma omp teams distribute num_teams(OMPVV_NUM_TEAMS_DEVICE) thread_limit(OMPVV_NUM_THREADS_HOST)
+  for (int i = 0; i < N; i++) {
+#pragma omp parallel num_threads(OMPVV_NUM_THREADS_HOST)
+    {
+#pragma omp loop bind(thread)
+      for (int j = 0; j < N; j++) {
+        x[i][j] += y[i]*z[i];
```
jhdavis8, line 122: Location of the data race in question.

tests/5.0/loop/test_loop_bind.c
```diff
@@ -0,0 +1,151 @@
...
+    x[i] = 1;
+    y[i] = i;
+    z[i] = 2*i;
+  }
+
+#pragma omp teams num_teams(OMPVV_NUM_TEAMS_DEVICE) thread_limit(OMPVV_NUM_THREADS_HOST)
+  {
+#pragma omp loop bind(teams)
+    for (int i = 0; i < N; i++) {
+      x[i] += y[i]*z[i];
```
spophale, line 39: I expect all master threads in the number of teams will access the same x. Adding teams binding inside teams is redundant, but we could try both combinations.

jhdavis8:
I modified the test in response to the slack discussion. However, the data race issue remains a concern with the bind(thread) test. Will continue to investigate. GCC still fails the test sometimes due to the race.

jhdavis8:
Updated it again to fix the data race. Passes GCC 10.2.0 consistently now. For the bind(thread) tests, a local array is created, and after the loop region writes to the local array, the local array data is written out to the appropriate index of the larger results array using `omp_get_thread_num()`/`omp_get_team_num()`

spophale (approved):

tmh97 (approved):

PR #240, opened 2021-01-06
Adding tests for OMP_TARGET_OFFLOAD

tmh97:
This tests checks behavior based on the value of the OMP_TARGET_OFFLOAD environment variable. The previous commits used setenv() to set OMP_TARGET_OFFLOAD, however, this new test relies on the user to set the env variable.

First, the test checks if the current OpenMP version supports OMP_TARGET_OFFLOAD. If it does not, an error will be thrown.

Next, the test uses getenv() to return the value of OMP_TARGET_OFFLOAD. If OMP_TARGET_OFFLOAD is NOTSET or UNKOWN, a warning will be issued. 

Alternatively, if the value is set to DISABLED, DEFAULT, or MANDATORY, the test will check for specified behavior.
`DEFAULT` -> Target code is executed on the host if target device d.n.e. or is not supported by the implementation
`DISABLED` -> Behaves as if no target devices is available
`MANDATORY` -> Will terminate execution when target construct is encountered and a target device is not available


tob2:
I have not tried the code with GCC mainline (to become GCC 11.0.1), which supports this feature, but I wonder whether the `setenv("OMP_TARGET_OFFLOAD", "DEFAULT", 1);` comes way too late.

Namely, the test program sets it in `main` – but at least for GCC, in libgomp, the environment variables are parsed in as (static) constructor:
```C
static void __attribute__((constructor))
initialize_env (void)
{ …
```

spophale:
I agree with @tob2 . And that might be the reason why changing it multiple times in the program did not work. We could try to set it with other openmpvv env variables. But we may be able to test only one with that approach.

tests/5.0/program_control/test_omp_target_offload.c
```diff
@@ -0,0 +1,77 @@
...
+   OMPVV_TEST_AND_SET_OFFLOADING(isOffloading);
+ 
+   offload_policy_t policy = get_offload_policy();
+   
+   OMPVV_ERROR_IF(_OPENMP< 201811,"ERROR: OMP_TARGET_OFFLOAD NOT supported by VER. %d",_OPENMP );
+   OMPVV_TEST_AND_SET_VERBOSE(errors, _OPENMP < 201811);
+
+   on_init_dev = 1;
+
+   #pragma omp target map(tofrom: on_init_dev)
```
spophale, line 61: I think the test will be more meaningful if you added some token computation, e.g. increment variable and copy back.

tests/5.0/program_control/test_omp_target_offload.c
```diff
@@ -0,0 +1,77 @@
...
+     on_init_dev=omp_is_initial_device();
+
+   OMPVV_ERROR_IF(policy==DEFAULT && isOffloading == 1 && on_init_dev !=0, "Did not follow DEFAULT policy and executed target region on the host even though a device was available");
+   OMPVV_TEST_AND_SET(errors, policy==DEFAULT && isOffloading == 1 && on_init_dev != 0)
```
spophale, line 65: The second case is policy = DEFAULT, isoffloading ==0, in that case the computation should be correct and should have executed on host.

tests/5.0/program_control/test_omp_target_offload.c
```diff
@@ -0,0 +1,100 @@
...
+//
+// This test checks for offloading behavior when OMP_TARGET_OFFLOAD is set to DEFAULT,
+// DISABLED, or MANDATORY. By DEFAULT, if the target device does not exist or the implementation
+// does not support the target device, the target code is executed on the host. Support for the
+// DISABLED policy is optional. OMP_TARGET_OFFLOAD is set to disabled, the behavior is as if 
+// no target device is avaiable and target regions execute on the host. When set to Mandatory, 
+// the program will terminate execution when a target construct is encountered and a target device
+// is not available or supported by the implementation. 
+// 
+// This tet was adopted from OpenMP 5.0 Examples Doc -> target_offload_control.1.c
```
spophale, line 13: Typo: test 
Only add this line if most of the code was taken from the examples.

spophale (changes requested):

spophale (approved):

jhdavis8 (approved):

PR #241, opened 2021-01-11
Adding new 5.0 test for close map type modifier on map clause

tmh97:
This test checks support of the `close` map-type-modifier on a map clause. Maps several variables to the device, changes values in a target region, and checks for properly updated values back on the host.

Passes on the device with newest llvm trunk.
No support for close map-type-modifier with gcc 10.2.0 or xl 16.1.1-9

### DIFF
--- a/tests/5.0/target/test_target_map_with_close_modifier.c
+++ b/tests/5.0/target/test_target_map_with_close_modifier.c
@@ -24,8 +24,8 @@ int test_close_modifier() {
   int a[N];
 
   struct {
-  int var;
-  int b[N];
+    int var;
+    int b[N];
   } member; 
 
   member.var = 1;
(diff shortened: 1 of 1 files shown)


tests/5.0/target/test_target_map_with_close_modifier.c
```diff
@@ -0,0 +1,67 @@
...
+
+int test_close_modifier() {
+
+  int i;
+  int scalar = 19;
+  int a[N];
+
+  struct {
+  int var;
+  int b[N];
```
jhdavis8, line 28: Region inside struct definition is usually indented

jhdavis8 (approved):

spophale (approved):

spophale (approved):

Issue #242, opened 2021-01-12
Invalid "map(delete:" for "target data" and "target"


tob2:
From https://gcc.gnu.org/pipermail/gcc-patches/2021-January/563167.html

The following affects `5.0/target/test_target_mapping_before_delete.c` which uses `omp target ... map(delete:`.

"In *OpenMP 5.0,* I certainly see the *target data*:
> [163:16] A map-type in a map clause must be to, from, tofrom or alloc.

"restriction and similarly for *target*:

> [174:27] A map-type in a map clause must be to, from, tofrom or alloc.
restriction.

"*OpenMP 4.5* has those restrictions in [97:15] and [106:13]."

This shows up at:

```
5.0/target/test_target_mapping_before_delete.c:#pragma omp target map (to: scalar, a, member) map (from: x, y, z) map(delete: scalar, a, member)
5.0/target/test_target_mapping_before_delete.c:#pragma omp target map (tofrom: scalar, c, member) map (from: x, y, z) map (delete: scalar, c, member)
```

tob2:
Patch would be the following – but the question is whether the testcase should be removed for good as already the name `test_target_mapping_before_delete` does not make sense without the `delete:`.

Or, alternatively, change `before_delete` to `before_from` (file name, comments and function names)?

```diff
--- a/tests/5.0/target/test_target_mapping_before_delete.c
+++ b/tests/5.0/target/test_target_mapping_before_delete.c
@@ -42 +42 @@ int to_before_delete () {
-#pragma omp target map (to: scalar, a, member) map (from: x, y, z) map(delete: scalar, a, member)
+#pragma omp target map (to: scalar, a, member) map (from: x, y, z)
@@ -81 +81 @@ int to_from_before_delete(){
-#pragma omp target map (tofrom: scalar, c, member) map (from: x, y, z) map (delete: scalar, c, member)
+#pragma omp target map (tofrom: scalar, c, member) map (from: x, y, z)
```

spophale:
Thanks @tob2 ! 
It looks like delete is only allowed on target exit data as per 5.0.  I propose to edit the test to use target enter - exit and remove from under target.

spophale:
Addressed.

PR #243, opened 2021-01-12
Issue 242: Remove test_target_mapping_before_delete.c

tob2:
Issue #242

* `tests/5.0/target/test_target_mapping_before_delete.c`: Remove as `omp target` does not permit `map` with `delete:` clause.

Note: only clauses are `omp target map(to: ... map(delete:` and `omp target map(tofrom: ... map(delete:`, hence no need to retain it. 

Spec has: `A map-type in a map clause must be to, from, tofrom or alloc.`

@spophale, @tmh97, @jhdavis8 – please review.


jhdavis8:
Yup, looks like this test is invalid. I don't think there is a valid way to have a map(delete...) and map(to...) (or other variations) on the same construct in 5.0.

jhdavis8 (approved):

spophale (approved):

PR #244, opened 2021-01-12
Add 5.0 test of loop directive with collapse clause

jhdavis8:
This is a test of the loop+collapse. It follows the same format as collapse testing for other worksharing loop directive tests.

Passes GCC 10.2 only on Summit.

### DIFF
--- a/tests/5.0/loop/test_loop_collapse.c
+++ b/tests/5.0/loop/test_loop_collapse.c
@@ -1,0 +1,124 @@
+//===--- test_loop_collapse.c ------------------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test uses the collapse clause with the loop directive and tests that
+// for loops out of the scope of the collapsed loops are not parallelized.
+// This test tests using one and two collapsed loops.
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+#define N 128 //Array Size of 128 uses 16MB target memory and
+                       //scales n^3 in test_collapse2()
+
+int test_collapse1() {
+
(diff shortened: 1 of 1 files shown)


tests/5.0/loop/test_loop_collapse.c
```diff
@@ -0,0 +1,124 @@
...
+    b[x][0] = 0;
+    for (int y = 0; y < N; ++y) {
+      a[x][y] = x + y;
+      b[x][y+1] = 0;
+    }
+  }
+
+#pragma omp parallel num_threads(OMPVV_NUM_THREADS_DEVICE)
+  {
+#pragma omp loop collapse(1)
```
spophale, line 38: This is the implicit default behavior.

tests/5.0/loop/test_loop_collapse.c
```diff
@@ -0,0 +1,124 @@
...
+
+#pragma omp parallel num_threads(OMPVV_NUM_THREADS_DEVICE)
+  {
+    if (omp_get_thread_num() == 0) {
+      num_threads = omp_get_num_threads();
+    }
+#pragma omp loop collapse(2)
+    for (int x = 0; x < N; ++x) {
+      for (int y = 0; y < N; ++y) {
+        for (int z = 0; z < N; ++z) {
```
spophale, line 86: What worries me is the wording in the spec for concurrent: "If the order(concurrent) clause appears on the loop construct, the iterations of the associated *loops* may execute in any order, including concurrently. If the order clause is not present, the behavior is as if the order(concurrent) clause appeared on the construct."

Let's see how most implementations behave.

spophale: @tob2 can you comment: If there exists a collapse (>1) on the loop clause, is there still an expectation that any loop iteration may be executed concurrently ?

spophale (approved):

tmh97 (approved):

PR #246, opened 2021-01-13
New 5.0 tests of loop with reduction clause

jhdavis8:
These tests check the loop construct with the reduction clause. All reduction types are checked. The method used is similar to reduction testing for target teams distribute.

Currently only GCC passes these tests on Summit.


jhdavis8:
Small edits made to error and warning messages

jhdavis8:
Removed offloading check

spophale (approved):

tmh97 (approved):

PR #247, opened 2021-01-13
New 5.0 test for defaultmap

tmh97:
Two new 5.0 tests for defaultmap. These tests are concerned with the `default` and `none` implicit-behavior specified by the defaultmap clause.

Both tests pass on the newest clang and gcc trunks available on summit.


tests/5.0/target/test_target_defaultmap.c
```diff
@@ -0,0 +1,70 @@
...
+int test_default_map() {
+
+   struct test_struct {
+     int s; 
+     int S[N]; 
+   } 
+
+   int scalar; // scalar int variable -> scalar
+   int A[N]; // aggregate variable -> array
+   test_struct new_struct; // aggregate variable -> structure
```
tmh97, line 26: add struct keyword before declaration

spophale (approved):

jhdavis8 (approved):

PR #248, opened 2021-01-14
Add 5.0 test of loop directive with private clause

jhdavis8:
This is a test of the private clause on a loop directive. The test similar to private testing for target teams distribute, by operating on the private variable in a way that would cause high access contention if it were shared, and checking for correctness.


jhdavis8:
Removed the offloading check

tmh97 (approved):
Good to go.

spophale (approved):

PR #249, opened 2021-01-14
Add 5.0 test of loop directive with lastprivate clause

jhdavis8:
This test of the lastprivate clause, unlike other loop tests, is different from usual lastprivate testing for worksharing loop directives.

5.0 spec restricts the lastprivate clause to only be used on loop iteration variables when used with a loop directive. This test checks that the lastprivate'd loop iteration variables for a single-level loop and a two-level collapsed loop end with valid values after the end of the loop region.

### DIFF
--- a/tests/5.0/loop/test_loop_lastprivate.c
+++ b/tests/5.0/loop/test_loop_lastprivate.c
@@ -1,0 +1,98 @@
+//===--- test_loop_lastprivate.c ----------------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test uses the lastprivate clause with a loop directive. According to
+// specification, the list items on a lastprivate clause in this context may
+// only contain loop iteration variables of loops associated with the loop
+// directive. This test checks that the loop iteration variables associated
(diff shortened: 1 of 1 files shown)


jhdavis8:
GCC error:
```
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/loop/test_loop_lastprivate.c: In function 'test_two_loop_levels':
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/loop/test_loop_lastprivate.c:69:29: error: 'lastprivate' clause on a 'loop' construct refers to a variable 'y' which is not the loop iterator
   69 | #pragma omp loop lastprivate(x, y)
      |                             ^
```
I believe this is a compliance error as spec clearly states that the lastprivate may only contain iteration variables from "associated loops", which (as per our dicussion of the collapse construct) includes collapsed loops as being performed here.

jhdavis8:
Passes GCC only on Summit

spophale (approved):

tmh97 (approved):
This is ready to go.

PR #250, opened 2021-01-15
Add 5.0 test of parallel for construct with allocate clause

jhdavis8:
This tests the allocate clause on a for (worksharing-loop) directive. 5.0 spec requires that any list item on an allocate clause must be privatized by a data-sharing clause on the same directive, so this test does so. A private array is created in the parallel for region using a 64-byte-aligned allocator and the array is written to in a SIMD region with the aligned(x: 64) clause. The values are saved to shared array and the values are checked for correctness.

### DIFF
--- a/tests/5.0/parallel_for/test_parallel_for_allocate.c
+++ b/tests/5.0/parallel_for/test_parallel_for_allocate.c
@@ -1,0 +1,62 @@
+//===------ test_parallel_for_allocate.c ----------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// Tests the parallel for directive with allocator clause, based on the
+// OpenMP 5.0 example for allocators. The allocator testing first creates
+// an allocator, with 64-byte alignment and the default memory space,
+// then checks that 64-byte alignment is correct and that the memory can
+// be written to in the parallel for region in private arrays set to
+// allocate with the created allocator. The tests checks that the values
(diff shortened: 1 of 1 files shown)


jhdavis8:
This test passes Clang 11/trunk but no other compilers on Summit

tests/5.0/parallel_for/test_parallel_for_allocate.c
```diff
@@ -0,0 +1,62 @@
...
+
+  int* x;
+  int result[N][N];
+  omp_memspace_handle_t x_memspace = omp_default_mem_space;
+  omp_alloctrait_t x_traits[1] = {omp_atk_alignment, 64};
+  omp_allocator_handle_t x_alloc = omp_init_allocator(x_memspace, 1, x_traits);
+
+#pragma omp parallel for allocate(x_alloc: x) private(x) shared(result) num_threads(OMPVV_NUM_THREADS_HOST)
+  for (int i = 0; i < N; i++) {
+    x = (int *) malloc(N*sizeof(int));
```
spophale, line 34: check for successful allocation.

spophale (approved):

tests/5.0/parallel_for/test_parallel_for_allocate.c
jhdavis8: Check for NULL

tmh97 (approved):

PR #252, opened 2021-01-15
Add 5.0 test of scan directive and reduction(inscan..) on par for simd

jhdavis8:
This is a test of the new scan feature from OpenMP 5.0. The scan directive must appear with either an inclusive or exclusive clause, and the list item appearing in the inclusive/exclusive clause must also appear in a reduction(inscan...) clause on the for/simd construct that contains the scan directive. Thus, this test checks both scan and reduction(inscan...), by conducting both and inclusive and exclusive scan with a sum operator and checking the results against expected results from a sequential prefix sum.

Currently, this test passes GCC only on Summit. XL will not compile the test. It will compile with Clang, but the test reports incorrect values were returned (with 11.0.0-rc1 *and* 12.0.0-latest). This appears to be a compiler bug, so I will investigate further.

### DIFF
--- a/tests/5.0/scan/test_scan.c
+++ b/tests/5.0/scan/test_scan.c
@@ -2,11 +2,11 @@
 //
 // OpenMP API Version 5.0 Nov 2018
 //
-// This test checks the parallel for directive with the order(concurrent)
-// The test performs simple operations on an int array which are then
-// checked for correctness. The specification indicates only that iterations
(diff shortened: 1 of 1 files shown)


spophale (approved):

jhdavis8:
[Bug reported.](https://bugs.llvm.org/show_bug.cgi?id=49115)

tmh97 (approved):

PR #253, opened 2021-01-18
Adding new 5.0 test for nontemporal clause on simd construct

tmh97:
Test that checks for support of nontemporal clause on simd. Passes with newest llvm and gcc 10.2

### DIFF
--- a/tests/5.0/simd/test_simd_nontemporal.c
+++ b/tests/5.0/simd/test_simd_nontemporal.c
@@ -1,12 +1,13 @@
-//===--- test_simd_nontemporal.c ------------------------------------------------------------------------===//
+//===--- test_simd_nontemporal.c -----------------------------------------------------===//
 //
 // OpenMP API Version 5.0 Nov 2018
 //  
-// This test checks for support of the nontemporal clause on a simd construct. The nontemporal clause 
-// indicates that accesses to the storage location of list items have low temporal locality across the
-// iterations in which those storage locations are accessed. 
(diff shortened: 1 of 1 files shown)


tests/5.0/simd/test_simd_nontemporal.c
```diff
@@ -0,0 +1,46 @@
...
+   int i;
+   int a[N], b[N], c[N];
+
+   for (i = 0; i < N; i++) {
+      a[i] = 10; 
+      b[i] = i;
+      c[i] = 2 * i;
+   }   
+
+   #pragma simd nontemporal (a, b, c)
```
spophale, line 29: When you say nontemporal,  you are telling the compiler that the accesses have low temporal locality...increment "i" differently so that you are not accessing too many elements in a single cache line.

tmh97: @spophale made some changes, let me know if this is o.k. and then i'll go ahead and push the device version of this code.

tests/5.0/simd/test_simd_nontemporal.c
```diff
@@ -0,0 +1,50 @@
+//===--- test_simd_nontemporal.c ------------------------------------------------------------------------===//
```
jhdavis8, line 1: This header runs a bit long, I think we try to keep the line length always no greater than 82 characters (for narrow screens, etc.)

jhdavis8 (approved):

spophale (approved):
Looks good.

PR #258, opened 2021-01-21
Add 5.0 test of loop with order concurrent on device

jhdavis8:
This test is identical to the host test of loop with order concurrent except it is in the context of a target parallel construct instead of a parallel construct.

Passes GCC only on Summit.

### DIFF
--- a/tests/5.0/loop/test_loop_order_concurrent_device.c
+++ b/tests/5.0/loop/test_loop_order_concurrent_device.c
@@ -8,7 +8,7 @@
 // loop directive as well. The test creates a parallel region with a loop
 // construct nested within, and performs simple operations on an int array
 // which are then checked for correctness. Additionally, since loop binds to
-// a parallel region the test checks that the threads all wait before
+// a parallel region, the test checks randomly that other threads wait before
 // proceeding out of the loop region. The number of threads is checked in the
 // parallel region but after the loop construct because runtime API calls are
 // not permitted in loop directive regions.
@@ -18,6 +18,7 @@
 #include <omp.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <time.h>
 #include "ompvv.h"
 
 #define N 1024
@@ -30,6 +31,7 @@ int test_loop_order_concurrent_device() {
   int y[N];
   int z[N];
   int num_threads = -1;
+  int rand_indexes[OMPVV_NUM_THREADS_DEVICE];
 
   for (int i = 0; i < N; i++) {
     x[i] = 1;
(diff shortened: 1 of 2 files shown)


tests/5.0/loop/test_loop_order_concurrent_device.c
```diff
@@ -0,0 +1,80 @@
...
+
+#pragma omp target parallel num_threads(OMPVV_NUM_THREADS_HOST) map(tofrom: x[0:N], y[0:N], z[0:N], num_threads, total_wait_errors)
+  {
+    int wait_errors = 0;
+#pragma omp loop order(concurrent)
+    for (int i = 0; i < N; i++) {
+      x[i] += y[i]*z[i];
+    }
+    for (int i = 0; i < N; i++) {
+      if (x[i] == 1) {
```
spophale, line 48: I see what you are doing, but wouldn't it be better to check on the last array index (x[N-1]) ?

spophale: Or rand index between 0 & N-1

tmh97: checking a few random indices might be better

tests/5.0/loop/test_loop_order_concurrent_device.c
```diff
@@ -0,0 +1,80 @@
...
+  int z[N];
+  int num_threads = -1;
+
+  for (int i = 0; i < N; i++) {
+    x[i] = 1;
+    y[i] = i + 1;
+    z[i] = 2*(i + 1);
+  }
+
+#pragma omp target parallel num_threads(OMPVV_NUM_THREADS_HOST) map(tofrom: x[0:N], y[0:N], z[0:N], num_threads, total_wait_errors)
```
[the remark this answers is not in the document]
jhdavis8: Yes, it should be, I'll fix that

tmh97 (approved):

tests/5.0/loop/test_loop_order_concurrent_device.c
jhdavis8: I guess if we're going to test this aspect, might as well do it fully. I'm thinking a separate function with a 2-thread parallel region, with one thread that has all the work and the other thread checking to make sure it doesn't move forward without the other thread finishing.

jhdavis8: Todo: change to random index

spophale:
I don't see the change to test random array elements for completion.

jhdavis8:
I hadn't gotten around to it yet. Should be corrected now, both tests passing with GCC 10.2.0 on Summit.

spophale (approved):

Issue #259, opened 2021-01-22
5.0/.../test_omp_target_offload.c  – Avoid _OPENMP version check


tob2:
`tests/5.0/program_control/test_omp_target_offload.c` added in pull #240, 
https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.0/program_control/test_omp_target_offload.c
contains:

```C
   OMPVV_ERROR_IF(_OPENMP< 201811,"ERROR: OMP_TARGET_OFFLOAD NOT supported by VER. %d",_OPENMP );
   OMPVV_TEST_AND_SET_VERBOSE(errors, _OPENMP < 201811);
```
This is the **only testcase** which contains a `_OPENMP` **version check** test in sollve_vv. However:

* The OpenMP version is already coded in the directory (`5.0`), which is used by the `OMP_VERSION` make variable
* Most compilers start by implementing parts of a new spec version without bumping the version; the latter happens when a larger fraction is supported, depending on the compiler, this can be a smaller or larger fraction.

In particular, GCC mainline (version 11, currently in Stage 4, release approx. mid/end April) supports this feature since October, but GCC still sets `_OPENMP=201511` (= OpenMP 4.5)

PR #260, opened 2021-01-22
5.0/program_control/test_omp_target_offload.c: Remove _OPENMP check

tob2:
Compilers might support this OpenMP 5.0 feature but not enough of 5.0
to have set _OPENMP to 5.0. Like other tests under 4.5 or 5.0, check
the feature without a version check.

Issue #259
* tests/5.0/program_control/test_omp_target_offload.c: Remove error if _OPENMP < 201811.

### DIFF
--- a/tests/5.0/program_control/test_omp_target_offload.c
+++ b/tests/5.0/program_control/test_omp_target_offload.c
@@ -54,10 +54,6 @@ int main() {
    OMPVV_TEST_AND_SET_OFFLOADING(isOffloading);
    offload_policy_t policy = get_offload_policy();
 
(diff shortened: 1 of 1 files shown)


spophale (approved):
Yes, I see your point @tob2 . Especially since we have an additional check and separate set directories for 4.5 and 5.0.

tob2:
@spophale , @jhdavis8 , and @tmh97: Please review.

This is the only place where `_OPENMP` appears in all of sollve_vv – and as exmplained in the commit log and Issue #259, it can cause fails with compilers which actually do support this feature.

jhdavis8 (approved):

PR #264, opened 2021-01-22
Add device version of loop collapse test

jhdavis8:
This is identical to the already-merged loop collapse test but it runs in a target context.


tests/5.0/loop/test_loop_collapse_device.c
```diff
@@ -0,0 +1,125 @@
...
+}
+
+int main() {
+
+  //Check for offloading
+  OMPVV_TEST_OFFLOADING;
+
+  int errors = 0;
+  OMPVV_TEST_AND_SET_VERBOSE(errors, test_collapse1() != 0);
+  OMPVV_TEST_AND_SET_VERBOSE(errors, test_collapse2() != 0);
```
spophale, line 121: Shouldn't the var 'errors' be reset between the 2 calls ?

tmh97 (approved):

tests/5.0/loop/test_loop_collapse_device.c
jhdavis8: The purpose of the errors variable is to indicate to REPORT_AND_RETURN if *either* of the test functions failed, so if errors was reset after line 120, the test would still report a PASS even if `test_collapse1` failed.

spophale: I meant, report and then reset.

jhdavis8: Ah I see. The REPORT macro can only be used once per test since it sets the final PASS/FAIL result for the website, so it would have the same problem.

spophale (approved):

PR #268, opened 2021-01-25
Add device versions of 5.0 loop reduction tests

jhdavis8:
These tests are identical to the regular loop reduction tests except that they are checked in target regions.

### DIFF
--- a/tests/5.0/loop/test_loop_reduction_add_device.c
+++ b/tests/5.0/loop/test_loop_reduction_add_device.c
@@ -1,0 +1,69 @@
+//===--- test_loop_reduction_add_device.c -----------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test uses the reduction clause on a loop directive, testing that the
+// variable in the reduction clause is properly reduced using the add
+// operator. This test checks the above in a target context.
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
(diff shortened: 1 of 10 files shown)


jhdavis8:
These tests pass GCC only on Summit. However, the `and` and `or` tests fail to compile with the following error:
```unresolved symbol __sync_val_compare_and_swap_1
collect2: error: ld returned 1 exit status
mkoffload: fatal error: powerpc64le-unknown-linux-gnu-accel-nvptx-none-gcc returned 1 exit status
compilation terminated.
lto-wrapper: fatal error: /autofs/nccs-svm1_sw/summit/gcc/10.2.0/bin/../libexec/gcc/powerpc64le-unknown-linux-gnu/10.2.0//accel/nvptx-none/mkoffload returned 1 exit status
compilation terminated.
/usr/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status```

I will investigate and file a bug report if needed.

spophale (approved):

tmh97 (approved):

jhdavis8:
With further investigation, this bug is the same as the GCC failures we are already seeing in other `and` and `or` operator reductions on target, as discussed here: https://bitbucket.org/crpl_cisc/sollve_vv/pull-requests/17

My understanding is that these failures are already known, so I will remove the bug report tag and merge this test.

PR #269, opened 2021-01-25
Add 5.0 tests of master taskloop on host and device

jhdavis8:
These tests check the master taskloop construct on the host and device.

At present Clang passes both tests, XL will not compile, and GCC passes the device version but not the host version, due to an issue with omp_get_num_threads(). I will investigate this failure further.


spophale (approved):

jhdavis8:
With further investigation, the num_threads issue is not a bug. The way I was checking the number of threads was not spec compliant. Will fix in this test and other taskloop tests.

tmh97 (approved):

PR #272, opened 2021-01-27
Update test_omp_target_offload.c

spophale:
For 5.0 the value must be either DEFAULT, DISABLED or MANDATORY, hence any other value is an error.

### DIFF
--- a/tests/5.0/program_control/test_omp_target_offload.c
+++ b/tests/5.0/program_control/test_omp_target_offload.c
@@ -89,8 +89,8 @@ int main() {
    OMPVV_ERROR_IF(policy==MANDATORY && isOffloading == 1 && on_init_dev != 0, "Did not follow MANDATORY, instead executed target region on host even though device was available");
    OMPVV_TEST_AND_SET(errors, policy==MANDATORY && isOffloading == 1 && on_init_dev != 0);
  
-   OMPVV_WARNING_IF(policy==UNKNOWN,"OMP_TARGET_OFFLOAD has an unknown value");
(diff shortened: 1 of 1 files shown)


tests/5.0/program_control/test_omp_target_offload.c
```diff
@@ -89,8 +89,8 @@ int main() {
    OMPVV_ERROR_IF(policy==MANDATORY && isOffloading == 1 && on_init_dev != 0, "Did not follow MANDATORY, instead executed target region on host even though device was available");
    OMPVV_TEST_AND_SET(errors, policy==MANDATORY && isOffloading == 1 && on_init_dev != 0);
  
-   OMPVV_WARNING_IF(policy==UNKNOWN,"OMP_TARGET_OFFLOAD has an unknown value");
-   OMPVV_WARNING_IF(policy==NOTSET, "OMP_TARGET_OFFLOAD has not been set");
+   OMPVV_ERROR_IF(policy==UNKNOWN,"OMP_TARGET_OFFLOAD has an unknown value");
+   OMPVV_ERROR_IF(policy==NOTSET, "OMP_TARGET_OFFLOAD has not been set");
```
tob2, line 94: Other values/unknown are outside of the spec, hence, giving an error is surely sensible. While `notset` is valid, testing for unset is besides the point for this testcase – hence, an error is also fine → LGTM.

_(Requires manually setting the env var or something like PR #262 to avoid the error.)_

tob2 (approved):

jhdavis8 (approved):

tmh97 (approved):

Issue #274, opened 2021-01-29
test_loop_lastprivate{,_device}.c – reduce stack-memory usage? (Segfault with default stack size)


tob2:
Especially due to Fortran and how compilers handle local arrays, it is common to use unlimited stackspace (`ulimit -s unlimited` with Bash/POSIX Shell). However, on most systems, the stack size is still limited to 8192 kB.

On such a system,
* `tests/5.0/loop/test_loop_lastprivate.c` (added with PR #249)
* `tests/5.0/loop/test_loop_lastprivate_device.c` (added with PR #265)

segfaults  as the binary runs out of stack: The main problem is `SIZE` being 1024 and `int a[SIZE][SIZE];` (and likewise `b`) which each needs 4 MB of memory, leaving only 192kB for the rest of the program. Actually, using 8300 kB works.

Increasing manually the stack size is one option, but I still wonder whether reducing the stacksize (SIZE2 = 512?) would be still sensible.

@jhdavis8  thoughts?

PR #275, opened 2021-02-01
Issue #274: test_loop_lastprivate* reduce req stack size < default limit

tob2:
Issue #274
The two two-dimensional arrays used 8MB of stack space; with default Linux stack size limits (8192 kB), it was
very likely to segfault.  Reducing the size makes it run with default settings.

* tests/5.0/loop/test_loop_lastprivate.c: Add SIZE2 = 512.
  (test_two_loop_levels): Use it as second array bound.
* tests/5.0/loop/test_loop_lastprivate_device.c: Likewise.

### DIFF
--- a/tests/5.0/loop/test_loop_lastprivate.c
+++ b/tests/5.0/loop/test_loop_lastprivate.c
@@ -17,6 +17,7 @@
 #include "ompvv.h"
 
 #define SIZE 1024
+#define SIZE2 512
 
 int test_one_loop_level() {
   int a[SIZE];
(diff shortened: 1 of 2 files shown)


tests/5.0/loop/test_loop_lastprivate.c
```diff
@@ -1,4 +1,4 @@
-//===--- test_loop_lastprivate.c ----------------------------------------------===//
```
jhdavis8, line 1: I believe this line may have been deleted accidentally? Otherwise looks good.

jhdavis8 (changes requested):

tests/5.0/loop/test_loop_lastprivate.c
tob2: Thanks for pointing out - now re-added.

spophale (approved):

jhdavis8 (approved):

PR #280, opened 2021-02-02
New 5.0 test, target version of simd nontemporal test

tmh97:
Target version of simd test. Passes on newest llvm trunk and gcc 10.2.0

### DIFF
--- a/tests/5.0/target_simd/test_target_simd_nontemporal.c
+++ b/tests/5.0/target_simd/test_target_simd_nontemporal.c
@@ -33,7 +33,7 @@ int test_simd_nontemporal() {
       }   
 
    for (i = 0; i < N; i++) {
-      if (i % 100 == 0) { 
+      if (i % STRIDE_LEN == 0) { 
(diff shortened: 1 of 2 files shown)


tests/5.0/target_simd/test_target_simd_nontemporal.c
```diff
@@ -0,0 +1,52 @@
...
+   int a[N], b[N], c[N];
+
+   for (i = 0; i < N; i++) {
+      a[i] = 10; 
+      b[i] = i;
+      c[i] = 2 * i;
+   }   
+
+   #pragma simd target nontemporal (a, b, c)
+      for (i = 0; i < N; i += 100) {
```
jhdavis8, line 31: Should we turn `100` into a macro, called something like STRIDE_LEN? Just to allow some flexibility/avoid magic numbers.

jhdavis8 (approved):

spophale (approved):

PR #282, opened 2021-02-03
Add 5.0 tests of task construct with in_reduction clause

jhdavis8:
This is a test of the in_reduction clause on a task construct, on the host and device, and also the parallel construct with the reduction(task...) clause. Currently both tests pass with GCC, while XL does not support the feature. Clang passes on the host version but produces the following compile-time error for the device version:

```
nvlink error   : Undefined reference to '__kmpc_taskred_modifier_init' in '/tmp/test_task_in_reduction_device-3bc10c.cubin'
nvlink error   : Undefined reference to '__kmpc_task_reduction_modifier_fini' in '/tmp/test_task_in_reduction_device-3bc10c.cubin'
nvlink error   : Undefined reference to '__kmpc_task_reduction_get_th_data' in '/tmp/test_task_in_reduction_device-3bc10c.cubin'
clang-11: error: nvlink command failed with exit code 255 (use -v to see invocation)
```


tests/5.0/task/test_task_in_reduction_device.c
```diff
@@ -0,0 +1,69 @@
...
+  int z[N];
+  int sum = 0;
+  int expected_sum = 0;
+
+  for (int i = 0; i < N; i++) {
+    y[i] = i + 1;
+    z[i] = 2*(i + 1);
+  }
+
+#pragma omp target parallel reduction(task, +: sum) num_threads(OMPVV_NUM_THREADS_DEVICE) shared(y, z, num_threads) map(tofrom: sum, y, z, num_threads)
```
spophale, line 33: It is certainly not incorrect, but are y, z, and num_threads explicitly shared ?

spophale (approved):

tests/5.0/task/test_task_in_reduction_device.c
spophale: Also, y and z do not need to be copied back.

spophale (changes requested):
For device version, do not copy back y and z arrays.

tests/5.0/task/test_task_in_reduction_device.c
jhdavis8: They would be shared implicitly in this context but I figured it would be good to set them out explicitly.

jhdavis8:
https://bugs.llvm.org/show_bug.cgi?id=49150

Bug reported.

spophale (approved):

tmh97 (approved):

PR #284, opened 2021-02-03
New 5.0 test for standalone depobj construct

tmh97:
Test passes on newest llvm trunk and gcc 10.2.0

### DIFF
--- a/tests/5.0/depobj/test_depobj_depend_update_destroy.cpp
+++ b/tests/5.0/depobj/test_depobj_depend_update_destroy.cpp
@@ -48,14 +48,14 @@ void go(int turn, int a[], int b[], omp_depend_t *obj) {
    #pragma omp parallel num_threads(2)
    #pragma omp single 
    {
-      #pragma omp task depend(depobj *obj)
+      #pragma omp task depend(depobj: *obj)
       {      
          for (int i = 0; i < N; i++) {
               a[i] += 1;
          }
       }
       
-      #pragma omp task depend(in a[:n])
+      #pragma omp task depend(in: a[:N])
       { 
          for (int i = 0; i < N; i++) {
               OMPVV_TEST_AND_SET(errors, a[i]!=(i+turn));
--- a/tests/5.0/depobj/test_depobj_depend_update_destroy.cpp
+++ b/tests/5.0/depobj/test_depobj_depend_update_destroy.cpp
@@ -28,6 +28,7 @@ int main () {
 
    for (int i = 0; i < N; i++) {
       a[i] = i;
+      b[i] = 0;
    }
    
    #pragma omp depobj(obj) depend(inout: a)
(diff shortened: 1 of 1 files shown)


tests/5.0/depobj/test_depobj_depend_update_destroy.cpp
```diff
@@ -0,0 +1,65 @@
...
+#define N 1028
+#define TURN1 1
+#define TURN2 2
+
+void go(int turn, int a[], int b[], omp_depend_t *obj);
+int errors = 0;
+
+
+int main () {
+   int a[N], b[N];
```
jhdavis8, line 26: Looks like b is unused, so it can be removed

tests/5.0/depobj/test_depobj_depend_update_destroy.cpp
```diff
@@ -0,0 +1,65 @@
...
+      {      
+         for (int i = 0; i < N; i++) {
+              a[i] += 1;
+         }
+      }
+      
+      #pragma omp task depend(in: a[:N])
+      { 
+         for (int i = 0; i < N; i++) {
+              OMPVV_TEST_AND_SET(errors, a[i]!=(i+turn));
```
jhdavis8, line 61: Since both the tasks can execute simultaneously, maybe wrap this `TEST_AND_SET` in an `if` statement that checks if `turn == 1`, so it will only be checked for the execution where the second task must follow the first.

To check the final results, you can do a `OMPVV_TEST_AND_SET_VERBOSE` on `a` after destroying the depobj, in `main`.

Also, might be best to stick with the `_VERBOSE` version of `TEST_AND_SET` everywhere, just so this output can be disabled.

jhdavis8, line 61: Since both the tasks can execute simultaneously, maybe wrap this `TEST_AND_SET` in an `if` statement that checks if `turn == 1`, so it will only be checked for the execution where the second task must follow the first.

To check the final results, you can do a `OMPVV_TEST_AND_SET_VERBOSE` on `a` after destroying the `depobj`, in main.

Also, might be best to stick with the `_VERBOSE` version of `TEST_AND_SET` everywhere, just so this output can be disabled.

tests/5.0/depobj/test_depobj_depend_update_destroy.cpp
jhdavis8, line 26: Looks like `b` is unused, so it can be removed.

jhdavis8 (changes requested):

tests/5.0/depobj/test_depobj_depend_update_destroy.cpp
spophale: To be clearer, for turn == 2, the tasks could execute sequentially, simultaneously or in reversed order. The check on 61 needs to be only for turn 1. For turn 2 maybe modify "b" in one of the tasks and check on host that the value was correctly modified.

spophale (approved):

jhdavis8 (approved):

PR #287, opened 2021-02-06
Add 5.0 tests of taskloop construct with in_reduction clause

jhdavis8:
These tests check the taskloop construct with the in_reduction clause. They follow the format of other in_reduction tests, and have the same results on Summit.


spophale (approved):
Since you are explicitly mentioning shared, I think "sum" should also be shared. But I am being nitpicky at this point.

tmh97 (approved):

PR #289, opened 2021-02-09
Adding new 5.0 test for omp_get_device_num

tmh97:
Both gcc/10.2.0 and newest llvm fail on compilation with similar error -> `implicit declaration of function 'omp_get_device_num' is invalid in C99 [-Wimplicit-function-declaration]`

### DIFF
--- a/tests/5.0/program_control/test_omp_get_device_num.c
+++ b/tests/5.0/program_control/test_omp_get_device_num.c
@@ -45,8 +45,8 @@ int test_omp_get_dev_num(void) {
       }
    }
 
-   OMPVV_TEST_AND_SET_VERBOSE(errors, target_device_num == original_device_num);
- 
+   OMPVV_WARNING_IF(target_device_num == original_device_num, "omp_get_device_num() returned the same device number as host, cannot guarantee that target region properly offloaded to device");
+   
    for (int i = 0; i < N; i++) {
       OMPVV_TEST_AND_SET_VERBOSE(errors, a[i] != i + 5);
    }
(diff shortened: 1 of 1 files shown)


tests/5.0/program_control/test_omp_get_device_num.c
```diff
@@ -0,0 +1,64 @@
...
+ 
+   OMPVV_TEST_AND_SET_VERBOSE(errors, original_device_num != omp_get_initial_device());
+
+   for (int i = 0; i < N; i++) {
+      a[i] = 0;
+      b[i] = 5;
+      c[i] = i;
+   }
+   
+   #pragma omp target map(from: target_device_num)
```
tmh97, line 40: explicit  to: a,b and c

tests/5.0/program_control/test_omp_get_device_num.c
```diff
@@ -0,0 +1,64 @@
...
+   
+   #pragma omp target map(from: target_device_num) map(to: b, c) map(tofrom: a)
+   { 
+      target_device_num = omp_get_device_num();
+      for (int i = 0; i < N; i++) {
+         a[i] = b[i] + c[i];
+      }
+   }
+
+   OMPVV_TEST_AND_SET_VERBOSE(errors, target_device_num == original_device_num);
```
spophale, line 48: Why is this an error ?

spophale: Spec does not say so, I would only add an info msg saying that the target and original device have the same device no.

spophale (approved):

jhdavis8 (approved):

PR #292, opened 2021-02-12
Add Fortran version of task detach test

jhdavis8:
This test will not compile with gfortran or xlf on Summit. It follows the form of the C version of the test.

### DIFF
--- a/tests/5.0/task/test_task_detach.F90
+++ b/tests/5.0/task/test_task_detach.F90
@@ -35,8 +35,8 @@ PROGRAM test_task_detach
 
   CONTAINS
     INTEGER FUNCTION test_callback(event)
-      INTEGER(kind=omp_event_handle_t) :: event
-      omp_fulfill_event(event)
+      INTEGER(kind=omp_event_handle_kind) :: event
+      call omp_fulfill_event(event)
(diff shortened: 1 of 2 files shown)


spophale:
Does gfortran give an error or not supported?

tob2:
> Does gfortran give an error or not supported?

GCC 10 does not support it – but GCC 11 does. Currently, it fails on either because of bugs in the program:
* `kind=omp_event_handle_t` does not exist; it should be `kind=omp_event_handle_kind` (twice)
* `omp_fulfill_event(event)` – that's a subroutine; add `call`
* `test_callback`: That's a function (returning an integer), but it is used as `test_callback(flag_event)` (which resembles a subroutine call) – thus, either you need to use what `test_callback(flag_event)` returns (e.g. assign to a variable) or `test_callback` has to be converted to a subroutine (and a `call` has to be added).

spophale:
Thanks @tob2 .

jhdavis8:
Thank @tob2, I think the bugs you pointed out should be fixed now.

tmh97 (approved):

spophale (approved):

PR #293, opened 2021-02-12
5.0/new test/test nested declare target

spophale:
Currently fails on Summit with gcc 10.2. Passes with latest llvm 13 (trunk build) on Summit.


jhdavis8:
What's the difference between this and PR #160, which is `test_declare_target_nested.c`? Might be good to rename this or the other test to avoid ambiguity.

spophale:
Other than testing different forms of declare targets, this test is focusing on nesting of declare targets, that focuses on nested functions with the outer function included in declare target.

tmh97 (approved):

jhdavis8 (approved):

PR #294, opened 2021-02-21
New 5.0 test for target with in_reduction clause

tmh97:
Test currently passes if `target in_reduction`is switched with `task in_reduction`. Fails gcc/10.2.0 and newest clang release due to lack of support `error: 'in_reduction' is not valid for '#pragma omp target'`

### DIFF
--- a/tests/5.0/target/test_target_in_reduction.c
+++ b/tests/5.0/target/test_target_in_reduction.c
@@ -21,7 +21,7 @@ void compute_on_host(int *);
 
 int main ()
 {
-   int i;
+   int i, host_reduction_sum = 0, device_reduction_sum = 0;
    int sum = 0, total = 0, errors = 0;
    
    #pragma omp parallel master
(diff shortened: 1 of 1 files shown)


tests/5.0/target/test_target_in_reduction.c
```diff
@@ -0,0 +1,64 @@
+//===--- test_target_in_reduction.c ------------------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test checks task reductions for a target task resulting from a target
+// construct with the 'in_reduction' clause.
+//
+////===-------------------------------------------------------------------------------------===//
```
tmh97, line 8: fix header length

tests/5.0/target/test_target_in_reduction.c
```diff
@@ -0,0 +1,64 @@
...
+
+#pragma omp declare target to (compute_on_device)
+
+int main ()
+{
+   int i;
+   int sum = 0, total = 0, errors = 0;
+   
+   #pragma omp parallel master
+   #pragma omp taskgroup task_reduction(+:sum)
```
spophale, line 28: sum is shared, I think you are introducing a race condition here.

tests/5.0/target/test_target_in_reduction.c
```diff
@@ -0,0 +1,64 @@
...
+   
+   for (i = 0; i < N; i++) {
+      total += 2;
+   }
+  
+   if (sum != total) {
+         errors++;
+   }
+
+   OMPVV_ERROR_IF(sum != total, "Target task did not participate in the reduction");
```
spophale, line 45: This is not very informative. One way to differentiate the contributions of target and task are to add different values --> if you added a different number (line 54,62) then the value of sum would indicate which of the reductions did not work.

jhdavis8 (approved):

spophale (approved):

PR #295, opened 2021-02-22
New 5.0 test for target construct with device clause

tmh97:
Test passes on newest llvm release. GCC seems to be lacking support for device clause, specifically the device-modifier expressions.

### DIFF
--- a/tests/5.0/target/test_target_device.c
+++ b/tests/5.0/target/test_target_device.c
@@ -1,0 +1,81 @@
+//===--- test_target_device.c ----------------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test checks the target construct with device clause where device-
+// modifier is either ancestor or device_num. If no device_modifier is 
+// present, the behavior is the same as if device_num were present.
+//
+////===---------------------------------------------------------------------===//
+
+#include <stdlib.h>
+#include <stdio.h>
+#include <omp.h>
+#include "ompvv.h"
+
+#define N 1028
+
+int test_target_device_ancestor() {
+
+    int i, which_device;
+    int a[N];
+    int errors = 0; 
+
+    for (int i = 0; i < N; i++) {
+        a[i] = i;
+    }
+
+    #pragma omp target device(ancestor: 1) map(tofrom: a, which_device) 
+    {
+        for (int i = 0; i < N; i++) {
+            a[i] = a[i] + 2;
+        }
+ 
+        which_device = omp_is_initial_device();
+    }
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, which_device != 1);
+    OMPVV_WARNING_IF(which_device != 1, "Target region was executed on device. Due to ancestor device-modifier,"
+                                         "this region should execute on host");
+
+    return errors;
+
+}
+
+int test_target_device_device_num() {
+    
+    int i, which_device;
+    int b[N];
+    int errors = 0; 
+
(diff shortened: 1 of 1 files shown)


tests/5.0/target/test_target_device.c
```diff
@@ -0,0 +1,81 @@
...
+    
+    int i, which_device;
+    int b[N];
+    int errors = 0; 
+
+    for (int i = 0; i < N; i++) {
+        b[i] = i;
+    }
+
+    #pragma omp target device(device_num: 1) map(tofrom: b, which_device)
```
spophale, line 55: In all practical situations the assumption that device_num == 1 will work, although device numbers are implementation defined. Ideally the device_num should be set to a value between 0 - omp_get_num_devices() and the value should not be equal to omp_get_device_num() on host.

tests/5.0/target/test_target_device.c
```diff
@@ -0,0 +1,81 @@
...
+    {
+        for (int i = 0; i < N; i++) {
+            a[i] = a[i] + 2;
+        }
+ 
+        which_device = omp_is_initial_device();
+    }
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, which_device != 1);
+    OMPVV_WARNING_IF(which_device != 1, "Target region was executed on device. Due to ancestor device-modifier,"
```
jhdavis8, line 38: Should this be ERROR_IF since the errors var is being set if which_device != 1

jhdavis8 (approved):
Looks good besides minor change

spophale (changes requested):
For test_target_device_device_num() - query the number of devices to first check if any are available (before line 55).

tests/5.0/target/test_target_device.c
```diff
@@ -0,0 +1,93 @@
...
+int test_target_device_ancestor() {
+
+    int i, which_device;
+    int a[N];
+    int errors = 0; 
+
+    for (int i = 0; i < N; i++) {
+        a[i] = i;
+    }
+
```
tmh97, line 27: Check that device is actually available, if not issue warning/error

tmh97:
@spophale hey just pushed the changes we discussed the last meeting. I used omp_get_device_num() in the target region and it broke things with llvm. I could revert back to using omp_is_initial_device(), but I feel like omp_get_device_num() makes more sense here (line 77)

tests/5.0/target/test_target_device.c
```diff
@@ -0,0 +1,99 @@
...
+    OMPVV_TEST_AND_SET(errors, omp_get_num_devices() <= 0);
+    OMPVV_ERROR_IF(omp_get_num_devices() <= 0, "There are no target devices available, cannot properly "
```
spophale, line 29: msg --> Since no target devices were found this test will be skipped.

tests/5.0/target/test_target_device.c
```diff
@@ -0,0 +1,99 @@
...
+
+    for (int i = 0; i < N; i++) {
+        b[i] = i;
+    }
+
+    host_device_num = omp_get_device_num(); 
+
+    
+    OMPVV_TEST_AND_SET(errors, omp_get_num_devices() <= 0);
+    OMPVV_ERROR_IF(omp_get_num_devices() <= 0, "There are no target devices available, cannot properly test");
```
spophale, line 66: see msg above.

spophale (approved):
Looks good, just change the error messages.

PR #302, opened 2021-02-26
Add new 5.0 tests for requires with atomic_default_mem_order clause

tmh97:
Adding 3 different tests for each memory behavior that can be specified through the atomic_default_mem_order clause on a requires directive.

All tests pass gcc/10.20 and latest llvm trunk. 

Seq_cst and acq_rel are based on the same example, while relaxed behavior is based on a slightly different example.


jhdavis8 (approved):
LGTM

spophale (approved):

PR #303, opened 2021-02-28
New 5.0 test for target directive with allocate clause

tmh97:
Based off of existing 5.0 allocate test. Currently causes CUDA error with latest llvm trunk.

### DIFF
--- a/tests/5.0/target/test_target_allocate.c
+++ b/tests/5.0/target/test_target_allocate.c
@@ -1,0 +1,60 @@
+//===------ test_allocate_allocators.c ------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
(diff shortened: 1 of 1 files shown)


spophale:
Don't we already have this test? See sollve_vv/tests/5.0/allocate/test_allocate_allocator.c

tmh97:
> Don't we already have this test? See sollve_vv/tests/5.0/allocate/test_allocate_allocator.c

It is the same as that test, except that it is also checking for allocate clause on a target region. Just realized I did not change the name in the header to reflect the new test

jhdavis8 (approved):

tmh97:
@spophale ready for re-review

spophale (approved):

PR #304, opened 2021-03-01
Update test_declare_target_device_type_nohost.c

### LINKED ISSUE #301: Fallback code tries to call `nohost` function
The fallback code required by the standard will in the following example try to call a `nohost` function.
The compiler error clang gives is correct and expected. You need to guard the call, `ifdef`, `[begin/end] declare variant`, or something else.

https://github.com/SOLLVE/sollve_vv/blob/b9fc5c9f10a80dccec54fe75a2b2c84f0793e399/tests/5.0/declare_target/test_declare_target_device_type_nohost.c#L41

spophale:
Functions declared as nohost require additional protection against fall back mechanism that may require it to execute on host.

closes #301

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
+++ b/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
@@ -27,7 +27,6 @@ int a[N], b[N], c[N];
 int i = 0;
 #pragma omp end declare target
 
-//#pragma omp declare variant(target_function) match(construct={target})
 #pragma omp declare variant(target_function) match(device={kind(nohost)})
 void update() {
   for (i = 0; i < N; i++) {
@@ -36,7 +35,6 @@ void update() {
     c[i] += 5;
   }
 }
-//#pragma omp end declare variant
 
 #pragma omp declare target device_type(nohost)
 void target_function(){
@@ -45,8 +43,7 @@ void target_function(){
     b[i] += 2;
     c[i] += 3;
   }
-}
-#pragma omp end declare target 
+} 
 
 int test_declare_target_device_type_nohost() { 
 
--- a/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
+++ b/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
@@ -7,6 +7,7 @@
 // clause is present on the contained declare target directive, then its argument 
 // determines which versions are made available. If device_type(nohost) is present 
 // only a device version of the procedure is made available. 
+// The device version of the function is specified via declare variant.
 // 
 //===-------------------------------------------------------------------------------===//
 
@@ -16,7 +17,7 @@
 #include <stdlib.h>
 #include "ompvv.h"
 
-#define N 1024
+#define N 10
 int errors = 0;
 
 void target_function();
@@ -26,39 +27,40 @@ int a[N], b[N], c[N];
 int i = 0;
 #pragma omp end declare target
 
-#pragma omp begin declare target device_type(nohost)
-#pragma omp declare variant(target_function) match(device={kind(nohost)})
+#pragma omp declare variant(target_function) match(construct={target})
 void update() {
+  for (i = 0; i < N; i++) {
+    a[i] += 5;
+    b[i] += 5;
+    c[i] += 5;
+  }
+}
+
+#pragma omp begin declare target device_type(nohost)
+void target_function(){
   for (i = 0; i < N; i++) {
     a[i] += 1;
     b[i] += 2;
     c[i] += 3;
   }
 }
-#pragma omp end declare target
-
-//#pragma omp declare target to(update) device_type(nohost)
-
+#pragma omp end declare target 
 
 int test_declare_target_device_type_nohost() { 
 
+  update();
+
   #pragma omp target update to(a,b,c)
+
   #pragma omp target  
   {
-    if (!omp_is_initial_device()){
-      target_function();
-    }
-    
-    for (i = 0; i < N; i++) {
-      a[i] += i;
-      b[i] += 2 * i;
-      c[i] += 3 * i;
-    }
+    update();
   }
+
   #pragma omp target update from (a,b,c)
   
   for (i = 0; i < N; i++) { //check array values on host
-    if ( a[i] != 2 * i + 1 || b[i] != 4 * i + 2 || c[i] != 6 * i + 3 ) {
+    if ( a[i] != 6 || b[i] != 7 || c[i] != 8 ) {
       errors++;
     }
   }
@@ -67,12 +69,14 @@ int test_declare_target_device_type_nohost() {
 }
 
 int main () {
+
+  OMPVV_TEST_OFFLOADING;
   
   //initalize arrays on host
   for (i = 0; i < N; i++) {
-    a[i] = i;
-    b[i] = 2 * i;
-    c[i] = 3 * i;
+    a[i] = 0;
+    b[i] = 0;
+    c[i] = 0;
   }
 
   OMPVV_TEST_AND_SET_VERBOSE(errors, test_declare_target_device_type_nohost());
--- a/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
+++ b/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
@@ -27,7 +27,8 @@ int a[N], b[N], c[N];
 int i = 0;
 #pragma omp end declare target
 
-#pragma omp declare variant(target_function) match(construct={target})
+//#pragma omp declare variant(target_function) match(construct={target})
+#pragma omp declare variant(target_function) match(device={kind(nohost)})
 void update() {
   for (i = 0; i < N; i++) {
     a[i] += 5;
@@ -35,8 +36,9 @@ void update() {
     c[i] += 5;
   }
 }
+//#pragma omp end declare variant
 
-#pragma omp begin declare target device_type(nohost)
+#pragma omp declare target device_type(nohost)
 void target_function(){
   for (i = 0; i < N; i++) {
     a[i] += 1;
--- a/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
+++ b/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
@@ -38,8 +38,10 @@ int test_declare_target_device_type_nohost() {
   #pragma omp target update to(a,b,c)
   #pragma omp target  
   {
-    update();
-
+    if (!omp_is_initial_device()){
+      update();
+    }
+    
     for (i = 0; i < N; i++) {
       a[i] += i;
       b[i] += 2 * i;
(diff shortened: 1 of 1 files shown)


spophale:
With the latest llvm 13 build on Summit I still see:
```
nohost.c:42:6: error: function with 'device_type(nohost)' is not available on host
     update();
     ^
nohost.c:34:31: note: marked as 'device_type(nohost)' here
#pragma omp declare target to(update) device_type(nohost)
```

jdoerfert:
`device_type(nohost)` does not create a symbol on the host. Regardless if the call is in a base language conditional or not, there is no function to call, hence the error. Idk what gcc does and why but the clang behavior is expected. You need some guard that will not require a host symbol to be present, e.g., ifdef or put the call in a begin/end declare variant region.

spophale:
@jdoerfert I cannot find anything in the specification that requires the additional checks. I might be missing something. Could you point me to the relevant text ?

jdoerfert:
> @jdoerfert I cannot find anything in the specification that requires the additional checks. I might be missing something. Could you point me to the relevant text ?

https://www.openmp.org/spec-html/5.1/openmpsu70.html#x98-1080002.14.7

The device_type clause specifies if a version of the procedure or variable should be made available on the host, device or both. If host is specified only a host version of the procedure or variable is made available. If any is specified then both device and host versions of the procedure or variable are made available. **If nohost is specified for a procedure then only a device version of the procedure is made available**. If nohost is specified for a variable then that variable is not available on the host. If no device_type clause is present then the behavior is as if the device_type clause appears with any specified.

spophale:
@jdoerfert 
Yes, and shouldn't calling it within target region and adding initial device check be adequate ? Without getting into specifics of individual implementations, it seems unreasonable for users to know how LLVM internals work or to account for it.

jdoerfert:
> @jdoerfert
> Yes, and shouldn't calling it within target region and adding initial device check be adequate ? Without getting into specifics of individual implementations, it seems unreasonable for users to know how LLVM internals work or to account for it.

This has nothing to do with LLVM. The standard clearly states when a procedure is made available. If you try to call something that is not available there is no reason for it to work, in fact, it should not work. Emitting the function for the host, which is what GCC seems to do and you argue is "correct", would be the exact opposite of the intention for `device_type(nohost)`. Phrased differently: why should something be `device_type(nohost)` if it would be available on the host regardless?

spophale:
We are *not* calling it on the host, even with the assumption that there is a fall-back, the conditional would prevent it. Unless the specification specified that nohost must be used in combination with declare variant (or some such non-OpenMP mechanism) it is subject to difference in interpretation.

jdoerfert:
> We are _not_ calling it on the host, even with the assumption that there is a fall-back, the conditional would prevent it. Unless the specification specified that nohost must be used in combination with declare variant (or some such non-OpenMP mechanism) it is subject to difference in interpretation.

This conversation is going nowhere but drains both time and energy.

Here is the summary:
- Fall-back is not an assumption but a requirement.
- It is in the fall-back called on the host.
- Conditionals might prevent execution at runtime but the language still requires symbols for such callees: https://godbolt.org/z/66ae8M

If you don't believe me, maybe ask an OpenMP expert or implementer of your choosing.

spophale:
Yes, I agree. Things that may be obvious to compiler developers may not be read and understood the same way as a user. Analyzing further, "made available" while describing device_type is being interpreted differently by LLVM and GCC. We are endeavoring to get to a 5.0 compliant test and that is all there is to it. 

Maybe this will be a good correction for 5.2.

jacobdweightman:
@jdoerfert, I think you raise a good point that we shouldn't be able to refer to a function which doesn't exist (i.e. the host version of `update`). @spophale, you make a good point that the spec doesn't mention any checks. I think the misunderstanding here is compounded by the absence of `device_type` in the examples document.

With that said, I would like to suggest another possibility. I'm not certain that this is legal because of the multiple definitions of `void update()`, but could we perhaps do something like this?
```
#pragma omp begin declare target device_type(host)
// This host definition should never be called thanks to @spophale's new check!
void update() {
  assert(false && "update() should never be called from the host!");
}
#pragma omp end declare target

#pragma omp begin declare target device_type(nohost)
void update() {
  for (i = 0; i < N; i++) {
    a[i] += 1;
    b[i] += 2;
    c[i] += 3;
  }
}
#pragma omp end declare target
```

jdoerfert:
Declare variant (see my first post) is what you were looking for:

```
#pragma omp begin declare variant match(device={kind(host)})
 // This host definition should never be called thanks to @spophale's new check!
 void update() {
   assert(false && "update() should never be called from the host!");
 }
#pragma omp end declare variant

#pragma omp begin declare variant match(device={kind(nohost)})
 void update() {
   for (i = 0; i < N; i++) {
     a[i] += 1;
     b[i] += 2;
     c[i] += 3;
   }
 }
 #pragma omp end declare variant
 ```
Compiles then fine with Clang: https://godbolt.org/z/aPTfvMhdx

**Comment by jacobdweightman** [comment]
Hmm... it figured making the suggestion was worth a shot, but your counter-proposal of using `declare variant` instead of `declare target device_type(nohost)`— which would be a great solution to this problem in a practical application — seems less than useful in a program which aims to test `declare target device_type(nohost)`...

I wonder what example programs the OpenMP committee had in mind when they added this to the spec? I can't seem to come up with a conforming way to actually make use of a function declared with this clause. It would be really nice if there were something in the examples document.

**Comment by jdoerfert** [comment]
> I can't seem to come up with a conforming way to actually make use of a function declared with this clause. It would be really nice if there were something in the examples document.

Sure you can: https://godbolt.org/z/h534v6anT

**Comment by jacobdweightman** [comment]
@jdoerfert I see, that makes sense. I guess I was specifically looking for something that doesn't use `declare variant`, and I think that's what @spophale was hoping for as well (correct me if I'm wrong). Given that `declare variant` is also a relatively new OpenMP feature, testing them independently as much as possible seems like a good goal, so that test failures reveal issues as specifically as possible.

However, if it isn't possible to test them independently, then I think rewriting this test with `declare variant` is a viable path forward. What do you think, @spophale?

**Comment by jdoerfert** [comment]
> @jdoerfert I see, that makes sense. I guess I was specifically looking for something that doesn't use `declare variant`, and I think that's what @spophale was hoping for as well (correct me if I'm wrong). Given that `declare variant` is also a relatively new OpenMP feature, testing them independently as much as possible seems like a good goal, so that test failures reveal issues as specifically as possible.

TBH, I'm not sure what you are really looking for. The behavior is defined independent from the use case. I can come up with the latter all day but the former is still fix.
FWIW, here is the canonical `#ifdef` solution: https://godbolt.org/z/8W5cWKGKv 
If you build a device runtime, e.g., `clang -fopenmp-is-device`, or via a different language, you can call the `nohost` functions.
...

**Comment by spophale** [comment]
Thanks @jacobdweightman and @jdoerfert for reviving the conversation. 
@jacobdweightman you are correct, for V&V the tests adding too many OMP features together is not ideal. Between the two options (#ifdef and declare variant ), the former is more tempting. As an example, the latter **needs** to be made available. I don't mind including both as separate tests. 

@jacobdweightman, @jdoerfert will the #ifdef __NVPTX__ solution work with Cray/HPE/AMD compilers ?

**Comment by jdoerfert** [comment]
> @jacobdweightman, @jdoerfert will the #ifdef **NVPTX** solution work with Cray/HPE/AMD compilers ?

Only one way to find out for sure, and I don't know where I'd go looking for those compilers. I'd be interested to know though :)

**Comment by jacobdweightman** [comment]
I just tried rerunning this test with the suggested
```
#ifdef NVPTX
update();
#endif
```
with the Cray compiler, and execution fails for offload to an AMDGPU in a way that suggests no call occurred. This _might_ work for offloading to an NVIDIA GPU with the Cray compiler, but using a specific GPU vendor's macro seems inherently non-portable.

Is there a portable compile-time check we could do instead?

**Comment by jacobdweightman** [comment]
Another possibility would be to use multiple files. Consider something that looks like this:
File 1:
```
void update();
#pragma omp declare target to(update) // technically this is implied

int main() {
    #pragma omp target
    update();
}
```
File 2:
```
#include <stdio.h>

void update() {
    printf("nohost!\n");
}
#pragma omp declare target to(update) device_type(nohost)
```
File 3:
```
void update() {
   /* empty or abort()*/
}

#pragma omp declare target to(update) device_type(host)
```
Would it be too difficult to add a multi-file test to SOLLVE?

**Comment by spophale** [comment]
@jacobdweightman We usually keep 1 file per test. This might mess with our reporting harness, let me look into it.

**Review by tmh97 — APPROVED** [review]

**Review by tmh97 — APPROVED** [review]

**Review by nolanbaker31 — APPROVED** [review]
looks good!

File: tests/5.0/declare_target/test_declare_target_device_type_nohost.c [hunk 1]
```diff
@@ -7,49 +7,62 @@
...
+void target_function(){
   for (i = 0; i < N; i++) {
     a[i] += 1;
     b[i] += 2;
     c[i] += 3;
   }
 }
-
-#pragma omp declare target to(update) device_type(nohost)
+#pragma omp end declare target
```
jacobdweightman, line 49: An `end declare target` cannot occur here.

In OpenMP 5.0, an `end declare target` can only occur after a `declare target` which has no clauses ([spec](https://www.openmp.org/spec-html/5.0/openmpsu62.html)).

This was changed in OpenMP 5.1, where `end declare target` can occur after a `begin declare target` with clauses ([spec](https://www.openmp.org/spec-html/5.1/openmpsu70.html)). My understanding is that without the addition of `begin`, it was difficult to disambiguate the forms with/without the `end`.

tests/5.0/declare_target/test_declare_target_device_type_nohost.c
```diff
@@ -7,49 +7,62 @@
...
+//#pragma omp declare variant(target_function) match(construct={target})
+#pragma omp declare variant(target_function) match(device={kind(nohost)})
+void update() {
+  for (i = 0; i < N; i++) {
+    a[i] += 5;
+    b[i] += 5;
+    c[i] += 5;
+  }
+}
+//#pragma omp end declare variant
```
jacobdweightman, line 39: Looks like there's some commented code lying around. Should this be deleted before merging?

tests/5.0/declare_target/test_declare_target_device_type_nohost.c
```diff
@@ -7,49 +7,62 @@
...
+void target_function();
+
 #pragma omp declare target
 int a[N], b[N], c[N];  
 int i = 0;
 #pragma omp end declare target
 
-void update() { 
+//#pragma omp declare variant(target_function) match(construct={target})
+#pragma omp declare variant(target_function) match(device={kind(nohost)})
```
jacobdweightman, line 31: Shouldn't this be `kind(host)`?

tests/5.0/declare_target/test_declare_target_device_type_nohost.c
```diff
@@ -7,49 +7,59 @@
...
     a[i] += 1;
     b[i] += 2;
     c[i] += 3;
   }
-}
-
-#pragma omp declare target to(update) device_type(nohost)
+} 
 
 int test_declare_target_device_type_nohost() {
```
jacobdweightman, line 48: Sorry, I just noticed this. This function returns an `int`, but contains no `return` statements.

PR #305, opened 2021-03-01
New 5.0 test for metadirective with when and default

tmh97:
First test fails due to no support for omp_get_device_num() in target region from gcc/10.2.0, passes if removed 
Second test passes on gcc/10.2.0
llvm does not support

### DIFF
--- a/tests/5.0/metadirective/test_metadirective_arch_is_nvidia.c
+++ b/tests/5.0/metadirective/test_metadirective_arch_is_nvidia.c
@@ -1,0 +1,58 @@
+//===---test_metadirective_is_nvidia.c -------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+// 
+// Test for metadirectives based on OpenMP 5.0 examples metadirective.1-3.c
+//
+////===---------------------------------------------------------------------===//
+
+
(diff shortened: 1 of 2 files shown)


tests/5.0/metadirective/test_metadirective_arch_is_nvidia.c
```diff
@@ -0,0 +1,58 @@
...
+
+   int which_device;
+   int errors = 0;
+
+   for(int i=0; i<N; i++) { 
+      v1[i] = (i+1); 
+      v2[i] = -(i+1); 
+   }
+
+   #pragma omp target map(to:v1,v2) map(from:v3) device(0)
```
spophale, line 29: There is an implicit assumption here that is not supported by the specification. device(0) need not be an Nvidia device.   The test will return true for all devices. The only guarantee the spec provides is 

> When called on the host device, it will return the same value as the
22 omp_get_initial_device routine.

tmh97: @spophale new commit addresses this.

spophale (approved):

jhdavis8 (approved):

PR #309, opened 2021-03-12
Add Fortran test of use_device_ptr on target data

jhdavis8:
I'm adding this test not necessarily because I think it is valid, but it order to discuss it at the next meeting. It appears the following restriction to the `is_device_ptr` clause for Fortran impairs our ability to test `use_device_ptr` in the same manner as we did for C:

```
A list item that appears in an is_device_ptr clause must be a dummy argument that does
4 not have the ALLOCATABLE, POINTER or VALUE attribute.
```

Any insight into how to work around this is appreciated. I am still thinking about how exactly it would be best to test this feature in Fortran. At the moment, Summit XLF and gfortran both will not compile this test due to it not meeting the aforementioned restriction.

### DIFF
--- a/tests/5.0/target_data/test_target_data_use_device_ptr.F90
+++ b/tests/5.0/target_data/test_target_data_use_device_ptr.F90
@@ -1,0 +1,62 @@
+!/===-- test_target_data_use_device_ptr.c ---------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This file is a test for the use_device_ptr when used with the map clause
+! with target data directive. This test uses arrays of size N whose values
+! are modified on the device and tested in the host. Once the array has
+! been mapped to the device, the use_device_ptr should be able to be used
+! with the ptr to the array and subsequently modify values on the device.
+! This test ensures address conversions of use_device_ptr clauses will
+! occur as if performed after all variables are mapped according to those
+! map clauses.
+!
+!/===----------------------------------------------------------------------===//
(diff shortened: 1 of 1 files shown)


tests/5.0/target_data/test_target_data_use_device_ptr.F90
```diff
@@ -0,0 +1,62 @@
+!/===-- test_target_data_use_device_ptr.c ---------------------------------===//
```
jhdavis8, line 1: Fix header lang

jhdavis8:
Try to move target into subroutine, use INTENT(in)

jhdavis8:
Update: I have rewritten the test to use the target data construct with use_device_ptr in a subroutine on a scalar variable with INTENT(in). This form of the test passes both Fortran compilers on Summit.

tests/5.0/target_data/test_target_data_use_device_ptr.F90
```diff
@@ -0,0 +1,56 @@
...
+    !$omp end target
+    !$omp end target data
+
+  END SUBROUTINE run_target_region
+  INTEGER FUNCTION use_device_ptr()
+    INTEGER :: errors, x
+    INTEGER :: scalar_device, scalar_host
+
+    errors = 0
+    scalar_device = x
```
tmh97, line 47: Should x be initialized as some integer value?

spophale (approved):

tmh97 (approved):

PR #310, opened 2021-03-18
Add 5.0 Fortran test of declare variant construct

jhdavis8:
This is a test of the declare variant construct in Fortran. The test creates function with variants for parallel and target contexts, which are all then called and the array modifications made in each function call are checked for correctness.

GCC and XLF on Summit do not support the feature at present.

### DIFF
--- a/tests/5.0/declare_variant/test_declare_variant.F90
+++ b/tests/5.0/declare_variant/test_declare_variant.F90
@@ -1,0 +1,94 @@
+!===--- test_declare_variant.F90 -------------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test uses the declare variant clause to create two variants of a
+! simple base function, one for use in a parallel region, and one for use
+! in a target region. The function sets each element of an array to its
+! index. Each variant is called on a separate array and the results are
+! checked.
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
+
+MODULE variants
+  use omp_lib
+CONTAINS
+  SUBROUTINE fn(arr)
+    INTEGER,INTENT(inout) :: arr(N)
+    INTEGER :: i
+    !$omp declare variant(p_fn) match(construct = {parallel})
+    !$omp declare variant(t_fn) match(construct = {target})
+
+    DO i = 1, N
+       arr(i) = i
+    END DO
+  END SUBROUTINE fn
+
+  SUBROUTINE p_fn(arr)
+    INTEGER,INTENT(inout) :: arr(N)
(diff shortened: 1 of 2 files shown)


jhdavis8:
Check GCC results again.

jhdavis8:
GCC 10.2.1-latest compile output on Summit:

```
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/declare_variant/test_declare_variant.F90:23:11:

   23 |     !$omp declare variant(p_fn) match(construct = {parallel})
      |           1
Error: Unclassifiable OpenMP directive at (1)
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/declare_variant/test_declare_variant.F90:24:11:

   24 |     !$omp declare variant(t_fn) match(construct = {target})
      |           1
Error: Unclassifiable OpenMP directive at (1)
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/declare_variant/test_declare_variant.F90:56:7:

   56 |   USE variants
      |       1
Fatal Error: Cannot open module file 'variants.mod' for reading at (1): No such file or directory
compilation terminated.
```

tests/5.0/declare_variant/test_declare_variant.F90
```diff
@@ -0,0 +1,94 @@
...
+  OMPVV_TEST_VERBOSE(test_body() .ne. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+
+CONTAINS
+  INTEGER FUNCTION test_body()
+    INTEGER :: errors, x
+    INTEGER,DIMENSION(N) :: a, b, c
+
+    DO x = 1, N
```
spophale, line 69: Default should do something else so that we can verify that the variant was called.

tests/5.0/declare_variant/test_declare_variant.F90
```diff
@@ -0,0 +1,94 @@
...
+MODULE variants
+  use omp_lib
+CONTAINS
+  SUBROUTINE fn(arr)
+    INTEGER,INTENT(inout) :: arr(N)
+    INTEGER :: i
+    !$omp declare variant(p_fn) match(construct = {parallel})
+    !$omp declare variant(t_fn) match(construct = {target})
+
+    DO i = 1, N
```
spophale, line 26: Default should do something different if we want to verify that the variant was exercised.

jhdavis8: Good catch, should be fixed now. I also adjusted the error checking to distinguish which variant was not used as expected.

spophale (approved):

PR #312, opened 2021-03-26
New 5.0 test for taskloop with reduction clause

tmh97:
Passes on llvm trunk and gcc/10.2.0, however, the test results are inconsistent with llvm and sporadically fails.

### DIFF
--- a/tests/5.0/taskloop/test_taskloop_reduction.c
+++ b/tests/5.0/taskloop/test_taskloop_reduction.c
@@ -1,0 +1,78 @@
+//===--- test_taskloop_reduction.c ------------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test checks the taskloop directive with the reduction clause specified.
+//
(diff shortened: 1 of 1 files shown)


tests/5.0/taskloop/test_taskloop_reduction.c
```diff
@@ -0,0 +1,78 @@
...
+}
+   
+   real_sum += N;
+   
+   for (int j = 0; j < num_threads; j++) {
+      for (int i = 0; i < N; i++) {
+         real_sum += a[i]*b[i];
+      }
+   }
+   printf("Value of sum is: %d\n ", sum);
```
tmh97, line 56: remove these

tmh97:
Test again tomorrow (march 27), may be a bug.

tmh97:
Ran 20 times with llvm 03/22 and gcc 10.2.0
llvm -> 14 pass 6 fail
gcc -> 19 pass 1 fail
When the test fails, the result is as if one thread did not participate in the reduction

PR #317, opened 2021-04-07
Add Fortran host and device 5.0 tests of loop construct

jhdavis8:
These tests will not compile on the Summit Fortran compilers due to lack of support but the compilers otherwise do not detect any errors at compile time. They follow the format of the C tests as usual.

### DIFF
--- a/tests/5.0/loop/test_loop_order_concurrent.F90
+++ b/tests/5.0/loop/test_loop_order_concurrent.F90
@@ -1,0 +1,86 @@
+!===--- test_loop_order_concurrent.F90 -------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test checks the loop directive with the order(concurrent) clause.
+! The order(concurrent) clause is assumed to be present if it is not
+! present, so this test covers the standalone loop directive as well. The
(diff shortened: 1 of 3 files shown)


spophale (approved):

tests/5.0/loop/test_loop_order_concurrent.F90
```diff
@@ -0,0 +1,81 @@
...
+
+PROGRAM test_loop_order_concurrent
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
+  OMPVV_TEST_OFFLOADING
+
+  OMPVV_TEST_VERBOSE(test_loop() .ne. 0)
```
tmh97, line 30: change to OMPVV_TEST_AND_SET_VERBOSE()

tmh97 (approved):

tests/5.0/loop/test_loop_order_concurrent.F90
jhdavis8: In this case I'm just using TEST_VERBOSE since there's no need for an explicit errors variable, this will generate and error if the condition isn't met

PR #331, opened 2021-04-23
Target version of 5.0 omp_get_supported_active_levels test

tmh97:
Target version of https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.0/program_control/test_omp_get_supported_active_levels.c.

- GCC and LLVM issue compile-time errors for `omp_get_supported_active_levels()` within a target region. 

- The 5.0 specification states "The omp_get_supported_active_levels routine returns the number of active levels of
parallelism supported by the implementation."

- For omp_get_max_levels, the specification states "The omp_get_max_active_levels routine returns the value of the max-active-levels-varICV, which determines the maximum number of nested active parallel regions on the device."

It seems reasonable to infer that `omp_get_supported_active_levels` should only be called from the sequential portion of the code, if so I can place that routine outside the target region (tested and works fine).

### DIFF
--- a/tests/5.0/program_control/test_omp_get_supported_active_levels.c
+++ b/tests/5.0/program_control/test_omp_get_supported_active_levels.c
@@ -23,8 +23,8 @@ int main() {
    int num_supp_active_levels;
    int max_active_levels;
 
-   errors = 0;
- 
+   errors = 0; 
+   
    num_supp_active_levels = omp_get_supported_active_levels();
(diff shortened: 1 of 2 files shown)


spophale (approved):

spophale:
I don't see any limitations in the spec that mention that they may not be called from inside the target region. But from vendor documentation, I see that there is only one value set (e.g. CCE sets max-active-levels-var to 4095), so maybe these calls are more meaningful on host.

PR #332, opened 2021-04-26
Adding declare target test for declare target with end declare target.

spophale:
This checks the first allowed declare target syntax.


spophale:
Tested with llvm trunk and gcc (10.2.1).

jhdavis8 (approved):
Looks good to me

tmh97 (approved):

PR #335, opened 2021-05-12
Update test_target_teams_distribute_device.c

simoatze:
Fix array section passed to map clause. The program uses `num_teams[dev]`, but according to spec that should be `num_teams[dev:1]`:

The syntax of array sections is mentioned in Section 2.1.5 of the OpenMP Spec 5.1:

```
To specify an array section in an OpenMP construct, array subscript expressions are extended with the following syntax:
[ lower-bound : length : stride] or
[ lower-bound : length : ] or
[ lower-bound : length ] or
[ lower-bound : : stride] or
[ lower-bound : : ] or
[ lower-bound : ] or
[ : length : stride] or
[ : length : ] or
[ : length ] or
[ : : stride]
[ : : ]
[ : ]
```

Thanks!

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_device.c
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_device.c
@@ -47,11 +47,11 @@ int main() {
   }
 
   for (int dev = 0; dev < num_devices; ++dev) {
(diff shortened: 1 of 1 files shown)


spophale:
Thanks @simoatze ! The patch is valid under 4.5 spec as well.

spophale (approved):

tmh97 (approved):

PR #336, opened 2021-05-13
Update test_target_enter_data_classes_inheritance.cpp

simoatze:
I believe there is a bug in this code.

The code in question is:

```
...
class B : public Mapper<B> {
protected:
  int n;
private:
  double* x;
  double sumB;

public:
  B(int n) : Mapper<B>(this), n(n) {
    x = new double[n];
    std::fill(x, x+n, 0);
    double* solutionX = x;
    int &cpy_n = n;
#pragma omp target enter data map(to:solutionX[0:n], cpy_n)
  }
....
```

The problematic issue is `int &cpy_n = n`, with `enter map(to:cpy_n)`.
`cpy_n` is a reference to `n`.  But what is `n` here?  Is it the class member `n`?  is it the argument to the method `B`?  It appears to be the argument `n`.  If I rename the argument `n` to `nn`, change the `n(n)` to `n(nn)`, then the program works, inside the constructor method `B` here `n` refers unambiguously to the member `n`.

What do you think?

Thanks!

### DIFF
--- a/tests/4.5/target_enter_data/test_target_enter_data_classes_inheritance.cpp
+++ b/tests/4.5/target_enter_data/test_target_enter_data_classes_inheritance.cpp
@@ -70,7 +70,7 @@ private:
   double sumB; 
 
 public:
-  B(int n) : Mapper<B>(this), n(n) {
+  B(int nn) : Mapper<B>(this), n(nn) {
     x = new double[n];
     std::fill(x, x+n, 0);
     // This is a work around to avoid referring to 


simoatze:
Also if you prefer we can just use `this->n` inside the constructor method `B` .

spophale:
We have not seen this test fail on XL, GNU or LLVM. 
The 'n' is used to set size of x.

simoatze:
Well, it seems to be passing by luck with other compilers (unless I am missing something).
Looking at the original test, it seems that this line: 

```int &cpy_n = n;```

intends to reference class member `n`. But instead it references argument `n` due to shadowing, thus reference `cpy_n` is invalid once the function terminates and argument `n` goes out of scope

For example, if I print the reference of `n` (argument), `this->n` (member) and `cpy_n`, using gcc I see:

```
$ g++ -I/proj/ta/tests/OpenMP_4.5_Test_Suite/correctness/sollve_vv/ompvv  -fopenmp -g test_target_enter_data_classes_inheritance.cpp -std=c++11
[satzeni@sky8 ~]$ ./a.out
&n: 0x7ffec4c10984
&this->n: 0x1ca6ebc
&cpy_n: 0x7ffec4c10984
```
It shows that `cpy_n` is being set to the value of the pointer of the argument `n` and not the member `n`. So, when executed on the device, even if the copy works, once it goes out of scope the present table would have a stale host pointer.

Does it makes sense?

Thanks!
Simone

spophale:
Interesting. 
Let me look at what is happening with the other compilers (and why is the test passing!!).

simoatze:
Any news about this? Thanks!

spophale (approved):

spophale:
I have not been able to pin-point why both tests pass, but looking at the code, the fix makes sense. LLVM trunk issues warnings for this test, but that is out of scope for this PR.

spophale:
@tmh97 , @nolanbaker31 ping!

nolanbaker31 (approved):

tmh97 (commented):
LGTM

PR #338, opened 2021-05-14
Adding separate compilation commands for amd and nvidia device type

tmh97:
Updating commands for cray and clang compilers, adding ifdefs for amd and nvidia device type

### DIFF
--- a/sys/make/make.def
+++ b/sys/make/make.def
@@ -71,16 +71,16 @@ FC?=none
 ifeq ($(DEVICE_TYPE), amd)
   # CRAY compiler (AMD)
   ifeq ($(CC), cc)
-    CFLAGS += -fopenmp -fopenmp-targets=amdgcn-amd-amdhsa -Xopenmp-target=amdgcn-amd-amdhsa
+    CFLAGS += -g -Ofast -fopenmp 
     CLINK = cc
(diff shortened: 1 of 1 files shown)


tob2:
I don't know whether you also want to do this separation also for GCC; by default, **GCC** compiles for all configured offloading devices - which is usually "none" or a specific one, but the compilers by SUSE and Debian/Ubuntu provide (if the necessary optional packages are installed) support for both AMD GCN and nvptx. – The binaries are multi-arch binaries, which support all compiled-in offloading targets (+ host fallback).

**GCC**
- **nvptx:** `-foffload=nvptx-none`  can be used to force `nvptx` offloading (and disabling other offloading targets); GCC (since GCC 11?) often requires `-foffload=-latomic` (or `-foffload=nvptx-none=-latomic`) to link libatomic.
- **AMD GCN:** `-foffload=amdgcn-amdhsa`  + `-foffload=-march=fiji` or likewise for `gfx900`, `gfx906`, or `gfx908` (or `-foffload=amdgcn-amdhsa=-march=...` - if you compile for multiple offload targets at the same time.)

spophale:
@tmh97 are you planning to incorporate GCC changes suggested by Tobias in this PR ?

spophale (approved):

jhdavis8 (approved):

PR #341, opened 2021-05-20
Add missing return statement to test_omp_get_device_num.c

### LINKED ISSUE #340: test_omp_get_device_num.c function doesn't have return statement
test_omp_get_device_num.c contains a function called `int test_omp_get_dev_num(void)`. However, despite nominally returning an int, the function definition contains no return statement.

For example, this results in the following warning from Clang:
> warning: non-void function does not return a value [-Wreturn-type]

The function definition simply seems to be missing its last line:
```
return errors;
```

jacobdweightman:
Closes #340


tmh97 (approved):

spophale (approved):

spophale:
Thanks @jacobdweightman !

PR #343, opened 2021-05-21
Fix error variable initialization in test_declare_variant.F90

### LINKED ISSUE #342: Fortran 5.0 test_declare_variant initializes the wrong error variable
test_declare_variant.F90 maintains 3 separate error counters: `default_errors`, `p_errors`, and `t_errors`. However, the test later initializes an undeclared variable called `errors`.

For example, this bug causes the Cray compiler to report a compilation failure and the following diagnostic:
> IMPLICIT NONE is specified in the host scope, therefore an explicit type must be specified for data object "ERRORS".

jacobdweightman:
Closes #342

### DIFF
--- a/tests/5.0/declare_variant/test_declare_variant.F90
+++ b/tests/5.0/declare_variant/test_declare_variant.F90
@@ -72,7 +72,9 @@ CONTAINS
        c(x) = 0
     END DO
 
-    errors = 0
+    default_errors = 0
+    p_errors = 0
+    t_errors = 0
 
     call fn(a)
 


spophale (approved):

jhdavis8 (approved):

od102:
this example is a copy of the openmp.org examples, **as such the proper reference needs to be provided**. also, what's the issue?

tmh97:
@od102 I can add some text in the header that gives credit to openmp.org examples. As for what the issue is, it looks like the wrong error variable was initalized, see closed issue #342

od102:
@tmh97 that'd be awesome. wrt the issue, ok, i see, thanks! after that i think we can close this one. again, thanks!

PR #345, opened 2021-06-04
Adding new 5.0 test for flush directive with no mem ordering

tmh97:
5.0 Test for flush directive when no memory ordering clause is specified. Passes on latest llvm trunk and gcc/10.2.0

### DIFF
--- a/tests/5.0/flush/test_flush_no_memory_order_clause.c
+++ b/tests/5.0/flush/test_flush_no_memory_order_clause.c
@@ -1,0 +1,41 @@
+//===---- test_flush_no_memory_order_clause.c -------------------------------===//
+// 
+// OpenMP API Version 5.0 
+// 
+// This is a test of the flush directive with no memory-order-clause specified.
+// Additionally, atomic constructs are used alongside the explicit flush directives
+// to specify memory ordering amongst the two threads.
+//
+// Based on OpenMP 5.0 Example aquire_release.3.c
+//===------------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+int errors = 0;
+
+int main() {
+
+   int x = 0, y = 0;
+   #pragma omp parallel num_threads(2)
+   {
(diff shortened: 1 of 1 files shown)


tests/5.0/flush/test_flush_no_memory_order_clause.c
```diff
@@ -0,0 +1,41 @@
...
+         #pragma omp flush
+         #pragma omp atomic write
+         y = 1;
+      } else {
+         int tmp = 0;
+         while (tmp == 0) {
+            #pragma omp atomic read
+	    tmp = y;
+         }
+         #pragma omp flush
```
spophale, line 36: why is this flush needed ?

tests/5.0/flush/test_flush_no_memory_order_clause.c
```diff
@@ -0,0 +1,41 @@
...
+      if (thrd == 0) {
+         x = 10;
+         #pragma omp flush
+         #pragma omp atomic write
+         y = 1;
+      } else {
+         int tmp = 0;
+         while (tmp == 0) {
+            #pragma omp atomic read
+	    tmp = y;
```
jhdavis8, line 34: Minor change, this should be indented a bit more to line up with the above line.

tests/5.0/flush/test_flush_no_memory_order_clause.c
jhdavis8: I don't think it is needed, as I understand the test.

jhdavis8 (approved):

tests/5.0/flush/test_flush_no_memory_order_clause.c
jhdavis8: Actually, I think it is needed since thread 1 is reading x after thread 0 writes to x and flushes. So thread 1 needs to do an acquire flush here to ensure it has the latest version of x when x's value is checked.

spophale: Right, the binding is to the calling thread only.

spophale (approved):

PR #346, opened 2021-06-04
5.0/loop/test_loop_order_concurrent*.F90: Fix random init

tob2:
* tests/5.0/loop/test_loop_order_concurrent.F90: Call standard
  intrinsic subroutine RANDOM_SEED instead of nonexisting
  init_random_seed.
* tests/5.0/loop/test_loop_order_concurrent_device.F90: Likewise.

@spophale @jhdavis8 @tmh97  – please review. Thanks!


jhdavis8 (approved):
Looks good, good catch

spophale (approved):

PR #347, opened 2021-06-07
Add Fortran versions of uses_allocators tests

jhdavis8:
These do not compile on either of the Summit Fortran compilers


spophale (changes requested):
The fortran tests themselves look good, the first line in the header still says *.c. Please change it to .F90.

tmh97 (approved):
Tests look good.

jhdavis8:
Fixed the header test names to match the correct file extension.

spophale (approved):

PR #348, opened 2021-06-07
New use_device_addr test in Fortran 5.0

jhdavis8:
This is a completely redesigned test for the use_device_addr clause on the target data construct for OpenMP 5.0 in Fortran. 

Thanks to @tob2 for pointing us to the GCC testcase which shows how to correctly use this feature in 5.0, working around the spec restrictions on `is_device_ptr` by converting the scalars to C addresses and then to one-element arrays.

It currently compiles and passes on the latest gfortran version on Summit. It will not compile with XLF.

### DIFF
--- a/tests/5.0/target_data/test_target_data_use_device_addr.F90
+++ b/tests/5.0/target_data/test_target_data_use_device_addr.F90
@@ -1,4 +1,4 @@
-!/===-- test_target_data_use_device_addr.c --------------------------------===//
+!/===-- test_target_data_use_device_addr.F90 ------------------------------===//
 !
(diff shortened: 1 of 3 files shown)


tests/5.0/target_data/test_target_data_use_device_addr.F90
```diff
@@ -0,0 +1,61 @@
+!/===-- test_target_data_use_device_addr.c --------------------------------===//
```
spophale, line 1: change to .F90

tmh97 (approved):

spophale (approved):

PR #353, opened 2021-06-23
Fix mapping for test_loop_collapse_device.c

tob2:
OpenMP 5.1 has: "A variable that is of type pointer, but not a function
pointer or (for C++) a pointer to a member function, is treated as if it
is the base pointer of a zero-length array section that had appeared as
a list item in a map clause."

However, the code relies on mapping more than a zero-length array or
only the pointer.

* tests/5.0/loop/test_loop_collapse_device.c (test_collapse1, test_collapse2):
  Actually map the whole array and not just a zero-length one.

NOTE: An alternative variant would be for `test_collapse1` (similar for the other function):
`a[:N],b[:N]` instead of `a[:N][:N],b[:N][:N+1]`

### DIFF
--- a/tests/5.0/loop/test_loop_collapse_device.c
+++ b/tests/5.0/loop/test_loop_collapse_device.c
@@ -34,7 +34,7 @@ int test_collapse1() {
     }
   }
 
-#pragma omp target parallel num_threads(OMPVV_NUM_THREADS_DEVICE) map(tofrom: a, b)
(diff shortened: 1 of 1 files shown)


tob2:
@spophale @tmh97  @jhdavis8  – please review.

jakubjelinek:
Note, zero length array section actually doesn't map anything but just does the pointer attachment to an existing mapping.
So it would work if there would map the whole arrays previously.

spophale:
Also true for OpenMP 5.0 -- Pg 315, lines: 1:2.

spophale (approved):

tmh97 (approved):

PR #359, opened 2021-07-19
Add sequential loop tile transformation test in C on the host

jhdavis8:
This test checks the 5.1 tile directive in a complete tiling context and a partial tiling context. The specification is very specific about how this directive should work. The compiler cannot reorder any two iterations where at least one of those iterations is from a complete tile. So, I created the first part of this test to check the actual order of iterations the compiler generates, and at present, it appears Clang allows the construct to be present but does not actually reorder the loop iterations, as far as I can tell. This means, at the moment, that it fails the test, but we could change it to a warning if we think that is more appropriate.

### DIFF
--- a/tests/5.1/tile/test_tile.c
+++ b/tests/5.1/tile/test_tile.c
@@ -1,0 +1,115 @@
+//===--- test_tile.c --------------------------------------------------------===//
+//
+// OpenMP API Version 5.1 Nov 2020
+//
+// This test checks the tile directive with required sizes clause. The tile
+// directive reorganizes `n` of the nested loops into `2n` tiled loops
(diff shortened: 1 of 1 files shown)


tests/5.1/tile/test_tile.c
```diff
@@ -0,0 +1,115 @@
...
+#pragma omp tile sizes(4, 4)
+  for (int i = 0; i < N; i++) {
+    for (int j = 0; j < N; j++) {
+      result[i][j] = time++;
+    }
+  }
+
+  time = 0;
+
+  /*
```
tmh97, line 49: remove this commented-out code?

spophale (approved):

tmh97 (approved):
good-to-go

PR #360, opened 2021-07-19
Add test of in_reduction nested in function from issue #50

### LINKED ISSUE #50: Task example from Jackub
**Directive and clause it applies to**

Tasks 

**Specification details**
From the implant email
The following is an example on Jakub’s blog using task reductions.

I don’t have the latest gcc handy, but Intel 19 does not accept this code, complaining that tasks in function bar don’t have an “associated task group.”  Should this compile? The spec states that “For a given a list item, the in_reduction clause deﬁnes a task to be a participant in a task reduction that is deﬁned by an enclosing region for a matching list item...”. I don’t see where we specify if it has to be lexically enclosed or if it can by dynamically enclosed, as the example implies.

(Blog article link: 
https://developers.redhat.com/blog/2019/03/19/whats-new-in-openmp-5-0/
)

Stephen
int r;

void bar (int i) {
  #pragma omp task in_reduction (+:r)
  r += work (i, 0);
  #pragma omp task in_reduction (+:r)
  r += work (i, 1);
}

int foo () {
  #pragma omp taskgroup task_reduction (+:r)
  bar (0);
  #pragma omp taskloop reduction (+:r)
  for (int i = 1; i < 4; ++i)
    { bar (i); r += i; }
}

jhdavis8:
See issue #50 for test idea. Checks that tasks with in_reduction can be nested in a function within a taskgroup. Currently passes both Clang and GCC on Summit. XL does not support in_reduction. Closes #50.

### DIFF
--- a/tests/5.0/task/test_task_in_reduction_dynamically_enclosed.c
+++ b/tests/5.0/task/test_task_in_reduction_dynamically_enclosed.c
@@ -1,0 +1,58 @@
+//===--- test_task_in_reduction_dynamically_enclosed.c ----------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test checks the task directive with the `in_reduction` reduction
+// participation clause. It performs simple array operations which are added
+// to a reduction variable in an explcit task with the in_reduction clause.
+// This checks the above in a case in which the in_reduction tasks are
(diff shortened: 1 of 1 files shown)


tmh97 (approved):
Ready to go

spophale (approved):

spophale:
TODO: Add device version.

PR #362, opened 2021-07-19
Add test of parallel reduction with two elements of the same array from issue #151

### LINKED ISSUE #151: This breaks clang
**Directive and clause it applies to**
reduction inside target 

**Specification details**
Having the reduction on two different elements of the same array breaks clang unsuccessfully. I am not 100% sure this is allowed by the spec, but if so it would make a nice test. 

**Pseudocode**
```
static double temps[2];

int main () {
   #pragma omp target
   {
       #pragma omp parallel reduction(+:temps[0], temps[1])
       {
           temps[0] += 1;
           temps[1] += 1;
       }
   }
   return 0;
}
```
**Any other comments**
None

jhdavis8:
See issue #151. This is a simple test of a reduction on the device with two array elements from the same array. Here are the per-compiler results:

Clang:
```
/tmp/test_parallel_reduction_nested-d6a240.o: In function `main':
test_parallel_reduction_nested.c:(.text+0x3c): undefined reference to `omp_is_initial_device'
/usr/bin/ld: link errors found, deleting executable `bin/test_parallel_reduction_nested.c.o'
clang-13: error: linker command failed with exit code 1 (use -v to see invocation)
```

GCC:
```
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/parallel_reduction/test_parallel_reduction_multiple_array_elements.c: In function 'main':
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/parallel_reduction/test_parallel_reduction_multiple_array_elements.c:28:31: error: 'temps' appears more than once in data clauses
   28 | #pragma omp parallel reduction(+:temps[0], temps[1])
      |                               ^
```

XL:
```
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/parallel_reduction/test_parallel_reduction_multiple_array_elements.c:28:44: error: variable can appear only once in OpenMP 'reduction' clause
#pragma omp parallel reduction(+:temps[0], temps[1])
                                           ^
/autofs/nccs-svm1_home1/jhdavis/sollve-project/sollve_vv/tests/5.0/parallel_reduction/test_parallel_reduction_multiple_array_elements.c:28:34: note: previously referenced here
#pragma omp parallel reduction(+:temps[0], temps[1])
                                 ^
```

Closes #151.

### DIFF
--- a/tests/5.0/parallel_reduction/test_parallel_reduction_multiple_array_elements.c
+++ b/tests/5.0/parallel_reduction/test_parallel_reduction_multiple_array_elements.c
@@ -1,0 +1,38 @@
+//===--- test_parallel_reduction_nested.c -----------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This is a simple test of the parallel directive with a reduction clause
+// which performs a reduction on two different elements of the same array,
+// in an offloaded region.
+//
+////===----------------------------------------------------------------------===//
+#include <assert.h>
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+static double temps[2];
+
+int main() {
+  int errors = 0;
+
+  OMPVV_TEST_OFFLOADING;
+
+  temps[0] = 0;
+  temps[1] = 0;
+
+#pragma omp target map(tofrom: temps)
+  {
+#pragma omp parallel reduction(+:temps[0], temps[1])
+    {
+      temps[0] += 1;
+      temps[1] += 1;
+    }
+  }
+
+  OMPVV_TEST_AND_SET_VERBOSE(errors, temps[0] != 1 || temps[1] != 1);
+
+  OMPVV_REPORT_AND_RETURN(errors);
+}


tests/5.0/parallel_reduction/test_parallel_reduction_multiple_array_elements.c
```diff
@@ -0,0 +1,38 @@
...
+#pragma omp target map(tofrom: temps)
+  {
+#pragma omp parallel reduction(+:temps[0], temps[1])
+    {
+      temps[0] += 1;
+      temps[1] += 1;
+    }
+  }
+
+  OMPVV_TEST_AND_SET_VERBOSE(errors, temps[0] != 1 || temps[1] != 1);
```
tob2, line 35: * If I use `reductions(+:temps)` as proposed variant in Issue 151 (i.e. #151), I get with GCC 12 and LLVM/Clang 11 and host fallback temps[0] == temps[1] == 12. — Thus, the assumption that the parallel region has only one thread seems to be far fetched.

* I think it makes sense to add this full-array variant as Issue 151 states that a compiler gave `nan` in this case, which indicates a worthwhile check. I do note that the OpenMP spec (here: 5.2) states under "5.5.5 Properties Common to All Reduction Clauses":

"If a list item is an array or array section, it will be treated as if a reduction clause would be applied to each separate element of the array section."

Thus, using `reduction(+:temps)` should be valid.

* I have not tried to understand whether `reduction(+:temps[0], temps[1])` is valid or not. However, the restriction "Any array element must be specified at most once in all list items on a directive." implies that it is valid. (From "5.5.5 Properties Common to All Reduction Clauses" in OpenMP)

tob2: > * I have not tried to understand whether `reduction(+:temps[0], temps[1])` is valid or not. However, the restriction "Any array element must be specified at most once in all list items on a directive." implies that it is valid. (From "5.5.5 Properties Common to All Reduction Clauses" in OpenMP)

Jakub believes:
> No, IMNSHO this is invalid.
> `reduction` is a privatization clause, and one can't privatize two elements of  the same variable separately.  It is valid to do `#pragma omp parallel reduction(+:temps[:2])`, but not `reduction(+:temps[0], temps[1])`

tob2: Regarding Jakub's comments (cf. OpenMP 5.2), `reduction` belongs to "Properties: privatization" and in the glossary item "private variable" reads "... A variable that is part of another variable (as an array element or a structure element) cannot be made private independently of other components. ...".

_(Regarding the misleading spec bit, I filled the OpenMP spec issue 3187.)_

PR #364, opened 2021-07-20
Addressing issue #105 and fixing header of random file

### LINKED ISSUE #105: NERSC Virtual methods tests
Email from Rahulkumar Gayatri

```
Hi Jose and Chris,
Following our earlier conversation, here are the 2 test cases for virtual and derived class instances on the device using OpenMP offload.
test_virtual1.cxx - only derived class object accessed on the device (Written by Chris)
test_virtual1.cxx - both base and derived class objects are accessed on the device (slight modification to the first test case)

There is also a makefile that you can edit depending on the compiler.

Attached is the tar file for these tests.

Regards,
Rahul.
```

- [ ] These were created as a proof of concept for Josh Meyers (copied) and the ImSim project. The eventual code will probably look very different.
- [ ] Add to the test head Also Chris Daily and Rahulkumar Gayatri from NERSC for their suggestion

Chris

[OpenMPOffloadVirtualTest.tar.gz](https://github.com/SOLLVE/sollve_vv/files/4256858/OpenMPOffloadVirtualTest.tar.gz)

UPDATED VERSION
[create-object-on-device-2.zip](https://github.com/SOLLVE/sollve_vv/files/4295552/create-object-on-device-2.zip)

tmh97:
Closes #105. 

One of the tests uses a derived class while the other does not. Additionally, virtual functions are used in the test that utilizes derived classes.

Currently passes on latest llvm but fails with gcc/10.2.0, need to test on newer gcc.


tmh97:
All tests pass on latest xlc++ and clang++ compilers. Fails on gcc/10.2.0

spophale:
@tmh97 since there is no (easy) way to go back to the 4.5 implementations to test the codes, could you move these to 5.0/application-kernels ? I did a quick check of the 5.0 spec and I don't see any restrictions that invalidate these use cases.

spophale (approved):

tmh97:
> @tmh97 since there is no (easy) way to go back to the 4.5 implementations to test the codes, could you move these to 5.0/application-kernels ? I did a quick check of the 5.0 spec and I don't see any restrictions that invalidate these use cases.

@spophale Just moved 'em

nolanbaker31 (approved):
looks good from what i can tell

Issue #81, opened 2020-01-31
Tests in C that are missing Fortran counterparts


jhdavis8:
We have a number of tests that are in C but still need to be written in Fortran:

- [ ] test_target_enter_exit_data_depend.c
- [ ] test_target_enter_data_depend.c
- [ ] test_target_update_if.c
- [ ] test_target_update_devicce.c
- [ ] test_target_update_depend.c
- [ ] test_target_update_to.c
- [ ] test_target_update_from.c
- [ ] test_target_teams_distribute_parallel_for_if.c
- [ ] test_target_teams_distribute_parallel_for_devices.c
- [ ] test_target_teams_distribute_parallel_for_private.c
- [ ] test_target_teams_distribute_parallel_for_fristprivate.c
- [ ] test_target_teams_distribute_parallel_for_map_defailt.c
- [ ] test_target_teams_distribute_parallel_for_map_to.c
- [ ] test_target_teams_distribute_parallel_for_map_tofrom.c
- [ ] test_target_teams_distribute_parallel_for_map_from.c 
- [ ] test_target_teams_distribute_parallel_for_defaultmap.c
- [ ] test_target_teams_distribute_parallel_for_depends.c
- [ ] test_target_teams_distribute_parallel_for_num_teams.c
- [ ] test_target_teams_distribute_parallel_for_thread_limit.c
- [ ] test_target_teams_distribute_parallel_for_num_threads.c
- [ ] test_target_parallel.c
- [ ] test_target_simd_collapse.c

seyonglee:
Update the list with the latest status (as of Jan. 21, 2022)

tmh97:
Hey @seyonglee I believe the list is up to date. We do have a few 4.5 fortran90 tests pending, once those are approved + merged I will update this list and ping you, thanks!

tmh97:
> Update the list with the latest status (as of Jan. 21, 2022)

Ignore my previous comment! Thank you for updating this @seyonglee :)

PR #369, opened 2021-07-27
Adding 3 new .F90 tests per issue #81 and fixing .c files

tmh97:
- 3 new tests addressing issue #81 
- Updating a few .c files
- test_target_teams_distribute_parallel_for_devices.F90 test fails on cray compiler due to device clause being ignored

All tests pass with nvhpc/21.11!

### DIFF
--- a/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_devices.F90
+++ b/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_devices.F90
@@ -51,7 +51,7 @@ CONTAINS
       DO dev = 1, num_dev
       !$omp target teams distribute parallel do device(dev) map(tofrom: isHost)
             DO i = 1, N
-               IF ((omp_get_team_num() .eq. 0) .or. (omp_get_thread_num() .eq. 0)) THEN
+               IF ((omp_get_team_num() .eq. 0) .and. (omp_get_thread_num() .eq. 0)) THEN
                   isHost(dev) = omp_is_initial_device()
                END IF
                a(i) = a(i) + dev
--- a/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_devices.F90
+++ b/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_devices.F90
@@ -1,0 +1,79 @@
+!===--- test_target_teams_distribute_parallel_for_devices.F90 ---------------===//
+!
+! OpenMP API Version 4.5 Nov 2015
+! 
+! Testing for multiple devices checking if it is possible to send work and data 
+! to different devices with the device clause used with omp target teams distribute 
+! parallel for
+! 
+!===-------------------------------------------------------------------------===//
+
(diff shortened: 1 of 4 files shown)


tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_devices.F90
```diff
@@ -0,0 +1,79 @@
...
+      END DO
+
+      DO dev = 1, num_dev
+         !$omp target enter data map(to: a) device(dev)
+      END DO
+
+      DO dev = 1, num_dev
+      !$omp target teams distribute parallel do device(dev) map(tofrom: isHost)
+            DO i = 1, N
+               IF ((omp_get_team_num() .eq. 0) .or. (omp_get_thread_num() .eq. 0)) THEN
```
spophale, line 54: Should be an "and" else all teams tread 0 will try to assign value.

spophale (commented):
Except test_target_teams_distribute_parallel_for_devices.F90 all LGTM.

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_private.F90
```diff
@@ -0,0 +1,77 @@
...
+            d(j) = c(j) * privatized
+         END DO
+      !$omp end target data
+
+      warning_threads = 0
+      warning_teams = 0
+
+      DO i = 1, N
+         OMPVV_TEST_AND_SET(errors, d(i) .ne. ((1 + i) * 2 * i))
+         warning_teams = warning_teams + num_teams(i)
```
nolanbaker31, line 68: Why would you use warning_teams = warning_teams + num_teams(i) if you previously just set warning_teams = 0? (Still correct, just curious)

nolanbaker31 (approved):
lgtm (still need to edit test_target_teams_distribute_parallel_for_devices to include 'and' instead of 'for')

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_private.F90
tmh97: If this was not inside a for loop I would just do 'warning_teams = num_teams(i)', but since it is nested inside I want to update the value of warning teams N number of times

tmh97:
test_target_teams_distribute_parallel_for_devices.F90 fails, device clause not recognized. All other tests pass.

tmh97:
Need second set of eyes on test_target_teams_distribute_parallel_for_devices.F90, currently fails with errors = 1025

tmh97:
The first test in this PR, tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_devices.F90, acquires the # of devices available and then proceeds to loop from 1 to number_of_devices and perform some operations. Test passes with latest nvfortran, but with gfortran it seems like the last device (4th device in this case since we are testing on summit which has 4 V100s) is treated like a host and fails. @tob2 any thoughts on why this behavior may occur?

PR #373, opened 2021-08-05
Adding test_target_enter_exit_data_depend.F90 per issue #81

tmh97:
The two integer functions defined in this test pass when called individually, however, a segmentation error occurs when both are called in the same program.

All tests pass with gcc/11.1.0 on summit


tmh97:
Also includes changes to file related to removing OMPVV_IS_SHARED_ENV

spophale (commented):
I tested test_target_enter_exit_data_depend.F90 with gfortran that came with gcc/11.1.0-2 on Summit. 
I get "Test failed on the device."

spophale (approved):

nolanbaker31 (approved):

PR #377, opened 2021-08-06
Adding target_teams_distribute_parallel_for_if_no_modifier.F90 per is…

tmh97:
Currently fails with errors = 1025, I need to investigate further.

### DIFF
--- a/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_if_no_modifier.F90
+++ b/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_if_no_modifier.F90
@@ -1,0 +1,141 @@
+!===--- test_target_teams_distribute_parallel_for_if_no_modifier.F90 ---------===!
+!
+! OpenMP API Version 4.5 Nov 2015
+!
+! In this test we want to try to check if the if clause is working 
+! when used with the combined construct target teams distribute
+! parallel for and no modifier is specified.
+! To do this we check if offloading is working, if it is not, it won't
+! be possible for us to tell if the test passed or not, since it depends
+! on offloading capabilities.
+!
+! The if has a directive-name-modifier that specifies to which directive
+! the if applies to (either both directives, to the target or to the parallel). 
+! We create three tests, one for no directive, another one for the target 
+! directive and another one for the parallel directive.
+!
+!===--------------------------------------------------------------------------===!  
+
+#include "ompvv.F90"
+
+#define ATTEMPT_THRESHOLD 70
+#define NUM_ATTEMPTS 100
+#define N 1024
+
+PROGRAM test_target_teams_distribute_parallel_for_if_no_modifier
+   USE iso_fortran_env
+   USE ompvv_lib
+   USE omp_lib
+   implicit none
+
+   OMPVV_TEST_VERBOSE(target_teams_distribute_if_no_modifier() .ne. 0)
+   OMPVV_REPORT_AND_RETURN()
+
+
+CONTAINS 
+   LOGICAL FUNCTION checkPreconditions()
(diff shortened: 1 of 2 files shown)


tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_if_no_modifier.F90
```diff
@@ -0,0 +1,141 @@
...
+
+      isOffloading = checkPreconditions()  
+         
+      DO i = 1, N
+         a(i) = 1
+         warning(i) = 0
+      END DO
+ 
+      DO attempt = 1, NUM_ATTEMPTS
+         !$omp target teams distribute parallel do if(attempt .ge. ATTEMPT_THRESHOLD) &
```
tmh97, line 100: Simplify, if clause is true: ensure it is running on device, produce error otherwise. Still map and modify a

tmh97:
Simplify this and the related PR

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_if_no_modifier.F90
seyonglee: It looks OK. 
One very minor issue is that the behavior of the Fortran version is different from that of the C version.
The C version offloads the compute region NUM_ATTEMPS - ATTEMPT_THRESHOLD times out of NUM_ATTMPTS trials, but the Fortran version offloads the compute region  NUM_ATTEMPS - ATTEMPT_THRESHOLD times out of NUM_ATTMPTS - 1 trials. (The Fortran version executes the compute region serially on the host one less times than the C version.)
One way to make the Fortran version behave exactly same as the C version would be to change the if condition in line 100 to "attempt .gt. ATTEMPT_THRESHOLD" and revert "NUM_ATTEMPTS - 1" in line 99 back to "NUM_ATTEMPTS".

seyonglee (approved):

seyonglee:
On Summit, IBM XLF compiler (16.1.1-10) fails because of the compiler bugs in handling macro combined with line continuation, which is an XLF compiler bug. If we remove the line continuation by merging multiple lines into a single one, it will pass the test with XLF too.
However, having long lines is not a good practice for Fortran because some Fortran compilers may still have line length limits.

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_if_no_modifier.F90
tmh97: @seyonglee thank you for this! Making the change now

tmh97:
> On Summit, IBM XLF compiler (16.1.1-10) fails because of the compiler bugs in handling macro combined with line continuation, which is an XLF compiler bug. If we remove the line continuation by merging multiple lines into a single one, it will pass the test with XLF too. However, having long lines is not a good practice for Fortran because some Fortran compilers may still have line length limits. One possible solution to detour the XLF compiler bug, which is NOT related to the OpenMP features, is to use multiple prints with short messages, instead of having a single print with a long message.

@spophale do we want to change the print statements to cater to XL or shall we leave as is?

tob2:
> However, having long lines is not a good practice for Fortran because some Fortran compilers may still have line length limits. 

Well, Fortran 2018 itself still has: "If a line consists entirely of characters of default kind (7.4.4), it shall contain at most 132 characters" for free-form source form.
Only Fortran 202x will have: "A line shall contain at most ten thousand characters."

Thus, pedantically, it is not Fortran-compilers problem but a language problem before 202x. But most compilers permit to increase the line-length limit.

spophale (approved):

PR #382, opened 2021-08-09
Adding test for target_teams_distribute_parallel_for_if with target modifier

tmh97:
Passes on gcc 11

Addresses issue #81

### DIFF
--- a/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_if_target_modifier.F90
+++ b/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_if_target_modifier.F90
@@ -79,7 +79,7 @@ CONTAINS
       
       raiseWarning = 0
       DO i = 1, N
-         OMPVV_TEST_AND_SET(errors, a(i) .ne. (1 + (ATTEMPT_THRESHOLD)))
+         OMPVV_TEST_AND_SET(errors, a(i) .ne. (ATTEMPT_THRESHOLD))
          IF (warning(i) .ne. 0) THEN
             raiseWarning = 1
          END IF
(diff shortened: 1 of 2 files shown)


tmh97:
@seyonglee Fixed this test to match behavior of the .c version (swapped .ge. with .gt.)

seyonglee:
It looks good; I also checked that the test passes with both GCC11.1.0 and XLF16.1.1-10.

seyonglee:
One suggestion: how about changing the line 67 of the Fortran version as following (change similarly in the C version too):
//From 
IF ((isOffloading .eqv. .TRUE.) .and. (omp_is_initial_device() .eqv. .true.)) THEN
//To
IF ((isOffloading .eqv. .FALSE.) .or. (omp_is_initial_device() .eqv. .true.)) THEN

The current condition checks only "offloaded but on a host device" case, but I think that "not-offloaded" is also an error.
Currently, the test prints a warning message for the "not-offloaded" case, but test will pass anyway. I think that it will be better to consider the "not-offloaded" case as an error too, since it will make it easier for users to detect the problem.

spophale (approved):

nolanbaker31 (approved):
looks good, i agree with seyong's change but this can be discussed

spophale:
@seyonglee I don't agree that if it does not offload it is an error. If for any reason the offload does not happen, fallback is legal and should not generate an error.

seyonglee:
@spophale As mentioned in the comment, if offloading is not working, it won't be possible to tell if the test passed or not, since it depends on offloading capabilities. If we cannot test  the intended feature, it may be better to say that this test fails (error)?

spophale:
The way I read the code, based on the threshold, some iterations will be offloaded and some will not. That is what "if" should do. When it is not supposed to be offloaded (but offload support is there), the variable is modified differently to capture this. Changing the conditions on ln 67 will lead to incorrect updates.

seyonglee:
The original condition in line 67 checks whether `if` clause works correctly or not, but the new condition will check both `if` clause and `offloading`. The question goes back to the base question: whether we should consider `not offloadable` as an error or not.

PR #385, opened 2021-08-11
Adding .F90 test for test_target_update_devices.F90

tmh97:
Test passes on spock with cray (ftn version 12.0.1) and on summit with gcc (10.2.0).

Potential bug with xl 16.01
Screen Shot 2021-08-11 at 12 05 22 AM

### DIFF
--- a/tests/4.5/target_update/test_target_update_devices.F90
+++ b/tests/4.5/target_update/test_target_update_devices.F90
@@ -1,0 +1,185 @@
+!===--- test_target_update_devices.F90 --------------------------------------===//
+!
+! OpenMP API Version 4.5 Nov 2015
+!
+!
+! This test checks if the target update directive works on different devices.
+! We check two different variants.
+! 1. setting up the default device with the API call omp_set_default_device()
+! 2. using the device clause of the target update directive.
+!
+! Testing metodology uses an array that gets mapped into the device at first
+! through target enter data. Then on each iteration we update the array in one
+! device, create a compute region in that device, and then update it back
+! We also record that the compute region is not executed in the host
+! with the omp_is_initial_device() API call. Unfortunately 4.5 has no device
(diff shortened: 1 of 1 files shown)


nolanbaker31 (approved):
lgtm

spophale:
Curious, did you try on host only ?

spophale (commented):
Check if host fallback works correctly with this test (ideally with LLVM).

tmh97:
@tmh97 check host fallback / no_offloading enabled

tests/4.5/target_update/test_target_update_devices.F90
```diff
@@ -0,0 +1,185 @@
...
+         CALL omp_set_default_device(dev)
+         !$omp target update to(h_matrix)
+         !$omp target map(alloc: h_matrix) map(tofrom: isHost(dev))
+            isHost(dev) = omp_is_initial_device()
+         
+            DO i = 1, N
+               h_matrix(i) = h_matrix(i) + 1
+            END DO
+          !$omp end target
+         !$omp target update from(h_matrix)
```
spophale, line 81: overwrite

tests/4.5/target_update/test_target_update_devices.F90
```diff
@@ -0,0 +1,185 @@
...
+      END DO
+
+      ! Unmap the matrix
```
spophale, line 84: Move unman to after checking

tests/4.5/target_update/test_target_update_devices.F90
tmh97, line 81: Move update from to outside this do loop, map sure h_matrix is not getting overwritten, and value of hmatrix is checked on the host for each available device

spophale (approved):

PR #386, opened 2021-08-11
Atomic Compare Test

nolanbaker31:
Does not result in errors, but I'm unsure if implementation properly tests atomic compare, as without it would still most likely not give errors.

### DIFF
--- a/tests/5.1/atomic/test_atomic_compare.c
+++ b/tests/5.1/atomic/test_atomic_compare.c
@@ -1,0 +1,48 @@
+//===--- test_atomic_compare.c ---===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// Creates an array with random numbers, and uses atomic compare to find the max,
+// testing against non-parallel maximum.
+//
+////===----------------------------------------------------------------------===//
+#include <assert.h>
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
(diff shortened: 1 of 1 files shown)


tmh97:
Tentative: May need to give credit to OpenMP examples doc

tests/5.1/atomic/test_atomic_compare.c
```diff
@@ -0,0 +1,48 @@
...
+
+   for(int i=0; i<N; i++){
+      arr[i] = rand()%1000;
+   }
+   for(int i = 0; i<N; i++){ // Sets max through non-parallel methods
+      if(arr[i] > max){
+         max = arr[i];
+      }
+   }
+   #pragma omp parallel for // Sets max using parallel for loop, using atomic to ensure max is correct
```
tmh97, line 32: pmax should be a shared variable

tmh97: After making this change `#pragma omp paralle for shared(pmax)` and testing with cray, clang, and gcc this should be good to go.

tmh97:
@nolanbaker31 changes look good, were you able to test this on a few different compilers?

spophale (approved):
I think it does stress the atomic else the max value would be overwritten and the test would give incorrect answers. Such tests need to be run more than once to catch race (if atomic is not correctly implemented).
Also need a device version.

PR #387, opened 2021-08-11
Adding .F90 test for test_target_teams_distribute_parallel_for with defaultmap per #81

tmh97:
Passes with XL but fails with GCC. It seems the two have different rules for implicit mapping of character variables in the second function test_defaultmap_off().

***C treats chars as scalar variable while Fortran does not ---> according to 4.5 spec***

### DIFF
--- a/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_defaultmap.F90
+++ b/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_defaultmap.F90
@@ -136,7 +136,7 @@ CONTAINS
          scalar_double = 20.45
       END DO
       
-      OMPVV_TEST_AND_SET_VERBOSE(errors, scalar_char .ne. 'a') ! Character type is not a scalar variable in fortran
+      OMPVV_TEST_AND_SET_VERBOSE(errors, scalar_char .ne. 'b') ! The Fortran character type is not a scalar variable in OpenMP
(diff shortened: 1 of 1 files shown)


tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_defaultmap.F90
```diff
@@ -0,0 +1,147 @@
...
+
+      DO i = 1, ITERATIONS
+         OMPVV_TEST_AND_SET_VERBOSE(errors, scalar_char_cpy(i) .ne. 'a')
+         OMPVV_TEST_AND_SET_VERBOSE(errors, scalar_short_cpy(i) .ne. 10)
+         OMPVV_TEST_AND_SET_VERBOSE(errors, scalar_int_cpy(i) .ne.  11)
+         OMPVV_TEST_AND_SET_VERBOSE(errors, ABS(scalar_float_cpy(i) - 5.5) .gt. 0.0001);
+         OMPVV_TEST_AND_SET_VERBOSE(errors, ABS(scalar_double_cpy(i) - 10.45) .gt. 0.00001);
+      END DO
+
+      ! Map the same array to multiple devices. Initialize with device number
```
spophale, line 67: For this, device clause will have to be used on target directive.

spophale (approved):

nolanbaker31 (approved):

seyonglee (approved):
It looks OK to me.

PR #390, opened 2021-08-17
Defaultmap with Present Clause

nolanbaker31:
Need to test - in progress. So far just using 3 data types and ensuring that they are the same on target region.

### DIFF
--- a/tests/5.1/defaultmap/test_defaultmap_present.c
+++ b/tests/5.1/defaultmap/test_defaultmap_present.c
@@ -37,14 +37,14 @@ int test_defaultmap_present() {
    new_struct.s = 10; new_struct.S[0] = 10; new_struct.S[1] = 10;
    ptr = &A[0]; 
    ptr[50] = 50; ptr[51] = 51;
-
-   #pragma omp target defaultmap(present) // Present means values should remain as if already declared in map as alloc
+   #pragma omp target map(tofrom: errors) \
+   defaultmap(present) // Present means values should remain as if already declared in map as alloc
    {     
-      OMPVV_TEST_AND_SET_VERBOSE(errors, scalar != 1);
-      OMPVV_TEST_AND_SET_VERBOSE(errors, A[0] != 0 || A[50] != 50);
-      OMPVV_TEST_AND_SET_VERBOSE(errors, A[50] != 50 || A[51] != 51);
-      OMPVV_TEST_AND_SET_VERBOSE(errors, new_struct.s != 10);
-      OMPVV_TEST_AND_SET_VERBOSE(errors, new_struct.S[0] != 10);
+      if(scalar != 1){errors++;}
+      if(A[0] != 0 || A[50] != 50){errors++;}
+      if(A[50] != 50 || A[51] != 51){errors++;}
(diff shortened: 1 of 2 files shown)


tests/5.1/defaultmap/test_defaultmap_present.c
```diff
@@ -0,0 +1,57 @@
...
+   int A[N]; // aggregate 
+   struct test_struct new_struct; // aggregate
+   int *ptr; // scalar, pointer 
+
+   scalar = 1; 
+   A[0] = 0; A[50] = 50;
+   new_struct.s = 10; new_struct.S[0] = 10; new_struct.S[1] = 10;
+   ptr = &A[0]; 
+   ptr[50] = 50; ptr[51] = 51;
+
```
spophale, line 40: Add a target enter data with the appropriate maps.

tmh97 (approved):
After adding target enter data and testing with latest gcc (summit), clang (summit), ibm (summit), and cray (spock) this is ready to merge.

tests/5.1/defaultmap/test_defaultmap_present.c
tmh97: will add target enter data map with alloc map type modifier and see what happens

tmh97:
@tmh97 try to remove new features and get this to compile on tulip/cori without issue

tests/5.1/defaultmap/test_defaultmap_present.c
tmh97: Test fails with #pragma omp target enter data map(alloc: etc)

tests/5.1/defaultmap/test_defaultmap_present.c
```diff
@@ -0,0 +1,57 @@
...
+   #pragma omp target defaultmap(present) // Present means values should remain as if already declared in map as alloc
+   {     
+      OMPVV_TEST_AND_SET_VERBOSE(errors, scalar != 1);
```
tmh97, line 43: @nolanbaker31 Issue with test running is that OMPVV_TEST_AND_SET_VERBOSE cannot be used in a target region since it utilizes fprintf under the hood

tmh97: @nolanbaker31 Instead, will need to map errors variable to from and simply have nested if statements that execute errors++ when required

tmh97:
Currently, this test fails with 6 errors on summit. 

- The setup as of now checks that the variables are **not** equal to the value there we defined as on the host. 
- I.e if scalar == 70, we are checking on device that scalar != 70. 
- Considering that defaultmap(present) maps variable types with alloc modifier, I believe they should never have the same value on the device as they have on the host, instead, they should be uninitialized.

A proper way to test would be to check if they **are** equal to the same value as previously defined on the host, in that case, we would increment errors by 1.
- I implemented this but still see that line 45 fails for both conditions (A[0] == 0) and (A[50] == 50). For some reason, the values of A[0] and A[50] are mapped (to or tofrom), but the indices of the same array that are modified by the pointer are mapped uninitalized as expected.

spophale (approved):

PR #391, opened 2021-08-20
Adding tests to check hint specified by numeric value.

spophale:
Tested and works with LLVM 13 on Summit.


tests/5.0/atomic/test_atomic_num_hint_device.c
```diff
@@ -0,0 +1,72 @@
+//===--- test_atomic_hint.c -------------------------------------------------===//
```
spophale, line 1: change to _device

tmh97 (approved):

nolanbaker31 (approved):

PR #392, opened 2021-08-24
C++ Attribute Specifier Test

nolanbaker31:
On GCC, currently gives error 'warning: attributes at the beginning of statement are ignored [-Wattributes]'.
On Clang/XL, gives error 'parallel', etc. 'was not declared in this scope'

I'm not sure if the implementation is incorrect or if compilers do not support this feature yet.

### DIFF
--- a/tests/5.1/test_cpp_attribute_specifier.cpp
+++ b/tests/5.1/test_cpp_attribute_specifier.cpp
@@ -1,0 +1,38 @@
+//===--- test_target_defaultmap_default.c --------------------------------------------------------------===//
+//
+//  OpenMP API Version 5.1 NOV 2021
+//
+//  This test simply uses a for loop, and tests multiple times that it is not running in parallel. Then,
+//  using C++ OMP attribute specifiers, creates a parallel for loop and ensures that it is running in
+//  parallel.
+//
+////===-------------------------------------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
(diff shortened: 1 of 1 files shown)


tests/5.1/test_cpp_attribute_specifier.cpp
```diff
@@ -0,0 +1,38 @@
+//===--- test_target_defaultmap_default.c --------------------------------------------------------------===//
```
nolanbaker31, line 1: change name

tob2:
For **GCC,** **OpenMP 5.1 C++ attributes** support was only added recently in **mainline** (the current development branch), already known as **GCC 12.0.0** but will be **released next spring.** Cf. https://gcc.gnu.org/gcc-12/changes.html _(slightly out of date as some more OMP 5.1 features have been implemented in the meanwhile.)_

Using GCC mainline _does_ compile your program without warnings - and running shows here: "Test passed on the host.". – I have not otherwise checked your program.

nolanbaker31:
Since Tobias has confirmed test passes, and replacing the C++ attribute with the regular functions themselves works properly, test should be ready to merge.

tmh97:
hey @nolanbaker31 once you change the name I will approve!

nolanbaker31:
@tmh97 I somehow always change the description but not the name🤦‍♂️ it has been fixed!!

tmh97 (approved):

Issue #393, opened 2021-08-27
tests/5.0/target/test_target_device.c does not follow the OpenMP 5.0 specification (ancestor restrictions)


marvol77:
The function `test_target_device_ancestor` in the file `tests/5.0/target/test_target_device.c` uses a `target` directive with the device-modifier `ancestor` which was introduced in OpenMP 5.0.

The [specification of OpenMP 5.0](https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5.0.pdf) (and also OpenMP 5.1) states the following restrictions when using `ancestor` (p. 174f):

> If a device clause in which the ancestor device-modifier appears is present on the construct, then the following restrictions apply:
> – A requires directive with the reverse_offload clause must be specified;
> [...]
> – No OpenMP constructs or calls to OpenMP API runtime routines are allowed inside the corresponding target region.

Thus, it should be added `#pragma omp requires reverse_offload` and the usage of the OpenMP API runtime routine `omp_is_initial_device()` is not allowed. Otherwise a compile error can be expected for the test. 

The second issue holds also for `tests/5.0/requires/test_requires_reverse_offload.c` where `omp_is_initial_device` is indirectly (via `host_function`) used in the target region of the `#pragma omp target device(ancestor:1)` construct.

tob2:
Question: Is there still something to solve or not?

I see two PR which are both merged, #394 (test_target_device.c) and #396 (test_requires_reverse_offload.c)

Is there anything else? If not, can this issue be closed?

tmh97:
@tob2 this is solved, closing this issue now. Thank you for checking in on this !

PR #394, opened 2021-08-31
Issue 393: Fix restriction violations on test_target_device.c

tob2:
Issue #393
* tests/5.0/target/test_target_device.c: Add 'requires reverse_offload'
  required by 'device(ancestor:)' + avoid using an OpenMP runtime
  routine inside that target region.

@spophale , @tmh97, @jhdavis8 – please review.

Cf. OpenMP 5.0 spec, 11th bullet point under 'Restrictions' in '2.12.5  target Construct':
https://www.openmp.org/spec-html/5.0/openmpsu60.html#x86-2880002.12.5

### DIFF
--- a/tests/5.0/target/test_target_device.c
+++ b/tests/5.0/target/test_target_device.c
@@ -15,6 +15,9 @@
 
 #define N 1028
 
+// Required for 'device(ancestor: 1)
+#pragma omp requires reverse_offload
+
 int test_target_device_ancestor() {
 
     int i, which_device;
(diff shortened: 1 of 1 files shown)


tmh97:
@tob2 thank you for addressing this issue, greatly appreciated!

tests/5.0/target/test_target_device.c
```diff
@@ -36,12 +39,15 @@ int test_target_device_ancestor() {
...
+	    // "No OpenMP constructs or calls to OpenMP API runtime routines are allowed"
+	    // Thus, settle on the following. Note: With unified-shared memory, it is
+	    // not actually testing that this target region is executed on the initial device.
+	    which_device = 42;
 	}
     }
 
-    OMPVV_TEST_AND_SET(errors, which_device != 1);
+    OMPVV_TEST_AND_SET(errors, which_device != 42);
     OMPVV_ERROR_IF(which_device != 1, "Target region was executed on device. Due to ancestor device-modifier,"
```
spophale, line 51: Needs to be removed.

spophale:
@tob2 I agree with the fix. Correspondingly, ln 51 needs to be removed IMO.

tests/5.0/target/test_target_device.c
tob2: I have now removed it – and accordingly updated the comment/which_device assignment.

nolanbaker31 (approved):
Change makes sense according to spec requirements for ancestor

spophale (approved):

PR #395, opened 2021-09-16
Making small change to test_target_device.c

tmh97:
Test currently fails on summit with latest gcc and llvm due to omp_get_device_num() routine.


nolanbaker31 (approved):
lgtm

spophale (approved):

PR #397, opened 2021-09-20
test_atomic_compare.c: Init pmax variable

tob2:
* tests/5.1/atomic/test_atomic_compare.c (test_atomic_compare):
  init pmax variable to 0.

Here, I got spurious fails, where pmax = 32756 and max = 996 – when pmax uninitialized value happened to be >max. As `max` is initialized to 0, do the same for pmax.

@spophale, @tmh97, @nolanbaker31, @jhdavis8 – please review. Thanks!

### DIFF
--- a/tests/5.1/atomic/test_atomic_compare.c
+++ b/tests/5.1/atomic/test_atomic_compare.c
@@ -20,7 +20,7 @@ int test_atomic_compare() {
 
   int arr[N];
   int errors = 0;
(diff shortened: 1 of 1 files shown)


nolanbaker31 (approved):
looks good - appropriate fix. pmax should have been 0, not sure how this was not found in my testing
thank you tobias

tmh97 (approved):

PR #399, opened 2021-09-23
Atomic Compare Device

nolanbaker31:


tests/5.1/atomic/test_atomic_compare_device.c
```diff
@@ -0,0 +1,50 @@
...
+#define N 100
+
+int test_atomic_compare() {
+  OMPVV_INFOMSG("test_atomic_compare");
+
+  int arr[N];
+  int errors = 0;
+  int max = 0, pmax x= 0;
+
+   for(int i=0; i<N; i++){
```
tmh97, line 25: make all for loop style look like similar to line 28

PR #401, opened 2021-09-28
Map w/ Present Clause test

nolanbaker31:


tmh97:
Break into a few simple tests:

First test should be: #pragma omp target map(present, delete: scalar_var) map(to: scalar)
Second test should be: #pragma omp target map(present, from: scalar_var) map(to: scalar)

PR #403, opened 2021-10-05
Target Update w/ Motion Modifier Present

nolanbaker31:
Very rough test of motion_present 5.1 test.

Motion-modifier 'present' can be used for both to & from, **should both be used**?

Also, SPEC 5.1 (https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5-1.pdf#page=229) lists how " For a list item that is replaced with a set of list items as a result of a user-defined mapper, the present modifier only applies to those mapper list items that share storage with the original list item." **Should this be tested or is it a minuscule case?**

### DIFF
--- a/tests/5.1/motion/test_motion_present.c
+++ b/tests/5.1/motion/test_motion_present.c
@@ -17,35 +17,37 @@
 
 #define N 1024
 
-int errors, i;
+int errors;
+int i;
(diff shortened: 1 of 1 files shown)


tmh97 (approved):
good to go

PR #405, opened 2021-10-25
Fix metadirectives testcases

tob2:
* tests/5.0/metadirective/test_metadirective_arch_is_nvidia.c:
  - Move omp_get_device_num() call inside loop.
  - Use 'for' instead of 'loop' as loop implies order(concurrent)
    which does not permit routine calls.
  - Add missing space between two words in an error message.
* tests/5.0/metadirective/test_metadirective_arch_nvidia_or_amd.c:
  - Add { } to apply metadirective also to the following loop.
  - Run device loop at least once to init a[i] also when only the
    host device is available.

As reported by a colleague, the testcases have several issues; I hope I got the fixes right.

@nolanbaker31 @spophale @tmh97  – please review.

(Rebased + force pushed: changed 'teams for' to 'teams distribute parallel for' as the nvptx as 'teams for' is not permitted.)

### DIFF
--- a/tests/5.0/metadirective/test_metadirective_arch_is_nvidia.c
+++ b/tests/5.0/metadirective/test_metadirective_arch_is_nvidia.c
@@ -32,19 +32,19 @@ int metadirective1() {
    #pragma omp target map(to:v1,v2) map(from:v3, target_device_num) device(default_device)
    {
       #pragma omp metadirective \
-                   when(   device={arch("nvptx")}: teams loop) \
-                   default(                     parallel loop)
-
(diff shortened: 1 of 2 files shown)


tob2:
I committed a follow-up fix (move omp_is_initial_device() call to the for loop) to avoid violating:

_"If a teams region is nested inside a target region, the corresponding target
construct must not contain any statements, declarations or directives outside
of the corresponding teams construct."_

nolanbaker31 (approved):
lgtm

PR #411, opened 2021-11-04
Adding -qopenmp flag for Intel ICX, ICPX, IFX compilers in Make.def

nolanbaker31:
According to [Intel]( https://www.intel.com/content/www/us/en/develop/documentation/get-started-with-cpp-fortran-compiler-openmp/top.html), 

> Compile the source code with an icx, icpx, or ifx compiler driver that invokes GPU offloading with:
> $ icx -qopenmp -fopenmp-targets=spir64 matmul_offload.c -o matmul

Adding -qopenmp as it seems this properly enables GPU offloading. I have tested this with all 3 compilers, and am getting limited errors.

### DIFF
--- a/sys/make/make.def
+++ b/sys/make/make.def
@@ -105,7 +105,7 @@ endif
 
 # Intel ICX compiler
 ifeq ($(CC), icx)
-  COFFLOADING = -fopenmp-targets=spir64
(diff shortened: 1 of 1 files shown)


tmh97 (approved):
good-to-go

spophale (approved):

PR #412, opened 2021-11-07
Adding new .def entry for nvidia compilers (nvc,nvc++,nvfotran)

tmh97:
Tested with c, c++, and F90 programs on summit.


nolanbaker31 (approved):

spophale:
Doesn't this file get overridden by the system specific file ? Do we need to make this change for the specific platform files as well ?

spophale (approved):

PR #413, opened 2021-11-07
Adding -fopenmp-version flags for clang/clang++ to make.def file

tmh97:
When OMP_VERSION is added as part of the make command and is set to either 5.0 or 5.1, the corresponding -fopenmp-version={5.0 or 5.1} flag will be set.

If OMP_VERSION is not specified or is set equal to 4.5, the flag will not be set. When the flag is not set, it appears that -fopenmp-version defaults to 5.0.

I tested -fopenmp-version=4.5 and it forced all of the 4.5 offloading tests to fail, seems to not work properly for 4.5 only.

### DIFF
--- a/sys/make/make.def
+++ b/sys/make/make.def
@@ -116,7 +116,13 @@ endif
 # Clang compiler
 ifeq ($(DEVICE_TYPE),nvidia)
   ifeq ($(CC), clang)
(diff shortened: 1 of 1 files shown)


nolanbaker31 (approved):

spophale (approved):

tmh97:
Tested 
1. 5.0 performs the same with and without the -fopenmp_version=50 flag
2. 4.5 performs the same with and without the -fopenmp_version=50 flag

PR #415, opened 2021-11-11
Fix memleak in tests/5.0/task/test_task_affinity*.c

tob2:
Silence address sanitizer memleak warning by freeing two
variables in
* tests/5.0/task/test_task_affinity.c
* tests/5.0/task/test_task_affinity_device.c

Found when re-testing those tests from pull request #354

PR #416, opened 2021-11-12
Plug memory leaks, fix cast errors

tob2:
Plug memory leaks reported by `-fsanitize=address` (ASAN).
Fix errors of the type
  `error: invalid conversion from ‘void*’ to ‘int*’`
with `malloc`.

Follow up to Pull Request #415, which plugged one memory leak,
tested with GCC 12.


nolanbaker31 (approved):
lgtm

tmh97 (approved):

PR #418, opened 2021-11-12
Fix enum conversion and int overflow

tob2:
* tests/4.5/target_teams_distribute/test_target_teams_distribute_defaultmap.c:
  Add casts or use enum value directly to avoid the compile-time
  error: invalid conversion from ‘int’ to ‘enum_type’
* tests/4.5/target_teams_distribute/test_target_teams_distribute_nowait.c: Zero
  initialize used array elements and use int64_t to avoid with UBSAN the error:
  runtime error: signed integer overflow: 2147457775 + 731520 cannot be represented in type 'int'

UBSAN = undefined-behavior sanitizer, `-fsanitize=undefined`; here, the result with GCC 12.

### DIFF
--- a/tests/4.5/application_kernels/declare_target_subroutine.F90
+++ b/tests/4.5/application_kernels/declare_target_subroutine.F90
@@ -52,6 +52,7 @@ PROGRAM declare_target_subroutine
 
   OMPVV_TEST_VERBOSE(sum .ne. N)
 
+  DEALLOCATE(a)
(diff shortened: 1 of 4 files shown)


tob2:
Add another commit to fix the same integer overflow, but this time for Fortran – and a memory leak (cf. also PR #416 )

tmh97 (approved):

nolanbaker31 (approved):
looks good

Issue #417, opened 2021-11-12
error: OpenMP runtime API call ‘int omp_get_thread_num()’ strictly nested in a ‘teams’ region


tob2:
### ISSUE: OpenMP only permits two API routines in strictly nested in a `teams` – but other routines are invoked.

Solution: Either wrap those calls in a permitted region like `parallel` or `distribute` or `loop`, remove them outside `teams` (but note that for a `teams` nested in `target` nothing else is permitted in the target region except for the `teams` region), or remove the API call completely.

The error messages below are obtained with GCC 12, which added the check for the API routine call.

OpenMP has (here 5.2 wording) has in "10.2 `teams` Construct".

> Restrictions to the `teams` construct are as follows:
>
> * [...]
> * `distribute` regions, including any distribute regions arising from composite constructs, `parallel` regions, including any `parallel` regions arising from combined constructs, `loop` regions, `omp_get_num_teams()` regions, and `omp_get_team_num()` regions are the only OpenMP regions that may be strictly nested inside the `teams` region.

### Errors reported by GCC

* https://github.com/SOLLVE/sollve_vv/blob/b87d5ef85192c56bbcd8fdeaab87eb1bbe2c1bb1/tests/5.0/loop/test_loop_bind_device.c#L45
```
tests/5.0/loop/test_loop_bind_device.c: In function ‘int test_loop_bind_teams()’:
tests/5.0/loop/test_loop_bind_device.c:45:27: error: OpenMP runtime API call ‘int omp_get_thread_num()’ strictly nested in a ‘teams’ region
   45 |     if (omp_get_thread_num() == 0 && omp_get_team_num() == 0) {
      |         ~~~~~~~~~~~~~~~~~~^~
tests/5.0/loop/test_loop_bind_device.c: In function ‘int test_loop_bind_thread_teams()’:
tests/5.0/loop/test_loop_bind_device.c:138:27: error: OpenMP runtime API call ‘int omp_get_thread_num()’ strictly nested in a ‘teams’ region
  138 |     if (omp_get_thread_num() == 0 && omp_get_team_num() == 0) {
      |         ~~~~~~~~~~~~~~~~~~^~
```

* https://github.com/SOLLVE/sollve_vv/blob/b87d5ef85192c56bbcd8fdeaab87eb1bbe2c1bb1/tests/5.0/loop/test_loop_nested_device.c#L45
[...]

PR #419, opened 2021-11-13
Issue 417: Only call permitted routines strictly nested in 'teams'

tob2:
Issue #417

OpenMP only permits omp_get_num_teams() and omp_get_team_num() strictly
nested inside of a teams construct, all other have to be wrapped in
distribute, parallel or loop. Hence, this patch wraps them.

I have used `parallel if(0)` and `parallel`, but also `distribute` (possibly with `dist_schedule(static,1)` clause) or `loop` could be used.


spophale (approved):

tmh97 (approved):

nolanbaker31 (approved):
looks good

PR #425, opened 2021-11-18
Resolving race in test_taskwait_depend.c 

### LINKED ISSUE #424: Potential data race with 5.0/taskwait/test_taskwait_depend.c
The test was added in #366 ("taskwait depend") by @nolanbaker31 

https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.0/taskwait/test_taskwait_depend.c

The code does the following — my impression is that the dependency does not really work as multiple threads write `i` - such that `i` can have a different value in the following. That is: The `x` vs. `y` is handled properly but `i` can do more that what is intended.

_(It does fail with GCC, which is a hint but does not proof that the testcase is wrong as it could also be a compiler bug. I still need to do some spec reading to really nail it.)_

```C
    #pragma omp parallel for
        for (int i=1; i<N; i++){
          #pragma omp task depend(inout: x) shared(x) // 1st Task
          x=i;
          #pragma omp task depend(inout: y) shared(y) // 2nd Task
          y=i;
          #pragma omp taskwait depend(in: x) //Requires the completion of the 1st task
          OMPVV_TEST_AND_SET(errors, x!= i);
          #pragma omp taskwait depend(in: x,y) //Requires the completion of both tasks
          OMPVV_TEST_AND_SET(errors, y!=i && x!=i);
        }
```

Additionally, I think the `&&` should be `||` in the second `OMPVV_TEST_AND_SET`.

spophale:
See #424 

Closes #424

### DIFF
--- a/tests/5.0/taskwait/test_taskwait_depend.c
+++ b/tests/5.0/taskwait/test_taskwait_depend.c
@@ -15,18 +15,20 @@
 
 #define N 1024
 int errors = 0;
-int x,y = 0;
 int test_wrapper() { //wrapper for taskwait depend function
     #pragma omp parallel for
         for (int i=1; i<N; i++){
+          int x,y,err = 0;
           #pragma omp task depend(inout: x) shared(x) // 1st Task
           x=i;
           #pragma omp task depend(inout: y) shared(y) // 2nd Task
           y=i;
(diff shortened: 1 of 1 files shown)


tob2 (approved):
LGTM - thanks!

nolanbaker31 (approved):

tmh97 (approved):

Issue #426, opened 2021-11-22
Partially present issue


simoatze:
Hi guys,

I have some doubts about the following two tests and I would like your opinion/help.

https://github.com/SOLLVE/sollve_vv/blob/19db67029128bfc146998745f7db4e21bbd9cf71/tests/4.5/target_enter_data/test_target_enter_data_depend.c
https://github.com/SOLLVE/sollve_vv/blob/19db67029128bfc146998745f7db4e21bbd9cf71/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_depend.c

Let's take for example the first one: [test_target_enter_data_depend.c](https://github.com/SOLLVE/sollve_vv/blob/19db67029128bfc146998745f7db4e21bbd9cf71/tests/4.5/target_enter_data/test_target_enter_data_depend.c)

This tests has two functions called in the main, and they are unrelated to each other. The first function that is called does something like this:

```#pragma omp target enter data map(alloc: h_array[0:N]) depend(out: h_array) 

#pragma omp target enter data map(to: val) depend(out: val) 

#pragma omp target map(tofrom: isHost) map(alloc: h_array[0:N]) depend(inout: h_array) depend(in: val) 
  {
    isHost = omp_is_initial_device();
    for (int i = 0; i < N; ++i) {
      h_array[i] = val; // val = DEVICE_TASK1_BIT
    }
  }```

Where `val` is a local variable (i.e. instantiated in the stack). `val` is added to the present table and its present counter is 1. It is used in the first target region. However, the `map(to: val)` is not really used because the target region doesn't `map(to: val)` again so val it is actually `firstprivate` in the target. What happens is that at the end of the target region val is not deallocated (e.g. there is no `map(delete: val) or whatever). So when that function returns val has still present count equal 1. Now the second function is called and the address of `h_array` (which is in the stack) is very close (or the same) to the variable `val`. The runtime looks up for `&h_array` in the present table but it finds val and it thinks that `h_array` is only partially present giving a runtime error. The issue is that the runtime doesn't know that `val` is not valid anymore. I am not sure if this is an user error, for example the user should make sure there are not left over local variable with present counter > 1. So in this example there shouldn't be `map(to: val)` at all or make sure we have a `map(delete: val)` in the target region.

Please, let me know if it makes sense and what are your thoughts about this.

Thanks!
Simone

simoatze:
Also, I thought there was something in the spec that says that local variable should be deleted at the end of a target region but I can't find it.

PR #427, opened 2021-11-23
Issue 426: Free device vars before stack-local var lifetime ends

tob2:
Issue #426 

Free memory belonging to stack-local variables on the device before
their lifetime ends on the host.

* tests/4.5/target_enter_data/test_target_enter_data_depend.c
  (test_async_between_task_target, test_async_between_target): Add
  stack local vars to 'omp target exit data'.
* tests/4.5/target_enter_exit_data/test_target_enter_exit_data_depend.c
  (test_async_between_task_target, test_async_between_target): Likewise.

I concur with @simoatze  that the memory should be freed – but I also could not spot it in the spec. (Hence, I asked on the omp-lang mailing list.) Still, I think it should be freed.

@simoatze – does the patch look right?

### DIFF
--- a/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_depend.c
+++ b/tests/4.5/target_enter_exit_data/test_target_enter_exit_data_depend.c
@@ -81,6 +81,7 @@ int test_async_between_task_target() {
 
   errors = 2.0*N != sum;
 
+#pragma omp target exit data map(release: h_array[0:N], in_1[0:N], in_2[0:N])
   free(h_array);
   free(in_1);
   free(in_2);
@@ -124,6 +125,7 @@ int test_async_between_target() {
   }
   errors = 2*N != sum;
 
+#pragma omp target exit data map(release: h_array[0:N], val)
   free(h_array);
   return errors;
 }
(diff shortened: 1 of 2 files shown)


simoatze:
Thanks, Tobias. Fix looks good to me. Another approach would be to not move those data at all and let them be firstprivate in the target region since they are only read. But either way works for me. If the test is trying to also stress data movement this makes sense.

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_depend.c
```diff
@@ -124,6 +125,7 @@ int test_async_between_target() {
   }
   errors = 2*N != sum;
 
+#pragma omp target exit data map(release: h_array[0:N], val)
```
spophale, line 128: Nothing wrong in the suggestions, just curious why "release" and not "delete" here ?

spophale (approved):

tests/4.5/target_enter_exit_data/test_target_enter_exit_data_depend.c
tob2: Either should work - one forces it, the other uses the ref count (which should go to 0). The idea was to have some variance in the hope that it might find some additional bugs (unlikely). — I can change it, if you have any preference.

spophale: Not at all, I like variance ! I have already approved the PR, will have the changes pushed in by Thursday. Thanks again.

nolanbaker31 (approved):

PR #428, opened 2021-11-24
test_loop_bind.c - invalid API in teams; improve ...device_type_nohost.c

tob2:
* tests/5.0/loop/test_loop_bind.c: Add parallel if(0) wrapper to avoid invalid omp_get_thread_num call strictly nested in teams. -> Issue #417.
* tests/5.0/declare_target/test_declare_target_device_type_nohost.c: Fix check when no offload device is available.

_The first issue I was missed when doing the updates for #417; the second one failed rather non-transparently if no no offloading device is available. I hope this default_device_check make sense. Alternatively, a var could be set to omp_is_initial_device() inside a separate target region; the >= 0 check is to avoid issues with omp_invalid_device/omp_initial_device which are both < 0._

@spophale @tmh97 @nolanbaker31 - please review.

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
+++ b/tests/5.0/declare_target/test_declare_target_device_type_nohost.c
@@ -58,9 +58,18 @@ int test_declare_target_device_type_nohost() {
 
   #pragma omp target update from (a,b,c)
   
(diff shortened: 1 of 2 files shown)


spophale (approved):
LGTM

nolanbaker31 (approved):

PR #429, opened 2021-12-15
test_target_uses_allocators_pteam.c: Remove invalid { ... } before 'for'

tob2:
OpenMP requires that the loop/_loop-nest_ follows immediately after the `distribute` directive without additional `{ ... }` grouping.

@spophale @nolanbaker31 @tmh97 – please review

_Side note: Found because GCC chokes on it. It still does not compile in GCC as the `uses_allocators` clause is not supported. (Strictly speaking, the clause is not needed as the allocator is used on the target construct (in the `allocate` clause) but not _in_ the target region. For code written for portability, I think it should be removed. But as this a testsuite, having valid but unnecessary clauses is fine.)_

 * * *

OpenMP 5.1 has in "2.11.6 distribute Loop Constructs":
> #pragma omp distribute [clause[ [,] clause] ... ] new-line
>    loop-nest

Likewise in OpenMP 5.2, just a bit more hidden: "11.6 distribute Construct" has "Association: loop".

And _loop-nest_  is "2.11.1 Canonical Loop Nest Form" (OpenMP 5.2: 4.4.1 Canonical Loop Nest Form):
> for (init-expr; test-expr; incr-expr)
>   loop-body

### DIFF
--- a/tests/5.0/target/test_target_uses_allocators_pteam.c
+++ b/tests/5.0/target/test_target_uses_allocators_pteam.c
@@ -27,12 +27,11 @@ int test_uses_allocators_pteam() {
   }
 
 #pragma omp target teams distribute uses_allocators(omp_pteam_mem_alloc) allocate(omp_pteam_mem_alloc: x) private(x) map(from: device_result)
-{
   for (int i = 0; i < N; i++) {
     x = 2 * i;
     device_result[i] = i + x;
   }
-}
+
   for (int i = 0; i < N; i++) {
     OMPVV_TEST_AND_SET_VERBOSE(errors, result[i] != device_result[i]);
   }


tob2:
> @spophale @nolanbaker31 @tmh97 – please review

PING

tmh97:
@tob2 hey tobias apologies for delay and thank for your patience! 

Just out of curiosity (bit of a nit-pick), is there anywhere in the spec that explicitly states that braces cannot be placed around a loop-nest in this manner?

The latest llvm trunk is able to compile and run this without any issues, however nvidia's hpc compiler (nvc) throws an error similar to gcc

tmh97 (approved):

tob2:
> Just out of curiosity (bit of a nit-pick), is there anywhere in the spec that explicitly states that braces cannot be placed around a loop-nest in this manner?

Well, as quoted: "loop-nest" is "for ..." and not just a structured block which contains (among other or only) a "for (...)".

I find the OpenMP 5.1 syntax pretty obvious, if I look at the syntax for `omp distribute` replace `loop-nest` by the listed syntax. (OpenMP 5.2 also has it but it is much more hidden.)

I think the wording is there to make parsing simpler – and avoid adding a lot of additional words which just permit adding `{ }` around a loop-nest and nothing before or after the loop-nest inside.

PR #432, opened 2022-01-06
Adding new 5.1 test for defaultmap with present clause for scalars

tmh97:
Currently fails with latest llvm, I believe due to compiler error. If the **defaultmap(present: scalar)** is switched with **map(present, alloc: scalar_var, float_var, double_var)** the code compiles and runs fine.

### DIFF
--- a/tests/5.1/target/test_target_defaultmap_present_scalar.c
+++ b/tests/5.1/target/test_target_defaultmap_present_scalar.c
@@ -32,7 +32,7 @@ int test_defaultmap_present_scalar() {
    double_var = 12.22;
 
    #pragma omp target enter data map(to: scalar_var, float_var, double_var)
-   #pragma omp target map(tofrom: errors) map(present, alloc: scalar_var, float_var, double_var)
+   #pragma omp target map(tofrom: errors) defaultmap(present: scalar)
    {
       if(scalar_var != 1){errors++;}
       if(float_var != 10.7f){errors++;}
--- a/tests/5.1/target/test_target_defaultmap_present_scalar.c
+++ b/tests/5.1/target/test_target_defaultmap_present_scalar.c
@@ -1,0 +1,58 @@
+//===--- test_target_defaultmap_present_scalar.c ----------------------------===//
+//
+//  OpenMP API Version 5.1 Aug 2021
+//
+//  This test checks behavior of the defaultmap clause when the specified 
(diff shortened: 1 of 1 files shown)


spophale (approved):

nolanbaker31 (approved):

tob2:
> Currently fails with latest llvm, I believe due to compiler error. If the **defaultmap(present: scalar)** is switched with **map(present, alloc: scalar_var, float_var, double_var)** the code compiles and runs fine.

The problem is that map modifiers require that the map-type modifier comes always last and is prsent. While it is fine that "close" or "always, none of the map types makes sense for "present".

For OpenMP 5.1, that's in "2.21.7.1 map Clause", https://www.openmp.org/spec-html/5.1/openmpsu119.html#x154-1840002.21.7.1

`map([[map-type-modifier[,] [map-type-modifier[,] ...]] map-type: ] locator-list)`

Note: the map-type is required if a map-type-modifier (such as "present") is specified.

In OpenMP 5.2, the "ultimate" property of the "map-type" has the same effect.

Thus, I suggest to fix the testcase by adding a map modifier – I think any will do, but "alloc" seems to make most sense.

Crossref: OpenMP spec Issues 3193 (if you have access) discusses the same issue. Thus, it will likely be fixed in OpenMP 6.0 and, assuming there will be one, the next Technical Report (TR11?) that is likely to be released before OpenMP 6.0.

tob2:
IGNORE my mutter above - while it is correct, that remark was about `map` and not `defaultmap`.

This testcase uses `defaultmap` and there `present` is explicitly permitted, cf. https://www.openmp.org/spec-html/5.1/openmpsu119.html#x154-1860002.21.7.3

tob2:
`defaultmap(present)` is valid IMHO. Thus, LLVM has a bug if it rejects it as you described.

From the behavior (first comment in this PR, esp. `present, alloc` part), it looks as if LLVM re-uses the parse function for `map` for `defaultmap`.

For GCC: `present` has not yet been implemented – hence, it is currently always rejected (cf. also https://gcc.gnu.org/onlinedocs/libgomp/OpenMP-Implementation-Status.html )

> (potentially the same as in issue 3193, i am looking to get access to that repo now)

Contributions to the spec (bug reports, wording suggestions, etc.) are welcome :-)

PR #440, opened 2022-01-24
Allocate directive new tests

spophale:
New 5.0 and 5.1 allocate tests. 
5.0 host test tested with LLVM. Device version not supported by GCC or LLVM.
5.1 feature not supported by LLVM trunk.

### DIFF
--- a/tests/5.0/allocate/test_allocate.c
+++ b/tests/5.0/allocate/test_allocate.c
@@ -1,0 +1,47 @@
+//===----------------- test_allocate.c ------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// Tests the allocate directive. A variable 'x' is allocated using 
+// the omp_alloc call. If no clause is specified then the memory allocator s
+// pecified by the def-allocator-var ICV will be used. The tests 
+// checks that the values were written correctly, and then frees the memory.
+//
+//===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+#define N 1024
+
+int test_allocate() {
+  int errors = 0;
+  int* x;
+
+#pragma omp allocate(x) 
+  x = (int *) omp_alloc(N*sizeof(int), omp_null_allocator);
+
+#pragma omp parallel for 
+  for (int i = 0; i < N; i++) {
+    x[i] = i;
+  }
+
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, x[i] != i);
+  }
+
+  omp_free(x,omp_default_mem_alloc);
+
+  return errors;
+}
+
+int main() {
+
+  int errors = 0;
+
+  OMPVV_TEST_AND_SET_VERBOSE(errors, test_allocate() != 0);
+
+  OMPVV_REPORT_AND_RETURN(errors);
+}
--- a/tests/5.0/allocate/test_allocate.c
+++ b/tests/5.0/allocate/test_allocate.c
@@ -21,7 +21,6 @@ int test_allocate() {
   int* x;
 
 #pragma omp allocate(x) 
-  x = (int *) omp_alloc(N*sizeof(int), omp_null_allocator);
 
 #pragma omp parallel for 
   for (int i = 0; i < N; i++) {
@@ -32,7 +31,6 @@ int test_allocate() {
     OMPVV_TEST_AND_SET_VERBOSE(errors, x[i] != i);
   }
 
-  omp_free(x,omp_default_mem_alloc);
 
   return errors;
 }
--- a/tests/5.0/allocate/test_allocate.c
+++ b/tests/5.0/allocate/test_allocate.c
@@ -18,7 +18,7 @@
 
 int test_allocate() {
   int errors = 0;
-  int* x;
+  int x[N];
 
 #pragma omp allocate(x) 
 
--- a/tests/5.0/allocate/test_allocate_on_device.c
+++ b/tests/5.0/allocate/test_allocate_on_device.c
@@ -7,7 +7,9 @@
 // According to the spec omp_alloc invocations that appear in target regions 
 // must not pass omp_null_allocator as the allocator argument, which must be 
 // a constant expression that evaluates to one of the predefined memory allocator 
-// values. The test checks that the values were written correctly, and then frees the memory.
+// values. Predefined allocators appearing in a uses_allocators clause cannot 
+// have traits specified. The test checks that the values were written correctly, 
+// and then frees the memory.
 //
 //===----------------------------------------------------------------------===//
 
@@ -22,7 +24,7 @@ int test_allocate_on_device() {
 
   int A[N], errors = 0;
 
-#pragma omp target map(tofrom: errors, A)
+#pragma omp target map(tofrom: errors, A) uses_allocators(omp_default_mem_alloc)
   {
    int* x;
    #pragma omp allocate(x) allocator(omp_default_mem_alloc)
--- a/tests/5.0/allocate/test_allocate_on_device.c
+++ b/tests/5.0/allocate/test_allocate_on_device.c
@@ -1,0 +1,55 @@
+//===----------------- test_allocate_on_device.c ------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// Tests the allocate directive with allocator clause in the target region.
+// A variable 'x' is allocated using and omp_default_mem_alloc allocator. 
+// According to the spec omp_alloc invocations that appear in target regions 
+// must not pass omp_null_allocator as the allocator argument, which must be 
+// a constant expression that evaluates to one of the predefined memory allocator 
+// values. The test checks that the values were written correctly, and then frees the memory.
+//
+//===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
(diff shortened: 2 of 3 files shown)


tests/5.0/allocate/test_allocate_on_device.c
```diff
@@ -0,0 +1,57 @@
...
+#define N 1024
+
+int test_allocate_on_device() {
+
+  int A[N], errors = 0;
+
+#pragma omp target map(tofrom: errors, A) uses_allocators(omp_default_mem_alloc)
+  {
+   int* x;
+   #pragma omp allocate(x) allocator(omp_default_mem_alloc)
```
tmh97, line 30: Should thing allocator be defined outside of the target region and then presented to the uses_allocators clause as an _allocator_

tests/5.1/allocate/test_allocate_allocator_align.c
```diff
@@ -0,0 +1,54 @@
+//===------ test_allocate_allocator_align.c ------------------------------------===//
+//
+// OpenMP API Version 5.1 Nov 2020
+//
+// Tests the allocate directive with allocator and align clause.
+// The declarative allocator statement uses the omp_default_mem_alloc handle
+// for default memory allocation for "x", aligned to 64-byte alignment via the 
+// align clause. Parallel region checks that 64-byte alignment is correct 
+// and that the memory can be written to in and the values
```
tmh97, line 9: This test passes on latest clang, present message about loop not vectorized

tests/5.0/allocate/test_allocate_on_device.c
```diff
@@ -0,0 +1,57 @@
+//===----------------- test_allocate_on_device.c ------------------------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// Tests the allocate directive with allocator clause in the target region.
+// A variable 'x' is allocated using and omp_default_mem_alloc allocator.
```
tmh97, line 6: This test fails with latest llvm trunk on summit due to 
`nvlink error   : Undefined reference to 'omp_alloc' in '/tmp/test_allocate_on_device-0fac84.cubin'
nvlink error   : Undefined reference to 'omp_free' in '/tmp/test_allocate_on_device-0fac84.cubin'`

tests/5.0/allocate/test_allocate.c
```diff
@@ -0,0 +1,47 @@
...
+#include "ompvv.h"
+
+#define N 1024
+
+int test_allocate() {
+  int errors = 0;
+  int* x;
+
+#pragma omp allocate(x) 
+  x = (int *) omp_alloc(N*sizeof(int), omp_null_allocator);
```
tob2, line 24: I think I don't understand this example. I do understand:
* `int *x; x = (int*) omp_alloc( .... );` — This allocates explicitly with a routine call, which has to be freed by the user (→ omp_free).
* `int A[10], *y; #pragma omp allocate(A, y)` – This allocates the memory - and it is automatically freed at the end of the scope of the variable [directive and variable are in the same scope]. (Though, I am not 100% sure about the 'y' semantic; I think it allocates a scalar pointer - but I might be wrong. All examples I saw for the 'allocate' directive use constant arrays [but ↓].)

In any case, I fail to understand why there is the `allocate` directive _and_ an `omp_alloc` for the same variable.

I think the only exception where the `omp allocate` does not stand on its own is for Fortran with allocatables/pointers, for which the directive line is followed by a Fortran ALLOCATE statement. But to avoid semantic overloading – and also to avoid issues in the compiler about whether the directive belongs to the declaration or execution part, the later was changed by introducing the new `allocators` construct in OpenMP 5.2 (and deprecating the `allocate` directive for this purpose, only).

_Disclaimer: I might have misunderstood the OpenMP specification._

BTW: GCC: _'omp allocate' support exists as patch but it is not yet in GCC 12.  However, calling `omp_alloc` in GCC in target regions is fine. (I think since GCC 11, but I have not checked.)_

tob2: Side note part two: GCC – additional minor comments regarding GCC's implementation status:
* `uses_allocators` on `target` is not supported, but `requires dynamic_allocators` is fulfilled (also in GCC 11; but since GCC 12 it is accepted by turning into a no op). – Thus, the omp_alloc etc. should work since GCC 11 (also in `target`). But GCC will choke on the `uses_allocators`
* GCC _does_ support the `allocate` clause as in `omp task private(a) allocate(a)` for C/C++ since GCC 11. And since GCC 12, it should support it also for Fortran and also OpenMP 5.1's `allocate (allocator (h), align (8): n2)` in C/C++.
  _I don't know whether you already have a testcase for the `allocate` clause or not._

spophale: @tob2  This is a 5.0 test. Allocate directive just declares the access traits required, while omp_alloc is where the allocation happens. I agree, allocate followed by omp_alloc for the same mem space might be an overkill (though completely legal). For the on device test, I am going to do the variation with the stack allocate (and that passes with LLVM, not checked with GCC yet). 
For my information, in the GCC implementation, if a regular malloc is used after the allocate directive, does it get intercepted to by omp runtime to verify the access traits of the memory and allocate memory accordingly?

seyonglee: @spophale In my understanding, Allocate directive in C/C++ is a stand-alone directive, not associated with a statement. Therefore, actual memory allocation happens in the Allocate directive, which means that this example has duplicated allocations for the same variable, which is an overkill but still legal.

spophale: Aah, I see. From the language- "the storage for each list item that appears in the directive will be provided by an allocation through a memory allocator" it was not very clear to me that the allocation happens via the allocate directive or we need to use an explicit **memory allocator**. It makes sense that allocate provides the memory.

spophale: Thanks @tob2 and @seyonglee !

tests/5.0/allocate/test_allocate.c
```diff
@@ -21,7 +21,6 @@ int test_allocate() {
   int* x;
 
 #pragma omp allocate(x) 
-  x = (int *) omp_alloc(N*sizeof(int), omp_null_allocator);
```
tob2: If you want to access x[0] to x[N], I think you either need `int x[N];` (C array) with `#pragma omp allocate(x)` — or you have to be content that only the pointer gets allocated (`*x`), which does not have room for `N` elements — or you still use `omp_alloc`/`omp_delete` with a pointer.

tests/5.0/allocate/test_allocate.c
tob2: > Allocate directive just declares the access traits required, while ... is where the allocation happens.

That's actually the case — but only for ALLOCATABLE/POINTER variables in Fortran:
```F90
!$omp allocate allocator(my_alloc)  ! this usage is deprecated - and replaced by the 'allocators' construct in OpenMP 5.1
allocate (x(N))   ! uses the Fortran allocate statement.
```

tests/5.0/allocate/test_allocate.c
seyonglee: @tob2 I agree too; in my previous comment, the overkill means that the pointer variable is allocated twice: one in "int *x" and the other in "#pragma omp allocate(x)", but no memory allocation occurs for the data to be pointed by the pointer variable. Another way to allocate data to be pointed by the pointer variable is to use an array section as an argument to the allocate directive ("#pragma amp allocate(x[0:N])").

tob2: @seyonglee :
> Another way to allocate data to be pointed by the pointer variable is to use an array section as an argument to the allocate directive ("#pragma amp allocate(x[0:N])").

I am not quite sure whether that's permitted or not. The current wording is (here for OpenMP 5.2):

"6.5 allocate Directive" - "Arguments" - "list = list of variable list item type"

With "A variable list item is a variable or an array section." – But for something that does has not been allocated yet, "array section" does not make sense (especially not: "The array section must be a subset of the original array." to quote from "3.2.5 Array Sections") .

_I have now filled an OpenMP Spec Issue for this (Issue 3232, not publicly visible)_

tob2: Bronis pointed out (in that Issue 3232) that "Array Sections" [OpenMP 5.2, p. 66, line 9]) has: “An array section can appear only in clauses for which it is explicitly allowed.”

Thus, "#pragma amp allocate(x[0:N])") is invalid.

seyonglee: @tob2 thank you for this clarification. 
I think that I have misunderstood the purpose of the "allocate" directive; so the purpose of the "allocate" directive is to allocate the corresponding "device memory" for the existing host data available on the host memory? Please correct me if there is anything wrong.

spophale: Thanks @tob2 and @seyonglee for the discussion! This is exactly why we write these tests! :) 
So, in summary, the correct way to go about the allocate directive is:
1) int x[N];
    #pragma omp allocate (x) allocator (any_allocator)
OR for dynamic allocation:
2) int *x;
     #pragma omp allocate (x) allocator (any_allocator)
     x = (int *) omp_alloc(N*sizeof(int), same_any_allocator); //the spec does not mention "same" but that makes sense.
    
I will make sure all these cases are covered in the tests.

seyonglee: Thanks @tob2 @spophale; now I have better understanding on the "allocate" directive.

tmh97 (approved):
looks-good-to-go

nolanbaker31 (approved):
lgtm

PR #441, opened 2022-01-24
Target Version of 'Task In Reduction Dynamically Enclosed'

nolanbaker31:
Currently failing with errors on LLVM.

Let me know if you have any ideas

### DIFF
--- a/tests/5.0/task/test_task_in_reduction_dynamically_enclosed_device.c
+++ b/tests/5.0/task/test_task_in_reduction_dynamically_enclosed_device.c
@@ -1,0 +1,60 @@
+//===--- test_task_in_reduction_dynamically_enclosed_device.c ----------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+//
+// This test checks the task directive with the `in_reduction` reduction
+// participation clause. It performs simple array operations which are added
+// to a reduction variable in an explcit task with the in_reduction clause.
+// This checks the above in a case in which the in_reduction tasks are
+// nested in a function in the taskgroup. Thanks to Jakub Jelinek for
+// suggesting this test.
+// This test checks the above in a target context.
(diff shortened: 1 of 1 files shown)


tmh97:
I believe this test is good-to-go, I think it only fails due to lack implementation. 'worked on' status as 02/04/2022

tmh97 (approved):

tests/5.0/task/test_task_in_reduction_dynamically_enclosed_device.c
```diff
@@ -0,0 +1,62 @@
...
+}
+
+int test_task_in_reduction_dynamically_enclosed() {
+  OMPVV_INFOMSG("test_task_in_reduction_dynamically_enclosed");
+  int errors = 0;
+  int expect = 2;
+
+#pragma omp taskgroup task_reduction(+:sum)
+  task_container(0);
+#pragma omp target taskloop reduction(+:sum) map(tofrom:sum)
```
tob2, line 40: This does not seem to be valid. I cannot find 'target taskloop' as permitted in OpenMP 5.0, 5.1 or 5.2.

In OpenMP 5.2, _17.3 Combined and Composite Directive Names_, there is for **taskloop** only:

> If _directive-name-A_ is **taskloop**, **for** or **do** then _directive-name-B_ may be **simd.**
> If _directive-name-A_ is **masked** then _directive-name-B_ may be **taskloop** or the directive name of a combined or composite construct for which _directive-name-A_ is **taskloop**

tob2: And I cannot find anything which permits **target** ... **masked/master** and then **taskloop** — as the leaf construct has to be 'do'/'for' (and in one case 'loop').

PR #451, opened 2022-01-31
New teams test with thread limit

jrreap:
Tests the thread_limit clause on the teams construct. Also tests it's inheritance from a parent target construct.


tests/5.1/teams/test_target_teams_default_firstprivate.c
```diff
@@ -0,0 +1,42 @@
...
+#define N 1024
+
+int main() {
+	int errors = 0;
+	int not_shared = 5;
+	int num_teams = 0; 
+
+	#pragma omp target teams default(firstprivate) map(tofrom:num_teams) shared(num_teams) num_teams(OMPVV_NUM_TEAMS_DEVICE)
+	{
+		num_teams = omp_get_num_teams();
```
spophale, line 30: race.

PR #453, opened 2022-02-02
Add new_test/test_target_defaultmap_to_from_tofrom.F90.

seyonglee:
Add a new Fortran test benchmark: tests/5.0/target/test_target_defaultmap_to_from_tofrom.F90
Tested with gcc/11.1.0 and xl/16.1.1-10 on Summit, but botf failed because defaultmap clause is not supported yet.


spophale (approved):

tmh97 (approved):
This is good-to-go!

PR #455, opened 2022-02-02
Add new_test/test target defaultmap none.f90

seyonglee:
Add a new Fortran test benchmark: tests/5.0/target/test_target_defaultmap_none.F90
Tested with both xl/16.1.1-10 and gcc/11.1.0 on Summit, but both failed because defaultmap clause is not supported yet.

### DIFF
--- a/tests/5.0/target/test_target_defaultmap_none.F90
+++ b/tests/5.0/target/test_target_defaultmap_none.F90
@@ -1,0 +1,95 @@
+!===--- test_target_defaultmap_default.F90 -------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+!   This test checks behavior of the defaultmap clause when the specified implicit-behavior  
+!   is none. The variable-categories avaiable for defaultmap are scalar, aggregate, allocatable,
+!   and pointer. When defaultmap(none) is used alone without a variable-category, then none is 
+!   the implicit-behavior for all variables referenced in the construct. Since none is used, 
+!   the data mapping attributes must be specified somewhere else. In this case, they are 
+!   specified using the map clause.
+!
+!//===----------------------------------------------------------------------===//
+#include "ompvv.F90"
+
+#define N 1024
+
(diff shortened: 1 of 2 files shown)


tests/5.0/target/test_target_defaultmap_none.F90
```diff
@@ -0,0 +1,95 @@
...
+    DO i = 1, N
+       A(i) = 0
+       B(i) = 0
+       new_struct%SA(i) = 0
+    END DO
+
+    errors = 0
+
+    !$omp target defaultmap(none) map(tofrom: scalar, A, B, new_struct, ptr)
+    scalar = 17 !scalar firstprivate, value not returned
```
spophale, line 57: I think if to from is explicit then the scalar value will be copied back.

spophale (approved):

tests/5.0/target/test_target_defaultmap_none.F90
seyonglee: The comment seems to be wrong; what I wanted to say was that scalar variables without TARGET, ALLOCATABLE, or POINTER attribute have an implicit data-sharing attribute of first-private. (OpenMP V5.0 page 315 line 9 ~ 12). Is it correct interpretation?

tmh97: @seyonglee IMO that is the correct interpretation, I see what you were implying with your comment. @spophale This looks good-to-go. I will approve and leave it up for you to merge. Thanks again @seyonglee !

tmh97:
I removed the defaultmap(none) and ran with nvhpc and gcc, passes on nvfortran but fails on gcc/11.1.10 with out-of-memory issue

tests/5.0/target/test_target_defaultmap_none.F90
seyonglee: Changed the comments for the scalar variable to clarify the intention.

PR #456, opened 2022-02-02
Add new_test/test_target_defaultmap_default.F90

seyonglee:
Add a new Fortran test benchmark: tests/5.0/target/test_target_defaultmap_default.F90
Tested with both xl/16.1.1-10 and gcc/11.1.0 on Summit, but both failed because defaultmap clause is not supported yet.


spophale (approved):

tmh97:
> Add a new Fortran test benchmark: tests/5.0/target/test_target_defaultmap_default.F90 Tested with both xl/16.1.1-10 and gcc/11.1.0 on Summit, but both failed because defaultmap clause is not supported yet.

runs fine with nvfortran!

tmh97 (approved):

PR #458, opened 2022-02-05
New teams test with thread limit

jrreap:
New test to validate that the thread limit works from a parent target construct to a nested teams construct

### DIFF
--- a/tests/5.1/teams/test_target_teams_thread_limit.c
+++ b/tests/5.1/teams/test_target_teams_thread_limit.c
@@ -1,0 +1,43 @@
+//===---------------- test_target_teams_thread_limit.c -----------------------===//
+//
+// OpenMP API Version 5.1 Nov 2020
+//
+// This test uses the thread_limit clause on the teams construct. Specifically
+// testing if a thread_limit from a above target construct properly carries
+// down to the nested teams construct, as if it were directly on the construct
+// as defined in the spec. The test validates that only the specified 
+// threads are created by summing a shared variable across all threads 
+// (and teams). If the threads are correctedly limited this should produce the 
+// expected value. Additional warnings are sent if specific issues occur.
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+#define N 1024
+
+int main() {
+	int errors = 0;
+	int shared = 0;
+	int num_teams = 0;
+
+	#pragma omp target thread_limit(4)
+	{
+		#pragma omp teams num_teams(OMPVV_NUM_TEAMS_DEVICE)
+		{
+			num_teams = omp_get_num_teams();
+			for (int i = 0; i < omp_get_num_threads(); i++) {
+				shared += 1;
+			}
+		}
+	}
+
+	OMPVV_WARNING_IF(num_teams != 8, "The number of teams was unexpected, the test results are likely inconcuslive")
+	OMPVV_WARNING_IF(shared > 32, "The sum was higher than expected. This likely means thread_limit isn't capping the maximum threads created.");
+	OMPVV_TEST_AND_SET(errors, (shared != 32));
+
+	OMPVV_REPORT_AND_RETURN(errors);
+}
--- a/tests/5.1/teams/test_target_teams_thread_limit.c
+++ b/tests/5.1/teams/test_target_teams_thread_limit.c
@@ -25,11 +25,13 @@ int main() {
 	int num_teams = 0;
 	int num_threads = 0;
 
+	int testing_thread_limit = OMPVV_NUM_THREADS_DEVICE/OMPVV_NUM_TEAMS_DEVICE;
+
 	OMPVV_TEST_OFFLOADING;
 
 	#pragma omp target map(tofrom:num_teams,num_threads,shared) //thread_limit(4)
 	{
-		#pragma omp teams thread_limit(8) num_teams(OMPVV_NUM_TEAMS_DEVICE)
+		#pragma omp teams thread_limit(testing_thread_limit) num_teams(OMPVV_NUM_TEAMS_DEVICE)
 		{
 
 			#pragma omp parallel
@@ -61,9 +63,10 @@ int main() {
 	printf("Threads %d\n", num_threads);
 	printf("Shared %d\n", shared);
 	printf("Teams %d\n", num_teams);
+	printf("Val %d\n", num_teams * testing_thread_limit);
 	OMPVV_WARNING_IF(num_teams != OMPVV_NUM_TEAMS_DEVICE, "The number of teams was unexpected, the test results are likely inconcuslive")
-	OMPVV_WARNING_IF(shared > 32, "The sum was higher than expected. This likely means thread_limit isn't capping the maximum threads created.");
-	OMPVV_TEST_AND_SET(errors, (shared != (num_teams * 4)));
+	OMPVV_WARNING_IF(shared > (num_teams * testing_thread_limit), "The sum was higher than expected. This likely means thread_limit isn't capping the maximum threads created.");
+	OMPVV_TEST_AND_SET(errors, (shared != (num_teams * testing_thread_limit)));
 
 	OMPVV_REPORT_AND_RETURN(errors);
 }
--- a/tests/5.1/teams/test_target_teams_thread_limit.c
+++ b/tests/5.1/teams/test_target_teams_thread_limit.c
@@ -28,7 +28,10 @@ int main() {
 	{
 		#pragma omp teams num_teams(OMPVV_NUM_TEAMS_DEVICE)
 		{
-			num_teams = omp_get_num_teams();
+			if (omp_get_team_num() == 0) {
+				num_teams = omp_get_num_teams();
+			}
+			
 			for (int i = 0; i < omp_get_num_threads(); i++) {
 				shared += 1;
 			}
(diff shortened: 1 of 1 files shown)


tests/5.1/teams/test_target_teams_thread_limit.c
```diff
@@ -0,0 +1,67 @@
...
+			#pragma omp parallel
+			{
+				if (omp_get_team_num() == 0) {
+					num_teams = omp_get_num_teams();
+				}
+
+				for (int i = 0; i < omp_get_num_teams(); i++) {
+					if (omp_get_team_num() == i) {
+						#pragma omp atomic write
+						num_threads = omp_get_num_threads();
```
tmh97, line 46: On line 46 we can do: if(omp_get_num_threads > 4) { errors ++;}

jrreap:
`thread_limit` does not behave as expected. Failing to actually limit threads to anything higher than 1. Submitting a bug with the vendor.

tests/5.1/teams/test_target_teams_thread_limit.c
```diff
@@ -0,0 +1,66 @@
...
+	OMPVV_TEST_OFFLOADING;
+
+	#pragma omp target map(tofrom:num_teams,shared,errors) thread_limit(testing_thread_limit)
+	{
+		#pragma omp teams num_teams(OMPVV_NUM_TEAMS_DEVICE)
+		{
+
+			#pragma omp parallel num_threads(testing_thread_limit)
+			{
+				if (omp_get_team_num() == 0) {
```
spophale, line 38: add "&& omp_get_thread_num == 0 "

tests/5.1/teams/test_target_teams_thread_limit.c
```diff
@@ -0,0 +1,66 @@
...
+	int shared = 0;
+	int num_teams = 0;
+
+	int testing_thread_limit = OMPVV_NUM_THREADS_DEVICE/OMPVV_NUM_TEAMS_DEVICE;
+
+	OMPVV_TEST_OFFLOADING;
+
+	#pragma omp target map(tofrom:num_teams,shared,errors) thread_limit(testing_thread_limit)
+	{
+		#pragma omp teams num_teams(OMPVV_NUM_TEAMS_DEVICE)
```
tmh97, line 33: #pragma omp teams distribute parallel

jrreap, line 33: Collapse parallel into this with omp teams distribute

tests/5.1/teams/test_target_teams_thread_limit.c
```diff
@@ -0,0 +1,69 @@
...
+	int shared = 0;
+	int num_teams = 0;
+
+	int testing_thread_limit = OMPVV_NUM_THREADS_DEVICE/OMPVV_NUM_TEAMS_DEVICE;
+
+	OMPVV_TEST_OFFLOADING;
+
+	#pragma omp target map(tofrom:num_teams,shared,errors) thread_limit(testing_thread_limit)
+	{
+		#pragma omp teams distribute parallel num_teams(OMPVV_NUM_TEAMS_DEVICE)
```
tob2, line 33: `teams distribute parallel` does not seem to be a valid combined construct – as also GCC 12 complains.

Permitted are (cf. OpenMP 5.1):
* teams distribute parallel for/do
* teams distribute parallel for/do simd
And also:
* teams distribute
* teams distribute simd
* teams loop

In words of OpenMP 5.2:
* If _directive-name-A_ is **teams** then _directive-name-B_ may be **loop**, **distribute** or the directive name of a combined or composite construct for which _directive-name-A_ is distribute.
* If _directive-name-A_ is **distribute** then _directive-name-B_ may be **simd** or the directive name of a combined or composite construct for which _directive-name-A_ is **parallel** and **for** or **do** is a leaf construct.

The following works - but I am not sure that it does what you intend:
```suggestion
		#pragma omp teams num_teams(OMPVV_NUM_TEAMS_DEVICE)
		#pragma omp parallel
```

jrreap: Ultimately we are trying to test the thread limit construct so having it be two pragmas is totally ok for this test. Regardless will get this patched, thanks for the input!

tests/5.1/teams/test_target_teams_thread_limit.c
```diff
@@ -0,0 +1,65 @@
...
+		{
+
+			if (omp_get_team_num() == 0 && omp_get_thread_num() == 0) {
+				num_teams = omp_get_num_teams();
+			}
+
+			if (omp_get_thread_num() == 0) {
+				for (int i = 0; i < omp_get_num_teams(); i++) {
+					if (omp_get_team_num() == i) {
+						if (omp_get_num_threads() > testing_thread_limit) {
```
nolanbaker31, line 43: does 'testing_thread_limit' need to get mapped 'to' in order to be recognized?

jrreap: It technically does not! I can move that into the construct directly. Good catch!

tests/5.1/teams/test_target_teams_thread_limit.c
```diff
@@ -0,0 +1,65 @@
...
+							errors++;
```
jrreap, line 44: Change errors to array with the index being the team number (to prevent issues with it being shared)

tmh97 (approved):

spophale (approved):

tob2:
This fails to compile with GCC mainline with:
```
tests/5.1/teams/test_target_teams_thread_limit.c: In function ‘main’:
tests/5.1/teams/test_target_teams_thread_limit.c:42:56: error: OpenMP runtime API call ‘omp_get_thread_num’ strictly nested in a ‘teams’ region
   42 |                         if (omp_get_team_num() == 0 && omp_get_thread_num() == 0) {
      |                                                        ^~~~~~~~~~~~~~~~~~~~
tests/5.1/teams/test_target_teams_thread_limit.c:46:29: error: OpenMP runtime API call ‘omp_get_thread_num’ strictly nested in a ‘teams’ region
   46 |                         if (omp_get_thread_num() == 0) {
      |                             ^~~~~~~~~~~~~~~~~~~~
tests/5.1/teams/test_target_teams_thread_limit.c:48:34: error: OpenMP runtime API call ‘omp_get_num_threads’ strictly nested in a ‘teams’ region
   48 |                              if (omp_get_num_threads() > testing_thread_limit) {
      |                                  ^~~~~~~~~~~~~~~~~~~~~
```

tmh97:
@tob2 I'm taking a look through the 5.1 specification to see if we violated any constraints. Just a heads up, the test compiled, ran, and passed with GCC 11.2.0 and LLVM15-latest on summit

tob2:
Regarding the spec. OpenMP 5.1 has:

> "Restrictions Restrictions to the **teams** construct are as follows:"

The last bullet starts with:
> * **distribute,** **distribute simd,** distribute parallel worksharing-loop, distribute parallel worksharing-loop SIMD, **parallel** regions, including any **parallel** regions arising from combined constructs,

and then continues with the relevant bit:
> `omp_get_num_teams()` regions, and `omp_get_team_num()` regions are the only OpenMP regions that may be strictly nested inside the **teams** region.

https://www.openmp.org/spec-html/5.1/openmpse15.html

Regarding the testcase: commit c32244dddd38c4395efdfff7737e9a8b42f7ea92 removed some directives, which I think causes the issue above.

Regarding the error: I can reproduce that GCC 11 (contrary to GCC 12 or OG11 or mainline) does not show that error. – That's not surprising, given that the check was added to GCC in October 2021 (cf. https://gcc.gnu.org/PR102972 ). It actually was added as follow-up to sollve_vv Issue 409 :-)

tests/5.1/teams/test_target_teams_thread_limit.c
```diff
@@ -0,0 +1,65 @@
...
+	int errors[OMPVV_NUM_THREADS_DEVICE/OMPVV_NUM_TEAMS_DEVICE];
+	int num_teams = 0;
+        int sum_errors = 0;
+        int i;
+	int testing_thread_limit = OMPVV_NUM_THREADS_DEVICE/OMPVV_NUM_TEAMS_DEVICE;
+
+	OMPVV_TEST_OFFLOADING;
+      
+
+         for (i = 0; i<testing_thread_limit; i++){
```
tob2, line 32: I think this should be:
```suggestion
         for (i = 0; i<OMPVV_NUM_TEAMS_DEVICE; i++){
```
as later it is iterated over `num_teams`

tob2:
> @tob2 Thank you for clarifying! I am thinking of adding back #pragma omp parallel around lines 42-50 as relevant fix, thoughts?

Works for me – at least with the additional `error[i] = 0` initialization fix. (On my laptop, num_teams is 8 and testing_thread_limit is 1.)

tob2:
@tmh97 – do you intent to work on this patch?

Additionally: `ompvv/ompvv.h` has:
```
  #define OMPVV_NUM_THREADS_DEVICE 8
  #define OMPVV_NUM_TEAMS_DEVICE 8
  #define OMPVV_NUM_THREADS_HOST 8
```
and this code has the following — i.e. always `testing_thread_limit = 1`:
```C
 	int testing_thread_limit = OMPVV_NUM_THREADS_DEVICE/OMPVV_NUM_TEAMS_DEVICE;
...
	#pragma omp target map(tofrom:num_teams,errors) thread_limit(testing_thread_limit)
		#pragma omp teams num_teams(OMPVV_NUM_TEAMS_DEVICE) 
```
but it still assumes that there should be more than one thread:
```C
OMPVV_WARNING_IF(testing_thread_limit == 1, "Only one thread was allocated to each team, the test results are likely inconclusive");
```

tob2:
I now put the @tmh97's parallel suggestion and mine into a new Pull Request: #553.

PR #459, opened 2022-02-08
Target 'has_device_addr' 5.1 Test

nolanbaker31:
Currently passes, with the target 'has_device_addr' clause unrecognized. (currently not implemented in LLVM).

Let me know if test seems to properly test the clause, not sure if there is a better way to test that would only pass if the clause functions properly.

### DIFF
--- a/tests/5.1/target/test_target_has_device_addr.c
+++ b/tests/5.1/target/test_target_has_device_addr.c
@@ -28,16 +28,25 @@ int test_target_has_device_addr() {
   int *first_arr_device_addr;
   int *second_scalar_device_addr;
   int *second_arr_device_addr;
-  #pragma omp target map(from: first_scalar_device_addr, first_arr_device_addr)
+  #pragma omp target enter data map(to: x, arr)
+  #pragma omp target map(from: first_scalar_device_addr, first_arr_device_addr) map(to: x, arr)
   {
     first_scalar_device_addr = &x;
     first_arr_device_addr = &arr[0];
   }
-  #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr[N]) 
+  int *data_x, *data_arr;
+  #pragma omp target data use_device_addr(x,arr)
+  {
+  data_x = &x; data_arr = &arr[0];
+  #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr) 
   {
     second_scalar_device_addr = &x;
     second_arr_device_addr = &arr[0];
   }
+  }
+ #pragma omp target exit data map(release: x, arr)
+ OMPVV_TEST_AND_SET(errors, first_scalar_device_addr != data_x);
+ OMPVV_TEST_AND_SET(errors, first_arr_device_addr != data_arr);
   OMPVV_TEST_AND_SET(errors, first_scalar_device_addr != second_scalar_device_addr);
   OMPVV_TEST_AND_SET(errors, first_arr_device_addr != second_arr_device_addr)
   return errors;
--- a/tests/5.1/target/test_target_has_device_addr.c
+++ b/tests/5.1/target/test_target_has_device_addr.c
@@ -35,6 +35,7 @@ int test_target_has_device_addr() {
   }
   int *second_scalar_device_addr, *second_arr_device_addr;
   //check addresses are same on device region
+  #pragma omp target data use_device_addr(x, arr)
   #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr) 
   {
     second_scalar_device_addr = &x;
--- a/tests/5.1/target/test_target_has_device_addr.c
+++ b/tests/5.1/target/test_target_has_device_addr.c
@@ -3,9 +3,9 @@
 // OpenMP API Version 5.1 Nov 2020
 //
 // This test verifies the 'has_device_addr' feature added to the target construct.
-// We tested this by checking a scalar & array address in one target region, and
-// verifying that the address remains the same in a second target region which
-// uses the 'has_device_addr' feature.
+// We tested this by mapping a scalar & array to the device (using both target map
+// & omp_target_alloc) and ensuring that the address does not change when using
+// has_device_addr on another target region.
 //
 ////===----------------------------------------------------------------------===//
 #include <assert.h>
@@ -16,18 +16,17 @@
 
 #define N 1000
 
+//test by mapping to device use 'target map' construct
 int test_target_has_device_addr() {
   OMPVV_INFOMSG("test_target_has_device_addr");
   int errors = 0;
   int x = 10;
   int arr[N];
   for(int i=0;i<N;i++){
-      arr[i] = i;
+    arr[i] = i;
   }
   int *first_scalar_device_addr;
   int *first_arr_device_addr;
-  int *second_scalar_device_addr;
-  int *second_arr_device_addr;
   #pragma omp target enter data map(to: x, arr)
   #pragma omp target map(from: first_scalar_device_addr, first_arr_device_addr) map(to: x, arr)
   {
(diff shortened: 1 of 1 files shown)


tests/5.1/target/test_target_has_device_addr.c
```diff
@@ -0,0 +1,53 @@
...
+  #pragma omp target map(from: first_scalar_device_addr, first_arr_device_addr)
+  {
+    first_scalar_device_addr = &x;
+    first_arr_device_addr = &arr[0];
+  }
+  #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr[N]) 
+  {
+    second_scalar_device_addr = &x;
+    second_arr_device_addr = &arr[0];
+  }
```
tob2, line 40: _Side note: GCC 12 will have has_device_addr support (C/C++/Fortran) – to be committed today._

Regarding the testcase: I think `arr` works with host fallback or (unified) shared memory. But `x` will likely fail reliably. Namely:

```C
  int x = 10; int arr[N];
```
There is nothing which puts `x` and `arr` into device memory. Thus, they only live on the host until ...
```C
   #pragma omp target map(from: first_scalar_device_addr, first_arr_device_addr)
    { first_scalar_device_addr = &x; first_arr_device_addr = &arr[0]; }
```
Here, the two variables are implicitly mapped – as `firstprivate(x) map(tofrom: arr)`. – Especially the `firstprivate(x)` will ensure that in the next target section, `x` is almost guaranteed to have a different address as here. As `arr` is previously not mapped, it also will be unmapped at the end of this target region.

And in the next target region, `x` and `arr` will keep their host value (due to the has_device_addr) ... 
```C
   #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr[N]) 
```
... – thus, '&x' and '&arr' should give the same value as on the host – but not a device addr. _(Ignoring issues with the semantic of 'device address' and that it will work with shared memory/host fallback.)_

How about something like the following. Unfortunately, the suggestions diff does not seem to show the changed lines. Thus, here's a list:
* Added: omp target enter/exit data
* Explicit mapping of x and arr for the first target region (actually, only required for 'x')
* Wrapped second region into a omp target data + use_device_addr
* Removed a `[N]` – cf. note below.
* Added data_x/data_arr → additional test.
```suggestion
  #pragma omp target enter data map(to: x, arr)
  #pragma omp target map(from: first_scalar_device_addr, first_arr_device_addr) map(to: x, arr)
  {
    first_scalar_device_addr = &x;
    first_arr_device_addr = &arr[0];
  }
  int *data_x, *data_arr;
  #pragma omp target data use_device_addr(x,arr)
  {
  data_x = &x; data_arr = &arr[0];
  #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr) 
  {
    second_scalar_device_addr = &x;
    second_arr_device_addr = &arr[0];
  }
  }
 #pragma omp target exit data map(release: x, arr)
 OMPVV_TEST_AND_SET(errors, first_scalar_device_addr != data_x);
 OMPVV_TEST_AND_SET(errors, first_arr_device_addr != data_arr);
```
**NOTES:**
* On the first omp target, instead of `map(to: x)`, a `defaultmap` could have been used instead. And instead of `to` using `present` would be also possible.
* I also removed the `[N]` from `has_device_addr(arr[N])`: while array sections are permitted, array elements aren't. _As an array section (for an has_device_addr list item) gives the same result as the full variable, the array section is only syntactical sugar. — Similarly for `use_device_addr`, except that that for the latter, the array section must enclose mapped array elements; thus, only if the whole array is mapped, the array section can be left out._

**Disclaimer:** I have not compiled neither the original version nor the version using my patch. And I might have missed some fine print.

tests/5.1/target/test_target_has_device_addr.c
```diff
@@ -0,0 +1,53 @@
+//===--- test_target_has_device_addr.c --------------------------------------===//
```
tmh97, line 1: Add target_alloc test section as well

tests/5.1/target/test_target_has_device_addr.c
tmh97: Hey @tob2 these changes look good! Just to clarify, could you explain why you decided to add in the target data use_device_addr? Would the code not function as expected without it?

tob2: Well, in order that you can write `has_device_addr(x, arr)`, the two variables have to have a device address. The easiest way to achieve this is to use `target data ... use_device_addr(x, arr)` that both variables have the device address inside the target-data region.

In principle, something like `int *y = omp_target_alloc(...);` would also work, which then could be directly passed to `is_device_ptr()`, except that `y` is not a device address but only points to a device address (hence: is_device_ptr and not has_device_addr). I currently do not see how I can get rid of the 'target data', but probably using a reference type can achieve this somehow.

In contrast, for Fortran, something like the the following should work:
```F90
use iso_c_binding; use omp_lib
type(c_ptr) :: cptr
integer, pointer :: x(:,:)
cptr = omp_target_alloc (...)
call c_f_pointer(cptr, x, [10, 10])
!$omp target has_device_addr(x)
  x(:,:) = 42
!$omp end target
```

nolanbaker31:
Currently getting segmentation faults.
I don't believe this is due to lack of implementation for **has_dev_addr**, as it works for simpler tests. (also was seg-faulting before my last commit)
Any assistance would be helpful!

tests/5.1/target/test_target_has_device_addr.c
```diff
@@ -0,0 +1,96 @@
...
+  {
+    data_x = &x; data_arr = &arr[0];
+    #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr) 
+    {
+      second_scalar_device_addr = &x;
+      second_arr_device_addr = &arr[0];
+    }
+  }
+  #pragma omp target exit data map(release: x, arr)
+  OMPVV_TEST_AND_SET(errors, first_scalar_device_addr != data_x);
```
tob2, line 83: I don't understand this check – shouldn't this be **second**_scalar_device_addr ? Because the code does:

```C
   // first_scalar_device_addr points to alloc'd memory
   int *first_scalar_device_addr = (int *)omp_target_alloc(sizeof(int), omp_get_default_device());

   // second_arr_device_addr contains the device addr of arr[0];.
   int arr[N];
   int *data_x, *data_arr;   
   #pragma omp target enter data map(to: x, arr)
   #pragma omp target map(from: first_scalar_device_addr, first_arr_device_addr) map(to: x, arr)
     first_arr_device_addr = &arr[0];
     
     #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr) 
     {
      second_arr_device_addr = &arr[0];
```

Additionally, I think it makes sense to actually use first_..._device_addr in the code in the target region, e.g. something like: `first_arr_device_addr[0] = 1; first_arr_device_addr[N-1] = 2; x[0] = first_arr_device_addr[0]; x[N-1] = first_arr_device_addr[N-1];`

And then check the value on the host.

tests/5.1/target/test_target_has_device_addr.c
```diff
@@ -0,0 +1,96 @@
...
+  int dev = omp_get_default_device();
+  int *first_scalar_device_addr = (int *)omp_target_alloc(sizeof(int), omp_get_default_device());
+  int *first_arr_device_addr = (int *)omp_target_alloc(sizeof(int) * N, omp_get_default_device());
+  #pragma omp target enter data map(to: x, arr)
+  int *data_x, *data_arr;
+  int *second_scalar_device_addr, *second_arr_device_addr;
+  //check addresses are same on both target data region & device region
+  #pragma omp target data use_device_addr(x,arr)
+  {
+    data_x = &x; data_arr = &arr[0];
```
tmh97, line 75: split these into two lines

spophale, line 75: why do you need this ? 
For testing it is adequate to record the device address in target1 and target2 and compare the two. I think @tob2 is saying the same thing.

tests/5.1/target/test_target_has_device_addr.c
```diff
@@ -0,0 +1,96 @@
...
+  {
+    first_scalar_device_addr = &x;
+    first_arr_device_addr = &arr[0];
+  }
+  int *data_x, *data_arr;
+  int *second_scalar_device_addr, *second_arr_device_addr;
+  //check addresses are same on both target data region & device region
+  #pragma omp target data use_device_addr(x,arr)
+  {
+    data_x = &x; data_arr = &arr[0];
```
spophale, line 41: remove.

nolanbaker31:
@spophale @tmh97 this test should be ready to go. No LLVM implementation yet, not sure about GCC 12

tests/5.1/target/test_target_has_device_addr.c
```diff
@@ -0,0 +1,54 @@
...
+  int *first_scalar_device_addr;
+  int *first_arr_device_addr;
+  #pragma omp target enter data map(to: x, arr)
+  #pragma omp target map(from: first_scalar_device_addr, first_arr_device_addr) map(to: x, arr)
+  {
+    first_scalar_device_addr = &x;
+    first_arr_device_addr = &arr[0];
+  }
+  int *second_scalar_device_addr, *second_arr_device_addr;
+  //check addresses are same on device region
```
tob2, line 37: @nolanbaker31 wrote:
> No LLVM implementation yet, not sure about GCC 12

GCC12 supports `has_device_addr` in C/C++/Fortran.

But testing showed that the testcase fails for second_..._device_addr. The problem is that the code claims that x/arr has the device address – but it actually has the host address.

How about:

```suggestion
  //check addresses are same on device region
  #pragma omp target map(from:second_scalar_device_addr, second_arr_device_addr) has_device_addr(x, arr) 
```

PR #460, opened 2022-02-09
Add new_test/test_target_allocate.F90.

seyonglee:
Test fails with both GCC11.1.0 and XLF16.1.1-10 (not supported).

OpenMP V5.0 spec says the following restriction (section 2.11.4 - allocate clause, line 17 ~ 19 on page 159): 

"allocate clauses that appear on a target construct or on constructs in a target region must specify an allocator expression unless a requires directive with the dynamic_allocators clause is present in the same compilation unit."

However, the C version (and Fortran version too) does not use an allocator expression, but the C version compiled with GCC11.1.0 passes the test.

### DIFF
--- a/tests/5.0/target/test_target_allocate.F90
+++ b/tests/5.0/target/test_target_allocate.F90
@@ -35,7 +35,7 @@ CONTAINS
     END DO
 
 
-    !$omp target allocate(x) firstprivate(x) map(from: device_result)
+    !$omp target allocate(omp_default_mem_alloc:x) firstprivate(x) map(from: device_result)
     DO i = 1, N
        x = x + i
     END DO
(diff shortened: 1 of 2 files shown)


tests/5.0/target/test_target_allocate.F90
```diff
@@ -0,0 +1,49 @@
...
+    x = 0
+    host_result = 0
+    device_result = 0
+
+    DO i = 1, N
+       host_result = host_result + i
+    END DO
+
+
+    !$omp target allocate(x) firstprivate(x) map(from: device_result)
```
spophale, line 38: allocate(omp_default_mem_alloc: x)

seyonglee: Added the default allocator to both C and Fortran version.
Still only C version compiled by GCC 11.1.0 passes.

spophale (approved):

nolanbaker31 (approved):

tmh97 (approved):

PR #461, opened 2022-02-09
Add new_test/test_target_device.F90.

seyonglee:
Fail with both GCC11.1.0 and XLF16.1.1-10 (Not supported yet)

The original C version (and corresponding Fortran version too) has several issues:
- Several variables (which_device and target_device_num) can be used without initialization in some cases.
- It is unclear what the first subtest (test_target_device_ancestor()) does.


spophale (approved):

spophale:
@tmh97 can you take a look ?

nolanbaker31:
The first test on ancestor:
> If a device clause in which the ancestor device-modifier appears is present on the target construct and the device clause expression evaluates to 1, execution of the target region occurs on the parent device of the enclosing target region. If the target construct is not encountered in a target region, the current device is treated as the parent device. https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5-1.pdf#page=221

This test seems to just ensure that the target region runs correctly on the host. To me, though, the error message:
"Target region was executed on the device... this region should execute on host" doesn't make sense... 
which_device would map back as 75 either way?

https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5-1.pdf#page=221



I am confused about the purpose of modifying arrays on the device, they aren't used / tested.

seyonglee:
> This test seems to just ensure that the target region runs correctly on the host. To me, though, the error message (line 49): "Target region was executed on the device... this region should execute on host" doesn't make sense to me... which_device would map back as 75 either way? . . . I am also confused about the purpose of the arrays, they are modified on device regions but not tested or used in any other way

I agree that the test tries to check whether the target region runs correctly on the host ,and the value of `which_device` will be the same (`75`) regardless of the executed device. Therefore, I think that the `map-type` should be changed from `tofrom` to `to`. (And `which_device` should be explicitly initialized before it is used for testing.)

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
...
+
+    host_device_num = omp_get_device_num()
+    target_device_num = host_device_num
+
+    OMPVV_TEST_AND_SET(errors, omp_get_num_devices() .le. 0) 
+    OMPVV_ERROR_IF(omp_get_num_devices() .le. 0, "Since no target devices were found, this test will be skipped.")
+
+    IF ( omp_get_num_devices() .gt. 0 ) THEN
+
+    first_device_num = omp_get_num_devices() - 1
```
tmh97, line 78:
```suggestion
    requested_device_num = omp_get_num_devices() 
```
By making this change, we can ensure that if omp_get_num_devices() == 1, the test will not fail. If omp_get_num_devices == 1, then omp_get_num_devices() will equal 0, which will result in the test running on the host (causing a fail)

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
...
+    !$omp target device(device_num: first_device_num) map(tofrom: b, target_device_num)
```
tmh97, line 79:
```suggestion
    !$omp target device(device_num: requested_device_num) map(tofrom: b, target_device_num)
```

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
...
+
+    DO i = 1, N
+       b(i) = i
+    END DO
+
+    host_device_num = omp_get_device_num()
+    target_device_num = host_device_num
+
+    OMPVV_TEST_AND_SET(errors, omp_get_num_devices() .le. 0) 
+    OMPVV_ERROR_IF(omp_get_num_devices() .le. 0, "Since no target devices were found, this test will be skipped.")
```
tmh97, line 74: Should we issue a warning instead of causing the test to fail?

tmh97:
Posted a few recommended changes for discussion in tomorrow's meeting. I can make all requested/discussed changes if we agree upon them!

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
...
+
+    IF ( omp_get_num_devices() .gt. 0 ) THEN
+
+    !$omp target device(ancestor: 1) map(tofrom: a) map(to: which_device)
+    DO i = 1, N
+       a(i) = a(i) + 2
+    END DO
+    which_device = 75
```
tmh97, line 50:
```suggestion
    which_device = omp_get_num_device()
```
Would it make sense to check that omp_get_device_num() returns the same number as omp_get_device_num() on the host?

tests/5.0/target/test_target_device.F90
spophale: We should be able to skip the test and have the label "SKIPPED" or "UNSUPPORTED". That way we can differentiate between FAILS --> something is wrong vs. SKIP--> not implemented/supported.

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
...
+! 
+!//===----------------------------------------------------------------------===//
+#include "ompvv.F90"
+
+#define N 1024
+
+! Required for device(ancestor: 1)
+!$omp requires reverse_offload
+
+PROGRAM test_target_device
```
tob2, line 17: The requires directive is misplaced – you need to move it into the specification part of the PROGRAM – and all other _compilation units_. Regarding the latter, it also needs to be set in the "ompvv_lib" module. To do so, define "OMPVV_MODULE_REQUIRES_LINE" accordingly.

Note that the meaning of _compilation unit_ differs between Fortran and C/C++:

> _compilation unit_
> For C/C++, a translation unit.
> For Fortran, a program unit.

where "program unit" is defined in Fortran as (here: F2018):
> 3.116
> **program unit**
> main program, external subprogram, module, submodule, or block data program unit (5.2.1).

Namely: Something like the following – it can be also be moved further down in PROGRAM (if that's more readable) – but it must be placed before OMPVV_TEST_OFFLOADING (which is the first exec statement and ends the specification part).

```suggestion
PROGRAM test_target_device
  ! Required for device(ancestor: 1)
  !$omp requires reverse_offload
```

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
+!===--- test_target_device.F90 -------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test checks the target construct with device clause where device-
+! modifier is either ancestor or device_num. If no device_modifier is 
+! present, the behavior is the same as if device_num were present.
+! 
+!//===----------------------------------------------------------------------===//
+#include "ompvv.F90"
```
tob2, line 10: As mentioned, the 'requires' needs to be in all _compilation units_, thus:

```suggestion
#define OMPVV_MODULE_REQUIRES_LINE !$omp requires reverse_offload
#include "ompvv.F90"
```

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
...
+
+    !$omp target device(ancestor: 1) map(tofrom: a) map(to: which_device)
+    DO i = 1, N
+       a(i) = a(i) + 2
+    END DO
+    which_device = 75
+    !$omp end target
+    END IF
+
+    OMPVV_ERROR_IF(which_device /= 75, "Target region was executed on a target device. Due to ancestor device-modifier this region
```
tob2, line 54: Can this line – and likewise lines below in "FUNCTION target_device_device_num" – and likewise in the C file be moved before the "END IF"?

Currently, with host fallback, the verbose output shows:
```
[OMPVV_ERROR test_target_device.F90:42] Since no target devices were found, this test will be skipped.
[OMPVV_ERROR test_target_device.F90:54] Target region was executed on a target device. Due to ancestor device-modifier this region should execute on a host device
```
The last line sounds odd as everything _is_ executed on the host.

tests/5.0/target/test_target_device.c
```diff
@@ -23,6 +23,7 @@ int test_target_device_ancestor() {
     int i, which_device;
```
tob2, line 23: Minor nit:
```
test_target_device.c:23:9: warning: unused variable ‘i’ [-Wunused-variable]
   23 |     int i, which_device;
      |         ^
test_target_device.c: In function ‘test_target_device_device_num’:
test_target_device.c:60:9: warning: unused variable ‘i’ [-Wunused-variable]
   60 |     int i, target_device_num, host_device_num, first_device_num;
      |         ^
```

seyonglee: Removed the unused variable i, as suggested.

spophale:
Can we have a fresh set of eyes for review ? @nolanbaker31 @krishols @AnonNick @mjcarr458 @jrreap

spophale:
@seyonglee is this test ready for review ? I see some unresolved comments and suggestions in there.

tests/5.0/target/test_target_device.F90
tob2: The problem with that is that compilers (or at least GCC) will reject this with:
`Error: OpenMP runtime API call ‘omp_get_device_num’ in a region with ‘device(ancestor)’ clause`
_(with ...num_device changes to ...device_num)_

That's the last item in the following quote from OpenMP 5.0, https://www.openmp.org/spec-html/5.0/openmpsu60.html#x86-2820002.12.5
>  * If a device clause in which the ancestor device-modifier appears is present on the construct, then the following restrictions apply:
>    * A requires directive with the reverse_offload clause must be specified;
>    * The device clause expression must evaluate to 1;
>    * Only the device, firstprivate, private, defaultmap, and map clauses may appear on the construct;
>    * No OpenMP constructs or calls to OpenMP API runtime routines are allowed inside the corresponding target region.
 * * *
However, I also have an issue with "= 75" – with unified shared memory, this likely also gets assigned to 75 even if the reverse lookup works. Thus, this will give spurious FAILS.

I think it makes sense to call some procedure from that target region – and use that as better check:
* Not fully legal, but you could smuggle in an `omp_is_initial_device()` or `omp_get_device_num()` into that procedure as a compiler usually won't diagnose this and those API routines are likely to work correctly. — Downside: Not really legal, even if it likely works.
* The call to some procedure also is a check by itself – as this will give a link-time error if it is called from the device but only has a host version. There are two ways:
   * Write a procedure yourself. — Downside: If the procedure is in the same file, compilers are supposed to auto-tag them as 'declare target' for non-reverse offload. But if the compiler has a bug, it might still do it also for 'ancestor:1'.
   * Call some library function which is known to be available on the host (POSIX function) but unlikely to exist on the device. An example could be `v = getenv ("PATH")` – and check the env var exists (`v != NULL`) and has non-zero length (`v[0] != '\0'`); in Fortran, that would be: `call get_environment_variable ("PATH", length=len)` with `len > 0` check. — Could also be another procedure, which is likely supported on the host but not on the device.

tests/5.0/target/test_target_device.F90
seyonglee: The device id expression must evaluate to a non-negative integer value less than the value of `omp_get_num_devices()` or to the value of `omp_get_initial_device()`, and thus `requested_device_num = omp_get_num_devices()` is not a valid `target device` ID.

tests/5.0/target/test_target_device.F90
seyonglee: The `requested_device_num` will not have a valid value; see the comments above.

tests/5.0/target/test_target_device.F90
seyonglee: @tob2 As you said, there will be an issue `which_device = 75` if the system supports unified shared memory, and thank you for suggesting alternative solutions, but none of which seems to be guaranteed to work.

tests/5.0/target/test_target_device.F90
seyonglee: Updated.

tests/5.0/target/test_target_device.F90
seyonglee: Updated as suggested.

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
...
```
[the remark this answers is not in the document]
seyonglee: Updated as suggested.

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,93 @@
...
+    DO i = 1, N
+       b(i) = b(i) + 2
+    END DO
+    target_device_num = omp_get_device_num()
+    !$omp end target
+
+    END IF
+
+    OMPVV_TEST_AND_SET(errors, target_device_num .eq. host_device_num)
+    OMPVV_ERROR_IF(target_device_num .eq. host_device_num, "Target region was executed on host, this region should execute on specified target device number")
```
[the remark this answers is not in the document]
seyonglee: Updated as suggested.

seyonglee:
[Update on July 19, 2022]
Remaining issues:
1) When no target device is found, should the test fail or warn?
            - @spophale "We should be able to skip the test and have the label "SKIPPED" or "UNSUPPORTED". That way we can differentiate between FAILS --> something is wrong vs. SKIP--> not implemented/supported."
2) The current ancestor device-modifier test function (`target_device_ancestor`) cannot detect errors if the target system supports unified shared memory.
            - @tob2 suggested several alternative checking methods, but none of which seems to be  guaranteed to work.

seyonglee:
Changed the behavior from error to warning when no target devices are found; just skip the test in this case with a warning message.
    Added a warning message saying that target_device_ancestor() test may not be able to detect errors if the target system supports shared memory.
        - LLVM 15.0.0: C test failed.
        - XL 16.1.1-10: Both C and Fortran tests failed.

spophale (approved):

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,100 @@
...
+    which_device = 0
+    is_shared_env = 0
+
+    DO i = 1, N
+       a(i) = i
+    END DO
+
+    OMPVV_TEST_AND_SET(errors, omp_get_num_devices() .le. 0) 
+    OMPVV_WARNING_IF(omp_get_num_devices() .le. 0, "[SKIPPED] Since no target devices were found, this test will be skipped.")
+    OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env)
```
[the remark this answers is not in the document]
seyonglee: fixed;  thanks!

tests/5.0/target/test_target_device.F90
seyonglee: Changed.

seyonglee:
[Update on September 11, 2022]
       - Modified to use nested target regions for reverse offloading test.
        - LLVM 15.0.0: C test failed.
        - XL 16.1.1-10: Both C and Fortran tests failed.

spophale:
@nolanbaker31 , @krishols , @mjcarr458 , @jrreap - can I get another review ?

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,99 @@
...
+
+
+  !$omp target ! Run on the default device, which is the host for device_num = 0
+    !$omp target device(ancestor: 1) map(tofrom: a) map(to: which_device)
+    DO i = 1, N
+       a(i) = a(i) + 2
+    END DO
+    which_device = 75
+    !$omp end target
+  !$omp end target
```
tob2, line 57: The expected value on the host is 0 not 75:  `!$omp target` has an implicit `firstprivate(which_device)` – and this firstprivatized variable is then mapped to the device explicitly (`map(to: which_device) `).

Without (unified) shared memory, the expected value after the inner/reverse-offload region is `0` (not copied back to the device) – and (with and without shared memory) on outside of all target regions, it should be 0.

I think you need something like `map(to: which_device)` on the outer `target` directive.

Additionally, given that 'a' was set, shouldn't there be some check like:
```F90
    OMPVV_TEST_AND_SET(errors, any (a /= [(i+2, i =1, N)]))
```
albeit without using `map(to: a)`, there is a `map(tofrom: a)` clause on both directives: on the outer implicitly and on the inner explicitly. Hence, also without shared memory and without working reverse offload, it would fail.

Maybe something like that (as the next line mentions, 'errors2' still needs to be declared. There is no ..._VERBOSE inside the target region as not all implementations support I/O in the target region):
```suggestion
  errors2 = 0  ! a second variable - needs to be declared
  !$omp target map(tofrom: errors2) map(to:a, which_device) ! Run on the default device, which is the host for device_num = 0
    !$omp target device(ancestor: 1) map(tofrom: a) map(to: which_device)
    DO i = 1, N
       a(i) = a(i) + 2
    END DO
    which_device = 75
    !$omp end target
    if (which_device /= 75) errors2 = errors2 + 1
    if (any (a /= [(i, i = 1, N)])) errors2 = errors2 + 1
  !$omp end target
  OMPVV_TEST_AND_SET_VERBOSE(errors, errors2 /= 0)
  OMPVV_TEST_AND_SET_VERBOSE(errors, any (a /= [(i+2, i = 1, N)]))
```

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,99 @@
...
+
+    first_device_num = omp_get_num_devices() - 1
+    !$omp target device(device_num: first_device_num) map(tofrom: b, target_device_num)
+    DO i = 1, N
+       b(i) = b(i) + 2
+    END DO
+    target_device_num = omp_get_device_num()
+    !$omp end target
+
+    OMPVV_ERROR_IF(target_device_num .eq. host_device_num, "Target region was executed on host, this region should execute on specified target device number")
```
tob2, line 91: @Shouldn't this be better the following to ensure the specified and not any non-host device is used?
```suggestion
    OMPVV_ERROR_IF(target_device_num /= first_device_num, "Target region was not executed on the specified target device number")
```

Maybe also:
`OMPVV_TEST_AND_SET_VERBOSE(errors, any (b /= [(i+2, i = 1, N)]))`

tests/5.0/target/test_target_device.F90
```diff
@@ -0,0 +1,99 @@
...
+    errors = 0
+    which_device = 0
+    is_shared_env = .false.
+
+    DO i = 1, N
+       a(i) = i
+    END DO
+
+    OMPVV_TEST_AND_SET(errors, omp_get_num_devices() .le. 0) 
+    OMPVV_WARNING_IF(omp_get_num_devices() .le. 0, "[SKIPPED] Since no target devices were found, this test will be skipped.")
```
tob2, line 45: The message is not quite true: It is not skipped but it is just testing not as much. _(And it will fail – as the host is a shared-memory device.)_

tests/5.0/target/test_target_device.c
```diff
@@ -13,28 +13,33 @@
...
+    int is_shared_env = 0;
+    which_device = 0;
 
     for (int i = 0; i < N; i++) {
         a[i] = i;
     }
 
     OMPVV_TEST_AND_SET(errors, omp_get_num_devices() <= 0);
-    OMPVV_ERROR_IF(omp_get_num_devices() <= 0, "Since no target devices were found, this test"
+    OMPVV_WARNING_IF(omp_get_num_devices() <= 0, "[SKIPPED] Since no target devices were found, this test"
```
tob2, line 34: Same remark as above - it will not be skipped, but it won't test as much and it also won't pass. _(As the host is a shared-memory device.)_

tests/5.0/target/test_target_device.c
```diff
@@ -13,28 +13,33 @@
...
                                                  "will be skipped.");
+    OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env);
+    OMPVV_WARNING_IF(is_shared_env != 0, "[WARNING] target_device_ancestor() test may not be able to detect errors if the target system supports shared memory.")
 
-    if (omp_get_num_devices() > 0) {
+    #pragma omp target //Run on the default device, which is the host for device_num = 0
```
tob2, line 39: Same issue with the implicit `firstprivate(which_device)` as above.

tests/5.0/target/test_target_device.c
```diff
@@ -82,11 +89,13 @@ int test_target_device_device_num() {
  
             target_device_num = omp_get_device_num();
         }
+
+        OMPVV_ERROR_IF(target_device_num == host_device_num, "Target region was executed on host," 
+                   "this region should execute on specified target device number");
```
tob2, line 94: Maybe _(also note the tailing space at the end of the first string)_:
```suggestion
        OMPVV_ERROR_IF(target_device_num != first_device_num, "Target region was not executed " 
                   "on the specified target device number");   
```

spophale:
Discussed on 10/17. 

Adding additional check if the outer target region has been offloaded would simplify the error testing.

tests/5.0/target/test_target_device.F90
seyonglee: Updated.

tests/5.0/target/test_target_device.c
seyonglee: Updated.

tests/5.0/target/test_target_device.c
seyonglee: Updated.

tests/5.0/target/test_target_device.c
```diff
@@ -65,10 +70,11 @@ int test_target_device_device_num() {
     }
 
     host_device_num = omp_get_device_num(); 
+    target_device_num = host_device_num;
 
     
     OMPVV_TEST_AND_SET(errors, omp_get_num_devices() <= 0);
-    OMPVV_ERROR_IF(omp_get_num_devices() <= 0, "Since no target devices were found, this test"
+    OMPVV_WARNING_IF(omp_get_num_devices() <= 0, "[SKIPPED] Since no target devices were found, this test"
```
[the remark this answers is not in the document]
seyonglee: Updated.

tests/5.0/target/test_target_device.c
seyonglee: Changed the warning message with the following:
"[WARNING] may not be able to detect errors since the target system does not have a target device."

tests/5.0/target/test_target_device.c
seyonglee: Updated the C version to have the nested target region as the Fortran version does.

tests/5.0/target/test_target_device.F90
seyonglee: Changed `map(tofrom: a)` to `map(to:a)`, and modified the intermediate error checking codes (after inner target region but before the end of the outer target region).

seyonglee:
[Update on Oct. 31, 2022]
- Add code to check whether outer target region is running on a host device or with shared memory, and conditionally check the results at the end of the outer target region (after the inner target region finishes) only if the outer target region is running on a target device with discrete memory.
            - Updated C version too.
            - C error:  line 41: error: identifier "ancestor" is undefined
            - Fortran error: REQUIRES clause 'reverse_offload' is not supported.
- LLVM 15.0.0: C test failed.
            - error: unexpected OpenMP clause 'reverse_offload' in directive '#pragma omp requires'
- GCC 12.2.1: Both C and Fortran tests failed because both ran on the host (runtime error by the program).
            - C error: Condition test_target_device_device_num() failed 
                    [SKIPPED] Since no target devices were found, this testwill be skipped
            - Fortran error: [SKIPPED] Since no target devices were found, this test will be skipped.
                    Condition target_device_device_num() .ne. 0 failed
- XL 16.1.1-10: Both C and Fortran tests failed.
            - C error: undefined reference to `omp_get_device_num'
            - Fortran error: "./ompvv/ompvv.F90", line 107.7: 1515-019 (S) Syntax is incorrect.
                    "/gpfs/alpine/csc383/scratch/f6l/sollve/tests/5.0/target/test_target_device.F90", line 17.9: 1515-019 (S) Syntax is incorrect.

spophale (approved):

spophale:
@nolanbaker31 will review.

nolanbaker31 (approved):

PR #462, opened 2022-02-14
compiler not allowing default(private) as a valid data-sharing type

mjcarr458:
This is a test for default(private) on a task construct. Currently fails when placed inside of a target region.

### DIFF
--- a/tests/5.1/default/test_task_default_private.c
+++ b/tests/5.1/default/test_task_default_private.c
@@ -23,20 +23,21 @@
 int test_task_default_private(){
 	int errors = 0;
 	int test_num= 1;
-	int temp_errors1 = 0;
+	int sum = 0;
 	int *ptr1 = &test_num;
 
-	#pragma omp task shared(temp_errors1) private(test_num)
+	#pragma omp target map(tofrom: sum, test_num)
+	{
+	#pragma omp task shared(sum) private(test_num)
 	{
 
 		int test_num = 2;
-		if (*ptr1 != 1){
-			temp_errors1 = 1;
-		}
+		sum += test_num;
 		
+	}
 	}	
-	OMPVV_TEST_AND_SET_VERBOSE(errors, temp_errors1 != 0);
-	OMPVV_INFOMSG_IF(temp_errors1 == 1, "Original value was overridden");
+	OMPVV_TEST_AND_SET_VERBOSE(errors, sum != 2);
+	OMPVV_INFOMSG_IF(sum == 1, "Did not use variable delcared in task region");
 	return errors;
 }
 
--- a/tests/5.1/default/test_task_default_private.c
+++ b/tests/5.1/default/test_task_default_private.c
@@ -24,13 +24,11 @@ int test_task_default_private(){
 	int errors = 0;
 	int test_num= 1;
 	int sum = 0;
-	int *ptr1 = &test_num;
-
+	
+	#pragma omp task shared(sum) default(private)
+	{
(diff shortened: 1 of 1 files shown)


tests/5.1/default/test_task_default_private.c
```diff
@@ -0,0 +1,65 @@
...
+		#pragma omp task private(test_num, test_arr, sum)
+		test_num += 1;
+		for (int i = 0; i<N; i++){
+			test_arr[i] = 1;
+			sum += test_arr[i];
+		}	
+
+	}
+
+	int new_sum;
```
spophale, line 45: Initialize new_sum.

tests/5.1/default/test_task_default_private.c
```diff
@@ -0,0 +1,65 @@
...
+	for (int i = 0; i<N; i++){
+		new_sum += test_arr[i];
+	}
+
+	OMPVV_TEST_AND_SET_VERBOSE(errors, test_num != 2);
+	OMPVV_INFOMSG_IF(test_num == 1, "Scalar was not private, changes made inside task region were not kept");
+	OMPVV_TEST_AND_SET_VERBOSE(errors, sum != new_sum);
+        
+	OMPVV_INFOMSG_IF(new_sum == old_sum, "Array was not private, changes made inside task region were not kept");
+	printf("Hi");
```
spophale, line 55: remove print.

tests/5.1/default/test_task_default_private.c
```diff
@@ -0,0 +1,65 @@
...
+		}	
+
+	}
+
+	int new_sum;
+	for (int i = 0; i<N; i++){
+		new_sum += test_arr[i];
+	}
+
+	OMPVV_TEST_AND_SET_VERBOSE(errors, test_num != 2);
```
spophale, line 50: I don't think we can expect the privatized value reflected here.

tests/5.1/default/test_task_default_private.c
```diff
@@ -0,0 +1,65 @@
...
+	int test_num = 1;
+	int test_arr[N];
+	int sum = 0;
+	int old_sum = 0;
+	for (int i =0; i<N; i++){
+		test_arr[i] = i;
+		old_sum += i;
+	}
+
+	#pragma target map(tofrom: test_num, test_arr, sum)
```
tob2, line 34:
```suggestion
	#pragma omp target map(tofrom: test_num, test_arr, sum)
```

spophale (approved):

tests/5.1/default/test_task_default_private.c
```diff
@@ -0,0 +1,48 @@
...
+int test_task_default_private(){
+	int errors = 0;
+	int test_num= 1;
+	int temp_errors1 = 0;
+	int *ptr1 = &test_num;
+
+	#pragma omp task shared(temp_errors1) private(test_num)
+	{
+
+		int test_num = 2;
```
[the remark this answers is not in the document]
tmh97: :)

mjcarr458:
Create shared sum variable for target clause, add private value to sum and check for correctness outside target region.

tmh97 (approved):
This test is good to go, and if we remove target{} the test compiles successfully. 

I think the compiler bug we are seeing when we place task{} inside of target{} is due to lack of implementation.

mjcarr458:
The suggestion to swap the position of task{} and target{} fixed error. However the test will still produce an error because default(private) is not implemented for task yet. Other than that this branch should be good to approve.

PR #465, opened 2022-02-16
Add test_target_map_with_close_modifier.F90

seyonglee:
new_test in tests/5.0/target/test_target_map_with_close_modifier.F90
Both GCC 11.1.0 and XLF 16.1.1-10 fail due to the unsupported feature (close modifier).


tests/5.0/target/test_target_map_with_close_modifier.F90
```diff
@@ -0,0 +1,66 @@
...
+
+    errors = 0
+
+    DO i = 1, N
+       a(i) = i
+       member%b(i) = i
+    END DO
+
+    !$omp target map(close,tofrom : scalar, a, member)
+    scalar = 25
```
spophale, line 48: Minor  nitpick, I would recommend using value of "scalar" to update value that way it can check if it was mapped to accurately.

spophale (approved):

tests/5.0/target/test_target_map_with_close_modifier.F90
seyonglee: Then, change "scalar = 25" to "scalar = scalar + 25"? 
We can change other statements similarly. (We will have to change both C and Fortran versions.)

spophale: yes, it would be a more complete test IMO.

seyonglee: Updated both C and Fortran versions.
Tested with GCC 11.1, IBM XL 16.1.1-10, NVHPC 21.11, and LLVM 15.0, but the C version passes only when compiled by NVHPC 21.11. (Fortran version fails all.)

tests/5.0/target/test_target_map_with_close_modifier.F90
```diff
@@ -0,0 +1,66 @@
...
+
+CONTAINS
+  INTEGER FUNCTION test_close_modifier()
+    INTEGER, DIMENSION(N) :: a
+    INTEGER :: errors, i, scalar
+
+    TYPE memberT
+    INTEGER :: var
+    INTEGER, DIMENSION(N) :: b
+    ENDTYPE memberT
```
tmh97, line 33: I think this should be `END TYPE memberT`

tmh97 (approved):
good-to-go

PR #466, opened 2022-02-16
Add test_target_in_reduction.F90

seyonglee:
Add a new test (tests/5.0/target/test_target_in_reduction.F90).
Both GCC 11.1.0 and XLF 16.1.1-10 fail due the unsupported features ("parallel master", "target in_reduction", etc.)

### DIFF
--- a/tests/5.0/target/test_target_in_reduction.F90
+++ b/tests/5.0/target/test_target_in_reduction.F90
@@ -1,0 +1,75 @@
+!===--- test_target_in_reduction.F90 -------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test checks task reductions for a target task resulting from a target
+! construct with the 'in_reduction' clause.
+! 
+!//===----------------------------------------------------------------------===//
+#include "ompvv.F90"
+
+#define N 1028
+
+PROGRAM test_target_in_reduction
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  IMPLICIT NONE 
+  INTEGER :: i, host_reduction_sum, device_reduction_sum
+  INTEGER :: sum, total, errors
+
+  host_reduction_sum = 0
+  device_reduction_sum = 0
+  sum = 0
+  total = 0
+  errors = 0 
+
+  !$omp parallel master
+  !$omp taskgroup task_reduction(+:sum)
+  !$omp target in_reduction(+:sum)
+  compute_on_device(sum)
+  !$omp end target
+  !$omp task in_reduction(+:sum)
+  compute_on_host(sum)
+  !$omp end task
+  !$omp end taskgroup
+  !$omp end parallel master 
+  
+  DO i = 1, N
+    device_reduction_sum = device_reduction_sum + 2
+    host_reduction_sum = host_reduction_sum + 1
+  END DO
+
+  OMPVV_TEST_VERBOSE(sum /= (device_reduction_sum + host_reduction_sum))
+
+  OMPVV_TEST_VERBOSE(sum == device_reduction_sum)
+  OMPVV_ERROR_IF(sum == device_reduction_sum, "Host task did not participate in the reduction")
+
+  OMPVV_TEST_VERBOSE(sum == host_reduction_sum)
+  OMPVV_ERROR_IF(sum == host_reduction_sum, "Target task did not participate in the reduction")
+
+  OMPVV_REPORT_AND_RETURN()
+
+CONTAINS
+  SUBROUTINE compute_on_device(sum)
+    IMPLICIT NONE
+    INTEGER, INTENT(INOUT) :: sum
+    INTEGER :: i
+
+
+    DO i = 1, N
+       sum = sum + 2
+    END DO
+  END SUBROUTINE compute_on_device
+
+  SUBROUTINE compute_on_host(sum)
+    IMPLICIT NONE
+    INTEGER, INTENT(INOUT) :: sum
+    INTEGER :: i
+
+
+    DO i = 1, N
+       sum = sum + 1
+    END DO
+  END SUBROUTINE compute_on_host
+END PROGRAM test_target_in_reduction


tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,75 @@
...
+  INTEGER :: i, host_reduction_sum, device_reduction_sum
+  INTEGER :: sum, total, errors
+
+  host_reduction_sum = 0
+  device_reduction_sum = 0
+  sum = 0
+  total = 0
+  errors = 0 
+
+  !$omp parallel master
```
spophale, line 27: master is deprecated , we should use "masked" here.

tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,75 @@
...
+  !$omp taskgroup task_reduction(+:sum)
+  !$omp target in_reduction(+:sum)
+  compute_on_device(sum)
+  !$omp end target
+  !$omp task in_reduction(+:sum)
+  compute_on_host(sum)
+  !$omp end task
+  !$omp end taskgroup
+  !$omp end parallel master
```
spophale, line 36: master --> masked

tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,75 @@
...
+  host_reduction_sum = 0
+  device_reduction_sum = 0
+  sum = 0
+  total = 0
+  errors = 0 
+
+  !$omp parallel master
+  !$omp taskgroup task_reduction(+:sum)
+  !$omp target in_reduction(+:sum)
+  compute_on_device(sum)
```
spophale, line 30: 'call' missing

tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,75 @@
...
+  !$omp end target
+  !$omp task in_reduction(+:sum)
+  compute_on_host(sum)
```
spophale, line 33: 'call' missing

tests/5.0/target/test_target_in_reduction.F90
seyonglee: The "masked construct" was added from OpenMP V5.1, but this test is for OpenMP V5.0.

tests/5.0/target/test_target_in_reduction.F90
seyonglee: The "masked construct" was added from OpenMP V5.1, but this test is for OpenMP V5.0.

tests/5.0/target/test_target_in_reduction.F90
seyonglee: Thank you for catching this.

tests/5.0/target/test_target_in_reduction.F90
seyonglee: Thank you for catching this. Updated.

tests/5.0/target/test_target_in_reduction.F90
spophale: yes, apologies.

tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,75 @@
...
+  device_reduction_sum = 0
+  sum = 0
+  total = 0
+  errors = 0 
+
+  !$omp parallel master
+  !$omp taskgroup task_reduction(+:sum)
+  !$omp target in_reduction(+:sum)
+  CALL compute_on_device(sum)
+  !$omp end target
```
spophale, line 31: Should there a task wait between the two sibling tasks as they modify the same 'sum' variable ?

seyonglee: In this test case, both "target" tasks participate in the reduction specified by the enclosing taskgroup. The final reduction will be performed at the end of the taskgroup region with an implicit wait, and thus we don't need to insert a task wait.

spophale (approved):

tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,75 @@
...
+  IMPLICIT NONE 
+  INTEGER :: i, host_reduction_sum, device_reduction_sum
+  INTEGER :: sum, total, errors
+
+  host_reduction_sum = 0
+  device_reduction_sum = 0
+  sum = 0
+  total = 0
+  errors = 0 
+
```
tob2, line 26: Compiles and runs with GCC 12 – but shows on my non-offloading laptop:
`[OMPVV_RESULT test_target_in_reduction.F90] Test passed on the device.` (device!)
```suggestion

OMPVV_TEST_OFFLOADING

```

tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,77 @@
...
+    device_reduction_sum = device_reduction_sum + 2
+    host_reduction_sum = host_reduction_sum + 1
+  END DO
+
+  OMPVV_TEST_VERBOSE(sum /= (device_reduction_sum + host_reduction_sum))
+
+  OMPVV_TEST_VERBOSE(sum == device_reduction_sum)
+  OMPVV_ERROR_IF(sum == device_reduction_sum, "Host task did not participate in the reduction")
+
+  OMPVV_TEST_VERBOSE(sum == host_reduction_sum)
```
tmh97, line 50: Change to OMPVV_TEST_AND_SET_VERBOSE() or OMPVV_TEST_AND_SET()

tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,77 @@
...
+  !$omp end task
+  !$omp end taskgroup
+  !$omp end parallel master 
+  
+  DO i = 1, N
+    device_reduction_sum = device_reduction_sum + 2
+    host_reduction_sum = host_reduction_sum + 1
+  END DO
+
+  OMPVV_TEST_VERBOSE(sum /= (device_reduction_sum + host_reduction_sum))
```
tmh97, line 45: Change to OMPVV_TEST_AND_SET_VERBOSE() or OMPVV_TEST_AND_SET()

tests/5.0/target/test_target_in_reduction.F90
```diff
@@ -0,0 +1,77 @@
...
+
+  OMPVV_TEST_VERBOSE(sum == device_reduction_sum)
```
tmh97, line 47: Change to OMPVV_TEST_AND_SET_VERBOSE() or OMPVV_TEST_AND_SET()

tmh97:
When I change from:
!$omp parallel master
!$omp end parallel master to

!$omp parallel
!$omp master
!$omp end master
!$omp end parallel 
....

I run into SIGSEV seg fault (gcc 11.2.0)

spophale:
@tmh97 Did the same change pass with LLVM (trunk) ?

tmh97:
> @tmh97 Did the same change pass with LLVM (trunk) ?

I can't test with LLVM trunk since it does not include an implementation of Flang, I can test the .c file if that would be helpful.
Overall I think any tests including 'master' clause will fail on newer compilers that do not recognize that directive any longer

spophale:
We can try with CCE. C test could be adequate but may not exercise the same path as Fortran (whenever that is ready). Using "masked" should work where master fails.

tests/5.0/target/test_target_in_reduction.F90
seyonglee: Do not need to change in Fortran.

tests/5.0/target/test_target_in_reduction.F90
seyonglee: Do not need to change in Fortran.

tests/5.0/target/test_target_in_reduction.F90
seyonglee: Do not need to change in Fortran.

seyonglee:
> When I change from: !$omp parallel master !$omp end parallel master to
> 
> !$omp parallel !$omp master !$omp end master !$omp end parallel ....
> 
> I run into SIGSEV seg fault (gcc 11.2.0)

GCC 12.1.1 passes with the above changes.

seyonglee:
[Update on July 20, 2022]
- GCC 12.1.1: Both C and Fortran tests pass.
      - Fortran test still passes when `!$ omp parallel master` is changed to `!$omp parallel !$omp master`. 
- XL 16.1.1-10: C test passes while Fortran test fails.
- NVHPC 22.5: Both C and Fortran tests fail.
- LLVM 15.0.0: C test passes.

krishols (approved):
Looks good. Good catch with the .c test version as well.

PR #468, opened 2022-02-23
Add tests/5.0/target/test_target_task_depend_mutexinoutset.F90

seyonglee:
Add tests/5.0/target/test_target_task_depend_mutexinoutset.F90
Fails when compiled by GCC 11.1.0 and XLF 16.1.1-10 due to not supported feature (mutexinoutset).


spophale (approved):

tmh97 (approved):
this is good to go!

PR #475, opened 2022-02-28
New metadirective test for the nothing directive

jrreap:
Added a new test for the nothing directive in metadirective constructs


tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,53 @@
...
+   int errors = 0;
+   int base_threads = 0;
+   int threads = 0;
+
+   #pragma omp target map(from:base_threads,threads)  
+   {
+	   	#pragma omp atomic write
+	 	base_threads = omp_get_num_threads();
+      	 	
+		#pragma omp metadirective \
```
tmh97, line 28: device={kind(nohost)}

jrreap: Implemented and used to leverage a sort of "compile time" test

tmh97 (approved):
This test looks good to go, can you make sure all of the when and default statements are following the same indenting pattern as well?

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,79 @@
...
+#define N 1024
+
+int metadirectiveOffload() {
+   int errors = 0;
+   int scalar = 0;
+   int A[N];
+   for (i = 0; i < N; i++) {
+      A[i] = 0;
+   }
+
```
tmh97, line 28:
```suggestion
int num_devices = omp_get_num_devices();
```

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,105 @@
...
+      when( device={kind(nohost)}: nothing) \
+      default( target map(to:from scalar, A) )
+      {
+         scalar = 10;
+         for (i = 0; i < N; i++) {
+            A[i] = i + 2;
+         }
+      }
+
+
```
tmh97, line 55:
```suggestion
OMPVV_INFOMSG("Test ran with a number of available devices greater than 0");
    OMPVV_INFOMSG_IF(A[0] == 0, "Test recognized device was of arch/vendor/kind nvidia, amd, or nohost");
    OMPVV_WARNING_IF(A[0] == 2 || A[0] == 4, "Test could not recognize is device was of arch/vendor/kind nvidia, amd or, nohost, even though there was are devices available.");
 
    for (i = 0; i < N; i++) {
	OMPVV_TEST_AND_SET(A[i] == 0 && A[i] != 2 && A[i] != 4 ) {
	   errors++;
	}
    }
```

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,105 @@
...
+#include "ompvv.h"
+
+#define N 1024
+
+int metadirectiveOnDevice() {
+   int errors = 0;
+   int scalar = 0;
+   int A[N];
+
+   for (i = 0; i < N; i++) {
```
tob2, line 27: Missing `int i`; before this loop or `for (int i` in this and the following loops.

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,105 @@
...
+
+
+  for (i = 0; i < N; i++) {
+     if (A[i] != 0) {
+       errors++;
+     }
+  }
+
+
+  OMPVV_TEST_AND_SET_VERBOSE(errors)
```
tob2, line 63: Compiler complains:
`error: macro "OMPVV_TEST_AND_SET_VERBOSE" requires 2 arguments, but only 1 given`

Additionally:
`warning: variable ‘scalar’ set but not used [-Wunused-but-set-variable]`

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,105 @@
...
+     {
+        scalar = 10;
+        for (i = 0; i < N; i++) {
+           A[i] = i + 2;
+        }
+     }
+
+
+  OMPVV_WARNING_IF(A[0] == 0,"Even though no devices were available the test recognized kind/arch equal to nohost or nvptx or amd");
+  OMPVV_ERROR_IF(A[0] != 0 && A[0] != 2, errors)
```
tob2, line 90: `error: expected ‘)’ before ‘errors’`

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,105 @@
...
+} 
+
+int main () {
+  int errors = 0;
+  OMPVV_TEST_OFFLOADING;
+
+  if (omp_get_num_devices() > 0) {
+    metadirectiveOnDevice();
+  } else {
+    metadirectiveOnHost();
```
tob2, line 100: Shouldn't this be `errors = meta...` or `errors += meta...`?

Additionally, there is the following warning for those two meta... functions:
`warning: control reaches end of non-void function [-Wreturn-type]`

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,116 @@
...
+      when( device={arch("amd")}: nothing ) \
+      default( target map(to:from scalar, A) )
+      {
+         scalar = 10;
+         for (int i = 0; i < N; i++) {
+            A[i] = i + 2;
+         }
+      }
+
+   #pragma omp metadirective \
```
tob2, line 43: Before the #pragma, I expect an error checking — even if checking after both loops also works.

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,116 @@
...
+      {
+         scalar = 10;
+         for (int i = 0; i < N; i++) {
+            A[i] = i + 2;
+         }
+      }
+
+   OMPVV_INFOMSG("Test ran with a number of available devices greater than 0");
+   OMPVV_INFOMSG_IF(A[0] == 0, "Test recognized device was of arch/vendor/kind nvidia, amd, or nohost");
+   OMPVV_WARNING_IF(A[0] == 2 || A[0] == 4, "Test could not recognize if device was of arch/vendor/kind nvidia, amd or, nohost, even though there are devices available.");
```
tob2, line 57: The first message I don't understand — it is also output for `omp_get_num_devices() == 0`.

Additionally: I note that `scalar` is not checked for – resulting in the warning (here and in the other function):
`warning: variable ‘scalar’ set but not used [-Wunused-but-set-variable]`

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,116 @@
...
+
+   for (int i = 0; i < N; i++) {
+	    OMPVV_TEST_AND_SET(A[i] == 0 && A[i] != 2 && A[i] != 4 ) {
+	      errors++;
+	    }
```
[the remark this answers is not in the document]
jrreap: Oh yes I meant to fix that from your previous comment! Also thank you for all the help! Haven't had a chance to run this under GCC 12 (since before no one had support for the nothing directive) so as expected there are plenty of syntax errors

tests/5.1/metadirective/test_metadirective_nothing.c
jrreap: Good point, since theoretically something could go wrong in between. I'll add a check there too for extra safety

tests/5.1/metadirective/test_metadirective_nothing.c
jrreap: Scalar is mostly redundant (since if the array is not changed we know the test worked) so I cleaned that up. 

However that first message is only run from the on device method. It should only appear whenever `omp_get_num_devices() > 0` per the if statement (line 101) in the `main` function. The first message is meant to let the user know that this test ran on device and not host, maybe it could be reworded to be more clear?

tob2: > However that first message is only run from the on device method. It should only appear whenever `omp_get_num_devices() > 0` per the if statement (line 101) in the `main` function. The first message is meant to let the user know that this test ran on device and not host, maybe it could be reworded to be more clear?

Missed this – as I focused more on single lines, loosing the big context.

Regarding GCC 12: I think I have to disappoint you. While GCC 12 supports the **nothing** directive, it does not support metadirectives (missed the feature-freeze deadline); there is an initial implementation for metadirectives in the OG11 branch (on the GCC git; actual git name: 'devel/omp/gcc-11' + based on GCC 11, soon to be replaced by OG12 ). Thus, this has to wait for GCC 13/mainline.

However, this does not prevent me from doing some testing. :-)

* * *

Additionally, I think _either_ you _or_ I misunderstand how the nothing directive expands. In my understanding,
```C
   #pragma omp metadirective \
      when( device={kind(nohost)}: nothing ) \
      when( device={arch("nvptx")}: nothing) \
      when( device={arch("amd")}: nothing ) \
      default( target map(tofrom: A) )
      {
         for (int i = 0; i < N; i++)  A[i] = i + 2;
      }
```
becomes (when `nothing` matches) equivalent to
```C
#pragma omp nothing
      {
         for (int i = 0; i < N; i++)  A[i] = i + 2;
      }
```
which in turn is equivalent to
```C
      {
         for (int i = 0; i < N; i++)  A[i] = i + 2;
      }
```
Thus, the loop is still executed with `nothing` — but on the host and not on the device. Thus, with `tofrom`, I actually expect the same result for the code. 

However, if the result for `A` is the same,
` OMPVV_INFOMSG_IF(A[0] == 0, "Test recognized device was of arch/vendor/kind nvidia, amd, or nohost");`

will never trigger – either `omp metadirective` is not recognized (as with GCC <= 12 (+ current mainline/GCC13)) or if it is recognized, any of `nothing`, execution on the host or device execution (due to `tofrom` and/or shared memory) should give the loop value (`A[i] = i + 2`).

**Solution:** Maybe: call `omp_is_initial_device()` in the `{ ... }` block for checking? And/or use `firstprivate` instead of `tofrom`?

jrreap: Ah gotcha, well I appreciate you running the test on GCC :) Makes catching bugs a lot easier when you can actually (mostly) compile the code!

That said, I see what you're getting at. I think I got confused from an example in the 5.1 example doc (page 460) which in part looks like:
```C
#pragma omp metadirective when(user={condition(run_parallel)}: parallel)
{
  if(omp_in_parallel() == 1 && omp_get_thread_num() == 0)
    {printf("PASSED 2 of 3\n");}

  #pragma omp metadirective \
     when( construct={parallel}, \
     user={condition(unbalanced)}: for schedule(guided) private(b)) \
     when( construct={parallel} : for schedule(static))
  for (int i=0; i<n; i++) {a[i]=i; if(i==n-1) b=1;}
}
``` 
Which I interpreted as the block of code being run _if the directive resolved to a construct_ however thinking back this does not line up with how the nothing directive really works as you point out, so yes I think I assumed a bit too quickly how the directive really worked. Thank you for the awesome explanation!

`firstprivate` may be a way to go with the target directive, but maybe also using something like `omp_in_parallel` might work too to check and see if a parallel region was generated, or nothing (which would mean it'd be 0). Could be a cleaner approach possibly.

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,109 @@
...
+        for (int i = 0; i < N; i++) {
+           A[i] += omp_in_parallel();
+        }
+     }
+
+
+  OMPVV_WARNING_IF(A[0] == 0, "Even though no devices were available the test recognized kind/arch equal to nohost or nvptx or amd");
+  
+  for (int i = 0; i < N; i++) {
+     OMPVV_TEST_AND_SET(errors, A[i] == 0 && A[i] != 1)
```
tmh97, line 90: if A[i] ==0 (the case where kind arch or vendor is recognized and nothing comes into play) this test will fail. I think we want to remove this check)

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,109 @@
...
+
+   for (int i = 0; i < N; i++) {
+      A[i] = 0;
+   }
+
+   #pragma omp metadirective \
+      when( device={kind(nohost)}: nothing ) \
+      when( device={arch("nvptx")}: nothing) \
+      when( implementation={vendor(amd)}: nothing ) \
+      default( target parallel map(tofrom: A) )
```
spophale, line 34: In this case is the expectation that the default is never reached ? If so, the current context selectors don't cover all cases and is not future-safe. Since we are still on the host it will pass the the when in 31 and if it is not NVIDIA or AMD it will go to default. 
If we want to make sure that it NEVER reaches default, we could enclose the metadirective inside a target region (and remove "target" from default directive variant.

spophale, line 34: In this case is the expectation that the default is never reached ? If so, the current context selectors don't cover all cases and is not future-safe. Since we are still on the host it will pass the the when in 31 and if it is not NVIDIA or AMD it will go to default. 
If we want to make sure that it NEVER reaches default, we could enclose the metadirective inside a target region (and remove "target" from default directive variant.

jrreap: Not necessary since metadirectives have the ability to query device info outside of a target region.

tmh97 (approved):

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,96 @@
...
+   for (int i = 0; i < N; i++) {
+      A[i] = 0;
+   }
+
+   // We expect at least one of these when conditons to eval to true, thus having the nothing directive utilized
+   #pragma omp metadirective \
+      when( device={kind(nohost)}: nothing ) \
+      when( device={arch("nvptx")}: nothing) \
+      when( implementation={vendor(amd)}: nothing ) \
+      default( target parallel map(tofrom: A) )
```
spophale, line 35: should be "target parallel for"

tests/5.1/metadirective/test_metadirective_nothing.c
spophale: looks good.

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,96 @@
...
+            A[i] += omp_in_parallel();
+         }
+      }
+
+   for (int i = 0; i < N; i++) {
+      OMPVV_TEST_AND_SET(errors, A[i] != 0 || A[i] != 1);
+   }
+
+   OMPVV_INFOMSG("Test ran with a number of available devices greater than 0");
+   OMPVV_INFOMSG_IF(A[0] == 0, "Test recognized device was of arch/vendor/kind nvidia, amd, or nohost");
```
spophale, line 47: The test is running from the host...it will ALWAYS go for the default option. Even if Nvidia or AMD devices are available, if they are not on the said device they will always go to default.

jrreap: I was under the impression that the metadirective would query device info regardless of the "region" it was in. It's checking device info and then converting into a respective construct. A very big use case for metadirectives would be determining changes at runtime based on the device being used, or whether to run something on host or parallel depending on the system. If it needed to be in a target region to access that information, then the metadirective feels more redundant since you would need to program a target region and test it on multiple devices, or add preprocessor #if statements to get that dynamic logic.

spophale: No, it does not work that way. Think of a system that has both AMD and Nvidia GPU devices...what would happened then ? Metadirective checks what "is" and makes the selection. Yes, your point is correct, it will make a check at runtime to check which device is being used and make the appropriate selection. *Being used* here means that the code is already offloaded. Even with the code offloaded you could do different things on an NVIDIA device (say use loop) vs.  AMD ( teams distribute parallel for simd) as NVIDIA has better loop implementation.

jrreap: Gotcha, fair enough. Thank you for the explanation, I'll patch it now.

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,99 @@
...
+  int errors = 0;
+  int A[N];
+
+  for (int i = 0; i < N; i++) {
+     A[i] = 0;
+  }
+
+  // We expect all of these when statements to eval to false, causing body of code to run in parallel
+  #pragma omp metadirective \
+     when( device={kind(nohost)}: nothing ) \
```
spophale, line 66: host

tmh97, line 66: make symmetric to top function (will explain this to ya later @jrreap) idea is to not have it meet default case

jrreap: Sounds good, I'll leave this unresolved for now then

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,98 @@
...
+  int A[N];
+
+  for (int i = 0; i < N; i++) {
+     A[i] = 0;
+  }
+
+  // We expect all of these when statements to eval to false, causing body of code to run in parallel
+  #pragma omp metadirective \
+     when( device={kind(host)}: nothing ) \
+     when( implementation={vendor(amd)}: nothing ) \
```
spophale, line 67: don't use nothing here --  it needs to be a construct that won't be encountered...say something like:
when( construct={target}: distribute parallel for )

Since it is not called from the target directive, it should not execute.

tmh97: So we are trying to make this function match the functionality of the metadirectiveOnDevice() function.
The idea is, we **want** the when statements to evaluate to be true such that the code in the body of the function operates the same as OpenACC Serial.

We figured that by using the device={kind(host)} and implementation={vendor(amd)} that is it likely one of those will evaluate to true. Since there is not a lot of readily available information regarding context selectors, it is very tough to form when() statements that will evaluate to true when running on host.

We should either
1) Add more context-selectors to 'guarantee' that one of the when()'s in this function will evaluate to true and thus will operate like OpenACC serial (then we use the omp_is_parallel to tell whether the when()'s evaluate to true)
2) Revert to the previous commit **which does the opposite of the metadirectiveOnDevice function**. The opposite, meaning that the previous commit of this function expects the when()'s to evaluate to **false** and thus we expect the metadirective to run the default. This behavior doesn't really test 'nothing' properly, and we agreed in a few meetings ago we should shoot for option 1) to make the functions symmetrical and make the functions perform similar testing strategy

tmh97: I see what you are saying though ... it might be better for the metadirectiveOnHost function to have exactly what you said, directives that are not encountered! This way, we can make the default behavior **nothing** and then put a bunch of ridiculous context selectors that should not evaluate to true on the host (like arch=nvidia or device=nohost)

tmh97: This way we test both

- In the metadirectiveOnDevice, we test that **nothing** will work when paired with a **when** statement
- In the metadirectiveOnHost, we test that **nothing** will work when used as the **default** statement

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,98 @@
...
+   }
+
+   #pragma omp target map(tofrom: A)
+   {
+      // We expect at least one of these when conditons to eval to true, thus having the nothing directive utilized
+      #pragma omp metadirective \
+         when( device={kind(nohost)}: nothing ) \
+         when( device={arch("nvptx")}: nothing) \
+         when( implementation={vendor(amd)}: nothing ) \
+         default( parallel for map(tofrom: A) )
```
tob2, line 37:
```
test_metadirective_nothing.c:37:32: error: ‘map’ is not valid for ‘#pragma omp parallel for’
   37 |          default( parallel for map(tofrom: A) )
      |                                ^~~
```

spophale: Thanks @tob2 ! Yes, the map needs to go with the enclosing target directive.

tests/5.1/metadirective/test_metadirective_nothing.c
```diff
@@ -0,0 +1,99 @@
...
+     A[i] = 0;
+  }
+
+  // We expect all of these when statements to eval to false, causing body of code to run using 'nothing' as the default pragma
+  #pragma omp metadirective \
+     when( device={kind(nohost)}: parallel for ) \
+     when( device={arch("nvptx")}: parallel for) \
+     when( implementation={vendor(amd)}: parallel for ) \
+     default( nothing )
+     {
```
spophale, line 70: Unless I am missing something, the for loop will still be executed without a directive variant.

jrreap: Correct, but then that omp_in_parallel (which runs regardless) will return 0 as the nothing directive was hit. Thus the test would pass since the nothing directive properly rendered in the default case. If one of the above cases first before or if the nothing directive still runs in parallel, that for loop function would catch the error and fail the test.

spophale (approved):

krishols (approved):
Looks good to me - great coverage of different cases.

PR #476, opened 2022-03-01
Add new_test/test_parallel_master.F90

seyonglee:
Fails with both XLF 16.1.1-10 and GCC 11.1.0 (not supported feature: "parallel master")

### DIFF
--- a/tests/5.0/parallel_master/test_parallel_master.F90
+++ b/tests/5.0/parallel_master/test_parallel_master.F90
@@ -28,6 +28,8 @@ CONTAINS
     INTEGER, DIMENSION(N):: x, y, z
     INTEGER:: i
 
+    OMPVV_INFOMSG("test_parallel_master")
+
     DO i = 1, N
        x(i) = 1
        y(i) = i + 1
--- a/tests/5.0/parallel_master/test_parallel_master.F90
+++ b/tests/5.0/parallel_master/test_parallel_master.F90
@@ -1,0 +1,56 @@
+!/===--- test_parallel_master.F90 -------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test checks the parallel master directive. The test performs simple
+! operations on an int array which are then checked for correctness.
+!
+!/===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_parallel_master
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
+  OMPVV_TEST_VERBOSE(parallel_master() .ne. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+
+CONTAINS
+  INTEGER FUNCTION parallel_master()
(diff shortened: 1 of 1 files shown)


spophale (approved):

tests/5.0/parallel_master/test_parallel_master.F90
```diff
@@ -0,0 +1,58 @@
...
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_parallel_master
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
```
tob2, line 19: How about adding:
```suggestion

OMPVV_TEST_OFFLOADING
```
At least it looks odd to see without offloading: "Test passed on the device."

seyonglee: The current implementations do not use target constructs, and thus we don't need  the offloading test. 
@tob2 Do you suggest to change this test to use target construct to offload?

tmh97 (approved):

tests/5.0/parallel_master/test_parallel_master.F90
tob2: Well, I just see that 

1. that currently **all** files under `tests/` have a call to it. (Except for https://github.com/SOLLVE/sollve_vv/tree/master/tests/5.0/target_update/test_target_update_mapper_to_discontiguous.c which has _two_ calls)
2. Printing "Test passed on the **device.**" if there is no offloading is confusing!

I leave it to the **main developers to decide** whether it should be required for code which does not use offloading.

However, **if** it it is optional/**not called,** I would like to see **some fix** applied for Fortran to solve issue (2). Either following patch:
```diff
--- a/ompvv/ompvv.F90
+++ b/ompvv/ompvv.F90
@@ -108 +108 @@ OMPVV_MODULE_REQUIRES_LINE
-    LOGICAL, PRIVATE :: ompvv_isHost
+    LOGICAL, PRIVATE :: ompvv_isHost = .true.
```
Assuming that it never gets forgotten if there is a target region. Or using a tri-state like as done for C/C++ (https://github.com/SOLLVE/sollve_vv/blob/master/ompvv/ompvv.h), namely:
```C
int _ompvv_isOffloadingOn = -1;
...
  if (_ompvv_isOffloadingOn == -1) \
    printf("[OMPVV_RESULT: %s] Test %s.\n", __FILENAME__, (err == 0)? "passed":"failed"); \
  else \
    printf("[OMPVV_RESULT: %s] Test %s on the %s.\n", __FILENAME__, (err == 0)? "passed":"failed", (_ompvv_isOffloadingOn)? "device" : "host"); \
```

spophale: I agree with @seyonglee . We should only add OMPVV_TEST_OFFLOADING when using target offload constructs.
@tob2 I agree with the Fortran patch. It should be on host unless explicitly set to target, since that is the convention we are following in the tests.
I will create a separate issue/PR for that and resolve this conversation so that we move ahead with merging this test.

PR #477, opened 2022-03-01
New test/test parallel master device.f90

seyonglee:
Fails with both XLF 16.1.1-10 and GCC 11.1.0 (not supported feature: parallel master)


spophale (approved):

nolanbaker31 (approved):
construct deprecated in 5.1

tmh97 (approved):

PR #478, opened 2022-03-01
Add new_test/test_parallel_master_taskloop.F90

seyonglee:
Fails with both XLF 16.1.1-10 and GCC 11.1.0 (not supported feature)

### DIFF
--- a/tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
+++ b/tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
@@ -39,10 +39,10 @@ CONTAINS
     !$omp parallel master taskloop num_threads(OMPVV_NUM_THREADS_HOST) shared(x, y, z, num_threads) 
     DO i = 1, N
        x(i) = x(i) + y(i) * z(i)
+       IF (i .eq. 1) THEN
+          num_threads = omp_get_num_threads()
+       END IF
     END DO
-    IF (i .eq. 1) THEN
-       num_threads = omp_get_num_threads()
-    END IF
     !$omp end parallel master taskloop
 
     DO i = 1, N
--- a/tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
+++ b/tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
@@ -1,4 +1,4 @@
-!/===--- test_parallel_master_taskloop.F90 -------------------------------------===//
+!/===--- test_parallel_master_taskloop.F90 --------------------------------===//
 !
 ! OpenMP API Version 5.0 Nov 2018
 !
--- a/tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
+++ b/tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
@@ -17,6 +17,8 @@ PROGRAM test_parallel_master_taskloop
   USE omp_lib
   implicit none
 
+  OMPVV_TEST_OFFLOADING
+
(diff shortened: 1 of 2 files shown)


tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
```diff
@@ -0,0 +1,58 @@
...
+       x(i) = 1
+       y(i) = i + 1
+       z(i) = 2*(i + 1)
+    END DO
+
+    !$omp parallel master taskloop num_threads(OMPVV_NUM_THREADS_HOST) shared(x, y, z, num_threads) 
+    DO i = 1, N
+       x(i) = x(i) + y(i) * z(i)
+    END DO
+    IF (i .eq. 1) THEN
```
spophale, line 43: Since IF is outside the loop it will never be executed.

spophale: IMO since master is the only active thread in this context, it is okay to not guard this statement.

tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
```diff
@@ -0,0 +1,58 @@
...
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_parallel_master_taskloop
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
```
tob2, line 19: I think you want to add:
```suggestion

    OMPVV_TEST_OFFLOADING
```

tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
tob2: > Since IF is outside the loop it will never be executed.

Additionally, as the _end-directive_ is optional for _loop_ directives like **taskloop**, the `parallel master taskloop` ends with `end do` and GCC 12 rightly complains 4 lines later:
```
   46 |     !$omp end parallel master taskloop
      |                                      1
Error: Unexpected !$OMP END PARALLEL MASTER TASKLOOP statement at (1)
```

tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
seyonglee: Added OMPVV_TEST_OFFLOADING to both C and Fortran versions.

seyonglee:
Moved the If-statement inside the DO-loop in the Fortran version.

tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
seyonglee: Fixed; thanks.

tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
```diff
@@ -0,0 +1,60 @@
+!/===--- test_parallel_master_taskloop.F90 -------------------------------------===//
```
tmh97, line 1: adjust header length

tmh97 (approved):
Test looks good to go, however it fails with compilation errors using amdflang on spock (-fopenmp-version=50 flag used to produce behavior before master was deprecated)

tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop.F90
seyonglee: Adjusted; thanks.

spophale (approved):

PR #479, opened 2022-03-01
Add new_test/test_parallel_master_taskloop_device.F90

seyonglee:
Fails with both XLF 16.1.1-10 and GCC 11.1.0 (not supported feature)
Update  test_parallel_master_taskloop_device.c too.

### DIFF
--- a/tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop_device.F90
+++ b/tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop_device.F90
@@ -1,0 +1,63 @@
+!/===--- test_parallel_master_taskloop_device.F90 -------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test checks the parallel master taskloop directive. The test performs
+! simple operations on an int array which are then checked for correctness.
+! This test checks the construct in a target context.
+!
+!/===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_parallel_master_taskloop_device
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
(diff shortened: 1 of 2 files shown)


tests/5.0/parallel_master_taskloop/test_parallel_master_taskloop_device.F90
```diff
@@ -0,0 +1,63 @@
...
+
+    !$omp target map(tofrom: x, y, z, num_threads)
+    !$omp parallel master taskloop num_threads(OMPVV_NUM_THREADS_DEVICE) shared(x, y, z, num_threads) 
+    DO i = 1, N
+       x(i) = x(i) + y(i) * z(i)
+    END DO
+    IF (i .eq. 1) THEN
+       num_threads = omp_get_num_threads()
+    END IF
+    !$omp end parallel master taskloop
```
tob2, line 50: This fails in GCC 12 with:
```
   50 |     !$omp end parallel master taskloop
      |                                      1
Error: Unexpected !$OMP END PARALLEL MASTER TASKLOOP statement at (1)
```

The problem is:
* `taskloop` requires that a _loop-nest_ follows (OpenMP 5.2: "Association: loop").
* With such directives, it is permitted that the associated end-directive can be omitted – thus, it is implicitly added after the `END DO`

I assume the following was intended:
```suggestion
    !$omp parallel master taskloop num_threads(OMPVV_NUM_THREADS_DEVICE) shared(x, y, z, num_threads) 
    DO i = 1, N
       x(i) = x(i) + y(i) * z(i)
       IF (i .eq. 1) THEN
          num_threads = omp_get_num_threads()
       END IF
    END DO
    !$omp end parallel master taskloop
```

seyonglee: Thank you for this nice catch!

spophale (approved):

tmh97 (approved):

PR #480, opened 2022-03-01
Add new_test/test_parallel_master_taskloop_simd_device.F90

seyonglee:
Fails with both XLF 16.1.1-10 and GCC 11.1.0
Update test_parallel_master_taskloop_simd_device.c too.

### DIFF
--- a/tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd_device.F90
+++ b/tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd_device.F90
@@ -1,0 +1,57 @@
+!/===--- test_parallel_master_taskloop_simd_device.F90 -------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test checks the parallel master taskloop simd directive. The test
+! performs simple operations on an int array which are then checked for
+! correctness. This test checks the construct in a target context.
+!
+!/===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_parallel_master_taskloop_simd_device
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
+  OMPVV_TEST_OFFLOADING
+
+  OMPVV_TEST_VERBOSE(parallel_master_taskloop_simd_device() .ne. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+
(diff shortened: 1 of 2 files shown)


tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd_device.F90
```diff
@@ -0,0 +1,57 @@
...
+       x(i) = x(i) + y(i) * z(i)
+    END DO
+    !$omp end parallel master taskloop simd
+    !$omp end target
+
+    DO i = 1, N
+       OMPVV_TEST_AND_SET_VERBOSE(errors, x(i) .ne. 1 + y(i)*z(i))
+    END DO
+
+    OMPVV_INFOMSG("This test does not guarantee thread parallelism of the clause.")
```
spophale, line 53: Maybe say "This thread does not guarantee vector instructions were generated for the simd construct".

tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd_device.F90
```diff
@@ -0,0 +1,57 @@
...
+       y(i) = i + 1
+       z(i) = 2*(i + 1)
+    END DO
+
+    !$omp target map(tofrom: x) map(to: y, z)
+    !$omp parallel master taskloop simd num_threads(OMPVV_NUM_THREADS_DEVICE) shared(x, y, z) 
+    DO i = 1, N
+       x(i) = x(i) + y(i) * z(i)
+    END DO
+    !$omp end parallel master taskloop simd
```
spophale, line 46: Comparing to the other tests, the num_threads check has been removed. Maybe we need to add it back for this and C tests to ensure that there is parallelism.

tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd_device.F90
seyonglee: Updated both C and Fortran versions.

tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd_device.F90
seyonglee: Modified both C and Fortran versions to check the number of threads in a parallel region.

tmh97 (approved):
looks-good-to-go

PR #481, opened 2022-03-01
Add new_test/test_parallel_master_taskloop_simd.F90

seyonglee:
Fails with both XLF 16.1.1-10 and GCC 11.1.0
Update test_parallel_master_taskloop_simd.c too.

### DIFF
--- a/tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd.F90
+++ b/tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd.F90
@@ -27,6 +27,7 @@ CONTAINS
     INTEGER:: errors = 0
     INTEGER, DIMENSION(N):: x, y, z
     INTEGER:: i
+    INTEGER:: num_threads = -1
 
     OMPVV_INFOMSG("test_parallel_master_taskloop_simd")
 
@@ -36,9 +37,12 @@ CONTAINS
        z(i) = 2*(i + 1)
     END DO
 
-    !$omp parallel master taskloop simd num_threads(OMPVV_NUM_THREADS_HOST) shared(x, y, z) 
+    !$omp parallel master taskloop simd num_threads(OMPVV_NUM_THREADS_HOST) shared(x, y, z, num_threads) 
     DO i = 1, N
        x(i) = x(i) + y(i) * z(i)
+       IF (i .eq. 1) THEN
+          num_threads = omp_get_num_threads()
+       ENDIF
     END DO
     !$omp end parallel master taskloop simd
 
@@ -46,7 +50,9 @@ CONTAINS
        OMPVV_TEST_AND_SET_VERBOSE(errors, x(i) .ne. 1 + y(i)*z(i))
     END DO
 
-    OMPVV_INFOMSG("This test does not guarantee thread parallelism of the clause.")
+    OMPVV_WARNING_IF(num_threads .eq. 1, "Test ran with one thread, so parallelism of parallel master with taskloop can't be guaranteed.")
+    OMPVV_ERROR_IF(num_threads .lt. 1, "Test returned an invalid number of threads.")
+    OMPVV_INFOMSG("This test does not guarantee vector instructions were generated for the simd construct.")
 
     parallel_master_taskloop_simd= errors
   END FUNCTION parallel_master_taskloop_simd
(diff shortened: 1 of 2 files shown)


tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd.F90
```diff
@@ -0,0 +1,53 @@
...
+       x(i) = 1
+       y(i) = i + 1
+       z(i) = 2*(i + 1)
+    END DO
+
+    !$omp parallel master taskloop simd num_threads(OMPVV_NUM_THREADS_HOST) shared(x, y, z) 
+    DO i = 1, N
+       x(i) = x(i) + y(i) * z(i)
+    END DO
+    !$omp end parallel master taskloop simd
```
spophale, line 43: IMO adding a num_threads check will be useful.

tob2:
> Fails with both XLF 16.1.1-10 and GCC 11.1.0 Update test_parallel_master_taskloop_simd.c too.

But it compiles with GCC 12 :-) _i.e.current devel version, to be released in few weeks._

Observations just from compiling & running:
* Compiles & runs but on my laptop shows 
  `[OMPVV_RESULT test_parallel_master_taskloop_simd.F90] Test passed on the device.`
  The output should be "host". Seems as if `OMPVV_TEST_OFFLOADING` should be added.
* `Warning: Unused variable ‘ln’ declared`

tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd.F90
seyonglee: Modified both C and Fortran versions to check the number of threads in a parallel region.

tests/5.0/parallel_master_taskloop_simd/test_parallel_master_taskloop_simd.F90
```diff
@@ -0,0 +1,53 @@
...
+
+    DO i = 1, N
+       OMPVV_TEST_AND_SET_VERBOSE(errors, x(i) .ne. 1 + y(i)*z(i))
+    END DO
+
+    OMPVV_INFOMSG("This test does not guarantee thread parallelism of the clause.")
```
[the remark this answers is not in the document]
seyonglee: Changed to "This test does not guarantee vector instructions were generated for the simd construct."

seyonglee:
> * Compiles & runs but on my laptop shows
>   `[OMPVV_RESULT test_parallel_master_taskloop_simd.F90] Test passed on the device.`
>   The output should be "host". Seems as if `OMPVV_TEST_OFFLOADING` should be added.

@tob2 Where did you get this message? I don't see that message when run on Summit using GCC V11. The benchmarks do not have `OMPVV_TEST_OFFLOADING`.

> * `Warning: Unused variable ‘ln’ declared`
>   — turned out to be in the generic module; fixed in pull request [ompvv/ompvv.F90 (clean_fn): Remove unused var #485](https://github.com/SOLLVE/sollve_vv/pull/485).

I don't see this warning message either; maybe, I'm using later version than you, which has already fixed the problem.

seyonglee:
@tob2 It will be good if the latest GCC V12 

> > Fails with both XLF 16.1.1-10 and GCC 11.1.0 Update test_parallel_master_taskloop_simd.c too.
> 
> But it compiles with GCC 12 :-) _i.e.current devel version, to be released in few weeks._

It will be very helpful if GCC 12 can be installed on Summit too, which may increase the number of compilable Fortran tests a lot!

spophale:
LGTM

spophale:
This should be resolved with the makefile change. 
@tmh97 @nolanbaker31 pls review.

tmh97:
> @tob2 It will be good if the latest GCC V12
> 
> > > Fails with both XLF 16.1.1-10 and GCC 11.1.0 Update test_parallel_master_taskloop_simd.c too.
> > 
> > 
> > But it compiles with GCC 12 :-) _i.e.current devel version, to be released in few weeks._
> 
> It will be very helpful if GCC 12 can be installed on Summit too, which may increase the number of compilable Fortran tests a lot!

@seyonglee I requested this from ORNL folks :) They were extremely helpful and added GCC 11.2 (which was not previously available). Once GCC 12 has an official release they said they will also add that!

tmh97 (approved):

Issue #464, opened 2022-02-16
test_target_mapping_before_alloc.F90: Fails to compile due to con't line in preprocessor argument


tob2:
The issue is older, even though I only spotted it now. The **testcase was added via Pull Request #306** and currently has:
https://github.com/SOLLVE/sollve_vv/blob/e51d17b0b69182619be6833685e496a2401a2cbd/tests/5.0/target/test_target_mapping_before_alloc.F90#L52-L53

The problem is the continuation line, which in Fortran uses `&` as end-of-line marker.

However, **after preprocessing,** that line is changed to the following; the problem is the `& &` which appears mid line:
```F90
    errors = errors + test_and_set_verbose(scalar .ne. 80 .OR. a(2) .ne. 2 .OR. test_struct%var .ne. 1 & & .OR. test_struct%b(2) .ne. 2, "condition", "tests/5.0/target/test_target_mapping_before_alloc.F90", 52)
```
That's at least the result with GCC's C preprocessor, including the one internally used by gfortran. That causes **gfortran to fail with**
```
tests/5.0/target/test_target_mapping_before_alloc.F90:52:104:

   52 |     OMPVV_TEST_AND_SET_VERBOSE(errors, scalar .ne. 80 .OR. a(2) .ne. 2 .OR. test_struct%var .ne. 1 &
      |                                                                                                        1
Error: Syntax error in argument list at (1)
```

For gfortran (with `-ffree-line-length-none`), the **solution** is to remove the two `&`:
```diff
--- a/tests/5.0/target/test_target_mapping_before_alloc.F90
+++ b/tests/5.0/target/test_target_mapping_before_alloc.F90
@@ -52,2 +52,2 @@ CONTAINS
-    OMPVV_TEST_AND_SET_VERBOSE(errors, scalar .ne. 80 .OR. a(2) .ne. 2 .OR. test_struct%var .ne. 1 &
-         & .OR. test_struct%b(2) .ne. 2)
+    OMPVV_TEST_AND_SET_VERBOSE(errors, scalar .ne. 80 .OR. a(2) .ne. 2 .OR. test_struct%var .ne. 1 \
+         .OR. test_struct%b(2) .ne. 2)
```

The backslash `\` is only syntactical sugar, making clear that it is a continuation line – handled by the C preprocessor. I think that improves readability – but could also be removed.

[...]

PR #483, opened 2022-03-08
test_declare_target_nested_functions.F90: Avoid I/O in target w/ VERBOSE_MODE

tob2:
* tests/5.0/declare_target/test_declare_target_nested_functions.F90: Don't call
  OMPVV_ERROR_IF in target region; FAIL if offload devices are available but run
  on the host.
***
The problem is that `OMPVV_ERROR_IF` expands to
```
 ... WRITE(ERROR_UNIT, OMPVV_HEADER_FMT(OMPVV_ERROR)) ...
```
and I/O may not be supported in the target region.

Additionally, the message "test fails on the host" is not true – if I run the original program without offloading, I get the following output, i.e. "no errors". (Note: only the last one is shown without VERBOSE_MODE=1.)
```
[OMPVV_INFOMSG test_declare_target_nested_functions.F90:30] Test is running on host
[OMPVV_ERROR test_declare_target_nested_functions.F90:50] Target region executed on host&       & instead of device, test fails on the host
[OMPVV_ERROR test_declare_target_nested_functions.F90:41] Target region executed on host&       & instead of device, test fails on the host
[OMPVV_RESULT test_declare_target_nested_functions.F90] Test passed on the host.
```

Thus, I changed:
* No diagnostic inside the target region
* An error when the default device is not the host – but the function is still executed on the host
* The change also avoids the & ... & issue and the line-break issue, cf. Issue #464.

In the error case, the message is the following (triggered by `call omp_set_default_device(omp_get_initial_device())`):
```
[OMPVV_INFOMSG test_declare_target_nested_functions.F90:31] Test is running on device
[OMPVV_INFOMSG ompvv.F90:249] The value of errors is 2
[OMPVV_RESULT test_declare_target_nested_functions.F90] Test failed on the device.
[OMPVV_ERROR test_declare_target_nested_functions.F90:67]  Condition  on_device .neqv. is_offloading failed
[OMPVV_ERROR test_declare_target_nested_functions.F90:34]  Condition test_declared_functions() .ne. 0 failed
STOP 1
```

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_nested_functions.F90
+++ b/tests/5.0/declare_target/test_declare_target_nested_functions.F90
@@ -22,45 +22,48 @@ PROGRAM test_declare_target_nested_functions
     USE omp_lib
     implicit none
     INTEGER :: errors
+    LOGICAL :: is_offloading
     
     !$omp declare target to(outer_fn)
     
     errors = 0 
      
-    OMPVV_TEST_OFFLOADING
+    OMPVV_TEST_AND_SET_OFFLOADING(is_offloading)
 
     OMPVV_TEST_VERBOSE(test_declared_functions() .ne. 0)
 
     OMPVV_REPORT_AND_RETURN()
 
 CONTAINS  
-    INTEGER FUNCTION inner_fn(a)
+    INTEGER FUNCTION inner_fn(a, on_device)
       INTEGER:: a
-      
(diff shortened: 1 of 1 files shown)


tmh97 (approved):
This is good-to-go. I like these changes, ran on NVHPC to confirm

tmh97:
thanks @tob2 !

nolanbaker31 (approved):
looks good

PR #486, opened 2022-03-09
Add new_test/test_master_taskloop.F90

seyonglee:
Fails with XLF 16.1.1-10, GCC 11.1, and NVHPC 21.11.

### DIFF
--- a/tests/5.0/master_taskloop/test_master_taskloop.F90
+++ b/tests/5.0/master_taskloop/test_master_taskloop.F90
@@ -41,10 +41,10 @@ CONTAINS
     !$omp master taskloop
     DO i = 1, N
        x(i) = x(i) + y(i) * z(i)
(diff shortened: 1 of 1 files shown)


tests/5.0/master_taskloop/test_master_taskloop.F90
```diff
@@ -0,0 +1,61 @@
...
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_master_taskloop
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
```
tob2, line 20: Runs with GCC 12 (no offloading device on the laptop):
`[OMPVV_RESULT test_master_taskloop.F90] Test passed on the device.`

```suggestion

OMPVV_TEST_OFFLOADING
```

spophale (approved):

PR #490, opened 2022-03-09
test_omp_display_env.c: Don't call routine in target

tob2:
Move call to 'omp_display_env' OpenMP API routine out of the
target region (and remove then empty target region).

Reason: Fails in the real world and OpenMP 5.1 states:
"Restrictions to the omp_display_env routine are as follows.
* When called from within a target region the effect is unspecified."

In case of GCC, the routine is supported in mainline since July 2021 (GCC 12; to be released soon) – but only when called on the host. The function is not built for devices – such that linking fails with `undefined symbol: omp_display_env` (unless compiled with `-foffload=disable`).

@spophale @tmh97 @nolanbaker31  @seyonglee – please review.

### DIFF
--- a/tests/5.1/runtime_calls/test_omp_display_env.c
+++ b/tests/5.1/runtime_calls/test_omp_display_env.c
@@ -2,7 +2,7 @@
 //
 // OpenMP API Version 5.1 Aug 2021
 //
-// Tests the behavior of omp_display_env. This testshould print out the OpenMP
+// Tests the behavior of omp_display_env. This test should print out the OpenMP
 // version number and the initial values of the ICV's associated with the
 // environment variables
 //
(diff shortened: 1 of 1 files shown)


tmh97 (approved):

nolanbaker31 (approved):

tests/5.1/runtime_calls/test_omp_display_env.c
```diff
@@ -1,23 +1,20 @@
...
 //
-//OpenMP API Version 5.1 Aug 2021
+// OpenMP API Version 5.1 Aug 2021
 //
-//Tests the behavior of omp_display_env. This test 
-//should print out the OpenMP version number and 
-//the initial values of the ICV's associated with 
-//the environment variables.
-//------------------------------------------------
+// Tests the behavior of omp_display_env. This testshould print out the OpenMP
```
tob2, line 5: Fix typo:
```suggestion
// Tests the behavior of omp_display_env. This test should print out the OpenMP
```

PR #491, opened 2022-03-09
New omp_get_mapped_ptr test

krishols:
Test for memory routine omp_get_mapped_ptr(). Not current supported on Perlmutter llvm.

### DIFF
--- a/tests/5.1/memory_routines/test_get_mapped_ptr.c
+++ b/tests/5.1/memory_routines/test_get_mapped_ptr.c
@@ -1,0 +1,49 @@
+//===--- test_get_mapped_ptr.c -----------------------------------------------===//
+//
+//  OpenMP API Version 5.1 Aug 2021
+//
+//Test for the omp_get_mapped_ptr memory routine. This routine returns the device pointer associated with a host pointer for a given device. 
+//===--------------------------------------------------------------------------------------===//
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+#define N 1024
+
+int errors;
+
+int test_get_mapped_ptr() {
+int * device_ptr;
+int x = 4;
+int num_devices = omp_get_num_devices(); 
+int * arr_ptrs[num_devices];
+OMPVV_INFOMSG_IF(num_devices == 0, "No devices available. ");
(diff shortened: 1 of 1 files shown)


tests/5.1/memory_routines/test_get_mapped_ptr.c
```diff
@@ -0,0 +1,49 @@
...
+
+int test_get_mapped_ptr() {
+int * device_ptr;
+int x = 4;
+int num_devices = omp_get_num_devices(); 
+int * arr_ptrs[num_devices];
+OMPVV_INFOMSG_IF(num_devices == 0, "No devices available. ");
+// test to make sure it is NULL if no devices exist
+if (num_devices == 0) {
+	device_ptr = omp_get_mapped_ptr(&x, 1);
```
tob2, line 25: Using `device_num = 1` which is above `num_device == 0` is invalid. It violates (OpenMP 5.2 quote):

"**Constraints on Arguments**
The _device_num_ argument must be a conforming device number."

with

"A conforming device number is either a non-negative integer that is less than or equal to **omp_get_num_devices()** or equal to **omp_initial_device** or **omp_invalid_device.**"

tmh97: Hey @tob2 this is a test for 5.1, it looks like that phrasing you shared is from the 5.2 specification. 

In the 5.1 specification (section 3.8.11) get_mappped_ptr states the follow:
**"Constraints on Arguments**
The device_num argument must be greater than or equal to zero and less than or equal to the result of omp_get_num_devices

nolanbaker31:
Looks good with recommended changes.
(Like Tobias said, I would imagine it would work using 1, since that device would not be available if numdevices=0, but using omp_get_initial_device() would ensure that it is always null since it would be using host device for the ptr)

nolanbaker31 (approved):
Looks good with recommended changes.
(Like Tobias said, I would imagine it would work using 1, since that device would not be available if numdevices=0, but using omp_get_initial_device() would ensure that it is always null since it would be using host device for the ptr)

spophale (approved):

tmh97 (approved):
Passes on latest llvm dev build on summit. Thanks @tob2 and @marvol77 for your help with this test.

PR #492, opened 2022-03-11
Update ompvv.F90

spophale:
This edit was suggested by @tob2 to prevent tests from reporting execution on target when target device is absent.

### DIFF
--- a/ompvv/ompvv.F90
+++ b/ompvv/ompvv.F90
@@ -105,7 +105,7 @@ module ompvv_lib
   use iso_fortran_env
   implicit none
 OMPVV_MODULE_REQUIRES_LINE
-    LOGICAL, PRIVATE :: ompvv_isHost
+    LOGICAL, PRIVATE :: ompvv_isHost = .true.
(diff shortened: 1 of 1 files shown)


spophale:
#476

tob2 (approved):
> This edit was suggested by @tob2 to prevent tests from reporting execution on target when target device is absent.

Rather:
* Avoid claiming that execution was done _on the device_ for files without the macros `OMPVV_TEST_OFFLOADING` / `OMPVV_TEST_AND_SET_OFFLOADING` (which call test_and_set_offloading())
* The assumption is: all test-case files that contain target regions use those. (At least, it is more likely that they do.)

LGTM.

As mentioned in #476, an alternative would be a tri-state as in the C file, i.e. something like:
https://github.com/SOLLVE/sollve_vv/compare/master...tob2:ompvv.F90_isHost_tristate

nolanbaker31 (approved):
makes sense, looks good

tmh97 (approved):

PR #493, opened 2022-03-14
'target_memcpy_async' test

nolanbaker31:
Test for [`omp_target_memcpy_async`](https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5-1.pdf#page=444). Currently in progress.

I am new to using dependencies within OpenMP, and am not sure how a dependency would work with this construct. The data-attribute within `obj` will denote when memory is copied, but I am not sure specifically when this would occur.

My confusion comes from my understanding of dependencies. I know that when using `#pragma omptask depend`, for example, the task listed under this construct would limit other dependencies, but using _memcpy_ adds confusion to this.

I know I am being a bit vague, but any assistance or clarification would be greatly appreciated.

_Side note,_ this construct is not yet implemented in LLVM, unsure about GCC.

### DIFF
--- a/tests/5.1/target/test_target_memcpy_async.c
+++ b/tests/5.1/target/test_target_memcpy_async.c
@@ -1,0 +1,73 @@
+//===--- test_target_memcpy_async.c -----------------------------------===//
+//
+//  OpenMP API Version 5.1 Aug 2021
+//
+//  Inspired from OpenMP 5.1 Examples Doc, 5.16.4 & 8.9
+//  This test utilizes the omp_target_memcpy_async construct to
+//  allocate memory on the device asynchronously. The construct
+//  uses 'obj' for dependency, so that memory is only copied once
+//  the x variable is changed
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+#define N 1024
+
+int errors, i;
+
+int test_target_memcpy_async() {
+
+    int h, t, i, x;
+    errors = 0;
+    double *mem;
+    double *mem_dev_cpy;
+    *mem = 10; 
+    h = omp_get_initial_device();
+    t = omp_get_default_device();
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() < 1 || def_dev < 0);
+
+    mem_dev_cpy = (double *)omp_target_alloc( sizeof(double)*N, def_dev);
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, mem_dev_cpy == NULL);
+
+    omp_depend_t obj;
+
+    #pragma omp depobj(obj) depend(inout: x)
+
+    /* dst src */
+    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
+                                0,          0,
+                                t,          h,
+                                1,          *obj);
+
+    #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
+    #pragma omp teams distribute parallel for
+    for(i = 0;i<N;i++){
+        mem_dev_cpy[i] = cos((double)i); // initialize data
+    }
+    /* dst src */
+    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
+                                0,          0,
+                                h,          t,
+                                1,          *obj);
+
+    for(int i=0;i<N;i++){
+        OMPVV_TEST_AND_SET(errors, mem_dev_cpy[i]!=cos((double)i));
+    }
+    omp_target_free(mem_dev_cpy, def_dev);
+    #pragma omp depobj(obj) destroy // frees resource
+    return errors;
+}
+
+int main() {
+   errors = 0;
+   OMPVV_TEST_OFFLOADING;
+   OMPVV_TEST_AND_SET_VERBOSE(errors, test_target_memcpy_async() != 0);
+   OMPVV_REPORT_AND_RETURN(errors);
+}           
--- a/tests/5.1/target/test_target_memcpy_async.c
+++ b/tests/5.1/target/test_target_memcpy_async.c
@@ -53,7 +53,7 @@ int test_target_memcpy_async() {
         mem_dev_cpy[i] = cos((double)i); // initialize data
     }
     /* dst src */
-    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
+    omp_target_memcpy_async(mem, mem_dev_cpy, sizeof(double)*N,
                                 0,          0,
                                 h,          t,
                                 1,          *obj);
--- a/tests/5.1/target/test_target_memcpy_async.c
+++ b/tests/5.1/target/test_target_memcpy_async.c
@@ -46,7 +46,7 @@ int test_target_memcpy_async() {
                                 t,          h,
                                 1,          *obj);
 
-    #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
+    #pragma omp target is_device_ptr(mem_dev_cpy) device(t) depend(depobj: *obj)
     #pragma omp teams distribute parallel for
     for(i = 0;i<N;i++){
         mem_dev_cpy[i] = cos((double)i); // initialize data
--- a/tests/5.1/target/test_target_memcpy_async.c
+++ b/tests/5.1/target/test_target_memcpy_async.c
@@ -33,7 +33,7 @@ int test_target_memcpy_async() {
     OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() < 1 || t < 0);
 
     mem = (double *)malloc( sizeof(double)*N);
-    mem_dev_cpy = (double *)omp_target_alloc( sizeof(double)*N, def_dev);
+    mem_dev_cpy = (double *)omp_target_alloc( sizeof(double)*N, t);
 
     OMPVV_TEST_AND_SET_VERBOSE(errors, mem_dev_cpy == NULL);
 
--- a/tests/5.1/target/test_target_memcpy_async.c
+++ b/tests/5.1/target/test_target_memcpy_async.c
@@ -26,7 +26,6 @@ int test_target_memcpy_async() {
     errors = 0;
     double *mem;
     double *mem_dev_cpy;
-    *mem = 10; 
     h = omp_get_initial_device();
     t = omp_get_default_device();
 
--- a/tests/5.1/target/test_target_memcpy_async.c
+++ b/tests/5.1/target/test_target_memcpy_async.c
@@ -30,8 +30,9 @@ int test_target_memcpy_async() {
     h = omp_get_initial_device();
     t = omp_get_default_device();
 
-    OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() < 1 || def_dev < 0);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() < 1 || t < 0);
 
+    mem = (double *)malloc( sizeof(double)*N);
     mem_dev_cpy = (double *)omp_target_alloc( sizeof(double)*N, def_dev);
 
     OMPVV_TEST_AND_SET_VERBOSE(errors, mem_dev_cpy == NULL);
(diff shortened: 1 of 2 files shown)


tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+
+int errors, i;
+
+int test_target_memcpy_async() {
+
+    int h, t, i, x;
+    errors = 0;
+    double *mem;
+    double *mem_dev_cpy;
+    *mem = 10;
```
tob2, line 29:
```suggestion
```

tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+    int h, t, i, x;
+    errors = 0;
+    double *mem;
+    double *mem_dev_cpy;
+    *mem = 10; 
+    h = omp_get_initial_device();
+    t = omp_get_default_device();
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() < 1 || def_dev < 0);
+
```
tob2, line 34:
```suggestion
    OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() < 1 || t < 0);

    mem = (double *)malloc( sizeof(double)*N);
```

tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+    mem_dev_cpy = (double *)omp_target_alloc( sizeof(double)*N, def_dev);
```
tob2, line 35:
```suggestion
    mem_dev_cpy = (double *)omp_target_alloc( sizeof(double)*N, t);
```

tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+                                t,          h,
+                                1,          *obj);
+
+    #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
+    #pragma omp teams distribute parallel for
+    for(i = 0;i<N;i++){
+        mem_dev_cpy[i] = cos((double)i); // initialize data
+    }
+    /* dst src */
+    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
```
tob2, line 55:
```suggestion
    omp_target_memcpy_async(mem, mem_dev_cpy, sizeof(double)*N,
```

tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+                                0,          0,
+                                h,          t,
+                                1,          *obj);
+
+    for(int i=0;i<N;i++){
+        OMPVV_TEST_AND_SET(errors, mem_dev_cpy[i]!=cos((double)i));
+    }
+    omp_target_free(mem_dev_cpy, def_dev);
```
tob2, line 63: The first line of the suggestion has been superseded by the comment/suggestion above. But the `def_dev` → `t` is still needed.

```suggestion
        OMPVV_TEST_AND_SET(errors, mem[i]!=cos((double)i));
    }
    omp_target_free(mem_dev_cpy, t);
```

tob2 (commented):
I found some issues when compiling with GCC.

Note: I only compiled with test_target_memcpy – as the async version is not yet in GCC 12. (A patch is available but it missed the feature-freeze deadline for GCC 12.)

tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+        mem_dev_cpy[i] = cos((double)i); // initialize data
+    }
+    /* dst src */
+    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
+                                0,          0,
+                                h,          t,
+                                1,          *obj);
+
+    for(int i=0;i<N;i++){
+        OMPVV_TEST_AND_SET(errors, mem_dev_cpy[i]!=cos((double)i));
```
tob2, line 61: Here, a comment got lost. The check for equality does not work well with floating-point numbers. Cf.

* [Goldberg paper](http://www.validlab.com/goldberg/paper.pdf), [[2](http://www.google.de/search?q=Goldberg+What+Every+Computer+Scientist+Should+Know+About+Floating-Point+Arithmetic)]
* [Monniaux paper](http://hal.archives-ouvertes.fr/hal-00128124)
* [Semantics of Floating Point Math in GCC](https://gcc.gnu.org/wiki/FloatingPointMath)

_The following should work – but one can argue about the best value for UPL/unit of last precision. Probably a smaller value is fine – and with very high optimizations, it won't work. For instance, when the compiler replaces the library call by AMD GCN (Radeon)'s hardware intrinsics (those accept double but use only float precision)._

```suggestion
        double x = mem[i];
        double y = cos((double)i);
        int ulp = 10;
        OMPVV_TEST_AND_SET(errors, fabs (x - y) > ulp*DBL_EPSILON*fmax(fabs(x), fabs(y)));
```

tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+    for(i = 0;i<N;i++){
+        mem_dev_cpy[i] = cos((double)i); // initialize data
+    }
+    /* dst src */
+    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
+                                0,          0,
+                                h,          t,
+                                1,          *obj);
+
+    for(int i=0;i<N;i++){
```
tob2, line 60: ... and likewise here: if memcpy is asynchronous, you need a barrier before checking the value.

Actually, I fail to see how to create a simple barrier. Thus, I filled the OpenMP language-spec issue 3248 for it.

_`depend` can be used with the directives: depobj, interop, target, target enter data, target exit data, target
update, task, taskwait — but none of those really fit here._

_Disclaimer: Even though I tried hard, I might have missed some OpenMP feature in the spec which can be used here._

tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+
+    omp_depend_t obj;
+
+    #pragma omp depobj(obj) depend(inout: x)
+
+    /* dst src */
+    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
+                                0,          0,
+                                t,          h,
+                                1,          *obj);
```
nolanbaker31, line 47: `*obj` might have to be `depobj: *obj`

tests/5.1/target/test_target_memcpy_async.c
tob2: Actually, I forgot about the the _initial task_. Thus, it should be possible to just use:

* `#pragma omp taskwait depend(depobj: obj)`

tests/5.1/target/test_target_memcpy_async.c
marvol77: I think "obj" should be an array of omp_depend_t (instead of a single object) according to the spec:
"The dependences are specified by passing the number of omp_depend_t objects followed by an array of
omp_depend_t objects."
Then you just pass "obj".

tests/5.1/target/test_target_memcpy_async.c
marvol77: As omp_target_memcpy_async is handled like a new created task:
"The omp_target_memcpy_async routine executes as if part of a target task that is generated by the call to the routine and for which execution may be deferred." (OpenMP 5.1 spec)
... I would suggest to use 
`#pragma omp taskwait`
before filling/testing the corresponding device memory.

tests/5.1/target/test_target_memcpy_async.c
marvol77: As above: either define "obj" as array and pass "obj", or just pass "&obj".

tests/5.1/target/test_target_memcpy_async.c
```diff
@@ -0,0 +1,73 @@
...
+
+    #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
```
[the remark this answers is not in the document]
marvol77: I agree - otherwise it's not considered as dependence for omp_target_memcpy_async :)
Though I think it's
`depend(depobj: obj)`
"List items used in depend clauses with the depobj dependence type must be depend objects in the initialized state."
(In the OpenMP 5.1 examples document, p. 321, "obj" is defined as `omp_depend_t *obj` and thus used with `omp_depend_t *obj`. But in your test case "obj" is not a pointer.

tests/5.1/target/test_target_memcpy_async.c
nolanbaker31: seems correct, changing to &obj

tests/5.1/target/test_target_memcpy_async_depobj.c
```diff
@@ -0,0 +1,77 @@
...
+                                0,          0,
+                                t,          h,
+                                1,          &obj);
+
+    #pragma omp taskwait depend(depobj: obj)
+    #pragma omp target is_device_ptr(mem_dev_cpy) map(tofrom: x) device(t) depend(depobj: *obj)
+    #pragma omp teams distribute parallel for
+    for(i = 0;i<N;i++){
+        mem_dev_cpy[i] = i*2; // initialize data
+        x = 5;
```
spophale, line 54: race on "x". Let's keep the testing to arrays.

spophale, line 54: race on "x". Let's keep the testing to arrays.

tests/5.1/target/test_target_memcpy_async_depobj.c
```diff
@@ -0,0 +1,77 @@
...
+    }
+    OMPVV_TEST_AND_SET(errors, x!=5);
```
spophale, line 56: remove

spophale, line 56: remove

tests/5.1/target/test_target_memcpy_async_depobj.c
```diff
@@ -0,0 +1,79 @@
...
+
+    mem = (double *)malloc( sizeof(double)*N);
+    mem_dev_cpy = (double *)omp_target_alloc( sizeof(double)*N, t);
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, mem_dev_cpy == NULL);
+    
+    for(i = 0; i < N; i++){
+        mem[i] = i;
+    }
+    omp_depend_t obj;
```
tmh97:
```suggestion
    omp_depend_t obj[1];
```

nolanbaker31:
no syntax errors on clang 15.0.0-20220504, does not pass as memcpy async is not implemented

tests/5.1/target/test_target_memcpy_async_depobj.c
```diff
@@ -0,0 +1,76 @@
...
+    #pragma omp target is_device_ptr(mem_dev_cpy) device(t) depend(depobj: obj)
+    for(i = 0; i < N; i++){
+        mem_dev_cpy[i] = mem_dev_cpy[i]*2; // initialize data
+    }
+
+    /* copy to host memory */
+    omp_target_memcpy_async(mem, mem_dev_cpy, sizeof(double)*N,
+                                0,          0,
+                                h,          t,
+                                1,          obj_arr[0]);
```
tob2, line 59: Fails here - and below with the message:
```
test_target_memcpy_async_depobj.c: In function ‘test_target_memcpy_async_depobj’:
test_target_memcpy_async_depobj.c:47:52: error: incompatible type for argument 9 of ‘omp_target_memcpy_async’
   47 |                                 1,          obj_arr[0]);
      |                                             ~~~~~~~^~~
      |                                                    |
      |                                                    omp_depend_t
In file included from test_target_memcpy_async_depobj.c:11:
....lib/gcc/x86_64-pc-linux-gnu/13.0.0/include/omp.h:277:42: note: expected ‘omp_depend_t *’ but argument is of type ‘omp_depend_t’
  277 |                                     int, omp_depend_t *)
```
Either of the following should work and should be identical under the hood:
```suggestion
                                1,          &obj_arr[0]);
```
or
```suggestion
                                1,          obj_arr);
```
_Side remark: Since today, GCC mainline (alias GCC 13, to be released in a year) supports `omp_target_memcpy_async` and `omp_target_memcpy_rect_async`_

PR #497, opened 2022-03-17
omp_target_memcpy_async TEST with no dependent object

nolanbaker31:
Seperate test for omp_memcpy_async with 'depobj' of 0, where the [spec](https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5-1.pdf#page=445) states:

> the generated target task is not a dependent task if the program passes in a count of zero for depobj_count. depojb_list is ignored if the value of depobj_count is zero.

Test passes on latest LLVM clang using omp_target_memcpy, since omp_target_memcpy_async is not yet supported.

I am going to leave up PR #493 for now, am waiting for further feedback on a use case/strategy for using dependency with this directive.

### DIFF
--- a/tests/5.1/target/test_target_memcpy_async_no_obj.c
+++ b/tests/5.1/target/test_target_memcpy_async_no_obj.c
@@ -1,0 +1,69 @@
+//===--- test_target_memcpy_async_no_obj.c ----------------------------===//
+//
+//  OpenMP API Version 5.1 Aug 2021
+//
+//  Inspired from OpenMP 5.1 Examples Doc, 5.16.4 & 8.9
+//  This test utilizes the omp_target_memcpy_async construct to
+//  allocate memory on the device asynchronously. The construct
+//  uses 0 for 'depobj_count', so that the clauses is not dependent
+//  and memory is therefore copied synchronously.
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+#define N 1024
+
+int errors, i;
+
+int test_target_memcpy_async_no_obj() {
+
+    int h, t, i;
+    errors = 0;
+    double *mem;
+    double *mem_dev_cpy;
+    h = omp_get_initial_device();
+    t = omp_get_default_device();
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() < 1 || t < 0);
+
+    //allocate space on host & target
+    mem = (double *)malloc( sizeof(double)*N);
+    mem_dev_cpy = (double *)omp_target_alloc( sizeof(double)*N, t);
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, mem_dev_cpy == NULL);
+
+    /* copy to target */
+    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
+                                0,          0,
+                                t,          h,
+                                0,          0);
+
+    #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
+    #pragma omp teams distribute parallel for
+    for(i = 0;i<N;i++){
+        mem_dev_cpy[i] = i*2; // initialize data on device
+    }
+    /* copy to host */
+    omp_target_memcpy_async(mem, mem_dev_cpy, sizeof(double)*N,
+                                0,          0,
+                                h,          t,
+                                0,          0);
+
+    for(int i=0;i<N;i++){
+        OMPVV_TEST_AND_SET(errors, mem[i]!=i*2);
+    }
+    omp_target_free(mem_dev_cpy, t);
+    return errors;
+}
+
+int main() {
+   errors = 0;
+   OMPVV_TEST_OFFLOADING;
+   OMPVV_TEST_AND_SET_VERBOSE(errors, test_target_memcpy_async_no_obj() != 0);
+   OMPVV_REPORT_AND_RETURN(errors);
+}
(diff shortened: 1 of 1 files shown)


tests/5.1/target/test_target_memcpy_async_no_obj.c
```diff
@@ -0,0 +1,69 @@
...
+    #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
+    #pragma omp teams distribute parallel for
+    for(i = 0;i<N;i++){
+        mem_dev_cpy[i] = i*2; // initialize data on device
+    }
+    /* copy to host */
+    omp_target_memcpy_async(mem, mem_dev_cpy, sizeof(double)*N,
+                                0,          0,
+                                h,          t,
+                                0,          0);
```
tob2, line 55:
```suggestion
                                0,          NULL);
```
makes reading the code easier.

However, my main problem is that this code assumes: _No depend object → code runs synchronously._

But my reading of the spec is that it just runs asynchronously and the only synchronization barrier is then the end of the current task.

OpenMP 5.2 spec has:
> **Binding** – The binding task set for an omp_target_memcpy_async region is the generating task, which is the _target task_ generated by the call to the omp_target_memcpy_async routine.
>
> **Effect** – [...] The **omp_target_memcpy_async** routine executes as if part of a target task that is generated by the call to the routine and for which execution may be deferred. Task dependences are expressed with zero or more OpenMP depend objects. The dependences are specified by passing the number of depend objects followed by an array of the objects. The generated target task is not a dependent task if the program passes in a count of zero for _depobj_count._ _depobj_list_ is ignored if the value of _depobj_count_ is zero.

Reading/thinking about it, I think something like the following would work; to be inserted after both `omp_target_memcpy_async`
* `omp taskwait`  (I think this one is better)
* `omp barrier`

nolanbaker31: I am still struggling to understand tasks in OMP.... what would the 'current task' be in this case.. is it explicit? @tob2 

Anyways, I will add `omp taskwait` as it can't hurt

tob2: > what would the 'current task' be in this case

I think the following applies (emphasis is mine):

"An OpenMP program begins as a single thread of execution, called an initial thread. An initial thread executes sequentially, as if the code encountered is part of an **implicit task region,** called an **initial task region,** that is **generated by** the **implicit parallel region surrounding the whole program.**" (From OpenMP 5.2, 1.3 Execution Model)

spophale (approved):

tests/5.1/target/test_target_memcpy_async_no_obj.c
```diff
@@ -0,0 +1,71 @@
...
+
+    /* copy to target */
+    omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
+                                0,          0,
+                                t,          h,
+                                0,          NULL);
+
+    #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
+    #pragma omp teams distribute parallel for
+    for(i = 0;i<N;i++){
```
tmh97, line 48:
```suggestion
    for(i = 0; i < N; i++){
```

tmh97, line 48:
```suggestion
    for(i = 0; i < N; i++){
```

tests/5.1/target/test_target_memcpy_async_no_obj.c
```diff
@@ -0,0 +1,71 @@
...
+    }
+    /* copy to host */
+    omp_target_memcpy_async(mem, mem_dev_cpy, sizeof(double)*N,
+                                0,          0,
+                                h,          t,
+                                0,          NULL);
+
+    #pragma omp taskwait
+    
+    for(int i=0;i<N;i++){
```
tmh97, line 59:
```suggestion
    for(int i = 0; i < N; i++){
```

tmh97, line 59:
```suggestion
    for(int i = 0; i < N; i++){
```

tests/5.1/target/test_target_memcpy_async_no_obj.c
```diff
@@ -0,0 +1,71 @@
...
+        OMPVV_TEST_AND_SET(errors, mem[i]!=i*2);
```
tmh97, line 60:
```suggestion
        OMPVV_TEST_AND_SET(errors, mem[i] != i*2);
```

```

Comment (tmh97) [inline tests/5.1/target/test_target_memcpy_async_no_obj.c:60]:
```suggestion
        OMPVV_TEST_AND_SET(errors, mem[i] != i*2);
```

```

File: tests/5.1/target/test_target_memcpy_async_no_obj.c [hunk 5]
```diff
@@ -0,0 +1,71 @@
...
+//
+//  Inspired from OpenMP 5.1 Examples Doc, 5.16.4 & 8.9
+//  This test utilizes the omp_target_memcpy_async construct to
+//  allocate memory on the device asynchronously. The construct
+//  uses 0 for 'depobj_count', so that the clause is not dependent
+//  and memory is therefore copied synchronously.
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
```
tmh97, line 13:
```suggestion
#include <omp.hdsgdgsdfgsdgsd>
```

tmh97, line 13:
```suggestion
#include <omp.hdsgdgsdfgsdgsd>
```

tmh97 (approved):
looks good!

PR #499, opened 2022-03-17
Silences warnings, fix bugs found that way

tob2:
This patch silences:
* signed/unsigned comparisons (related to int / size_t)
* unused variables (some, there are more)
* uninitialized variable (implicitly mapped 'ptr'; added '= NULL')
* unused but set variable.
* 'missing braces around initializer' (gcc warned, g++ didn't)

Bug fixes:
* tests/5.0/simd/test_simd_nontemporal.c: Add missing 'omp' to pragma.
* tests/5.1/default/test_task_target_default_firstprivate.c: Likewise;
  don't map numerical constant 'N'.
* tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c:
  Use 'i += 2' instead of no op '2+i' as loop step.

The "OMPVV_REPORT_AND_RETURN(error_sum != 0.0);" appears as with VERBOSE_MODE, that's passed to '%d' and error_sum is double. (It also converts the value explicitly to an int for the return value.)

To do / follow up tasks:
* There are some more warnings, especially about uninitialized vars.
Additionally on the Fortran side:
* I see `BYTE` (vendor extension, should be `integer(kind=int8)`, where `int8` comes from `iso_fortran_env`)
* Likewise 2/4/8 should be replaced by int... or real...
* I see also some warnings regarding constants, where _{int,real}... should be appended).
* I also see some warnings that REAL variables are compared for equality ("a == b" instead of "abs(a-b) < eps"), which might be okay for small integer values.


nolanbaker31:
Might be good to throw up an issue for the 'follow up tasks'

nolanbaker31 (approved):
thanks for the cleanup, im surprised how many tests had unused variables !

tob2:
> Might be good to throw up an issue for the 'follow up tasks', just for documentation in case the pr gets merged

I have now created Pull Request #502 for the BYTE/integer-kind/equal comparison issue.

_There are still -Wcompare-reals, -Wunused-parameter, -Wunused-but-set-variable, etc. warnings, but those can be handled later._

spophale (approved):
Thanks @tob2 !

PR #500, opened 2022-03-17
Fix test_target_teams_distribute_parallel_for_devices.F90

tob2:
* Actually do something useful for the first device (device number: 0)
* Start with checking in reverse device-number order in order to not override the host value before having checked it.
* Add comments that this testcase assumes that 'map' is no no-op.

_The loop-walking change makes use of the knowledge that `omp_get_num_devices()` is the device number for the host/initial device. (In newer specs:_ a number _as another one is `omp_initial_device == -1`.)_

@tmh97 @nolanbaker31 @spophale – please review.

### DIFF
--- a/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_devices.F90
+++ b/tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_devices.F90
@@ -28,7 +28,7 @@ CONTAINS
    INTEGER FUNCTION target_teams_distribute_parallel_for_devices()
       INTEGER :: num_dev, errors, i, dev
       INTEGER, DIMENSION(N) :: a
-      LOGICAL, DIMENSION(N) :: isHost
+      LOGICAL, DIMENSION(0:N) :: isHost
       CHARACTER(len=400) :: numDeviceMsg, hostOrDevMsg
 
       errors = 0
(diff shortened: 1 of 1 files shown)


seyonglee:
Can we rely on the knowledge that `omp_get_num_devices()` is the device number for the host/initial device? As I know, this requirement is added from OpenMP V5.1, but this test is for OpenMP V4.5, where the value of the device number for the initial device is implementation-defined.

tob2:
> Can we rely on the knowledge that `omp_get_num_devices()` is the device number for the host/initial device?

Thinking about it, I think the simple solution would have been to change `do dev = 1, num_dev` to `dev = 0, num_dev-1` as that seems to be the original idea of the program. I did not shift the index as coming from the stance that num_dev is a valid number, I thought that the host device was used on purpose. – Well, it now it is!

I think most compilers changed rather early their code to have omp_initial_device() == omp_get_num_devices(), thus that should mostly work. — Still, it is better not to assume this. (I changed the code.)


Thanks for pointing out!

seyonglee (approved):
It looks good!

seyonglee:
1) Should the array `isHost` be initialized first?

tob2:
> 1. Should the array `isHost` be initialized first?

I think it does not really matter as all array elements referenced for reading the value are before set inside the device loop. But, of course, it also does not harm.

>  2. Does it actually matter to check the `isHost` array in the reverse order (`num_dev` to `0`) in the last loop?
>        `isHost` array is updated by the host in the last iteration, and thus it will not be overwritten by other devices.

The `isHost` bit is fine – but the original testcase also checks for the value of `a` and that one is supposed to have the value `a(i) == 1 + device_num`. To check the value, `!$omp target exit data map(from: a) device(dev2)` is used. Unless checking the host value first, this will always fail as the host value is overridden.

Side note (and unchanged): The code will also fail if `target enter data map(to: a)` is a no op and the original host _storage location_ is accessed on the devices. _(Likely the case for (unified-)shared memory, unless the memory is still mapped into device memory as permitted by the standard.)_

spophale (approved):

PR #504, opened 2022-03-25
test_requires_dynamic_allocators.F90: Free allocator/var inside target

tob2:
Move omp_destroy_allocator and DEALLOCATE inside the target region
into which they belong as omp_init_allocator and ALLOCATE are in the
target region.


spophale (approved):
LGTM.

nolanbaker31 (approved):

tmh97 (commented):
Good to go, thanks @tob2

PR #383, opened 2021-08-09
Adding test for target_teams_distribute_parallel_for_if with parallel modifier

tmh97:
Passes with GCC 11

Addresses issue #81

PR #508, opened 2022-03-30
Fixing issue where test does not properly check for offloading

tmh97:
Addressing @tob2 's comments on #383 

Fixes:

1. Ensure that the checkPreconditions() function is called. In the previous version, this function was defined but was never called. This is the function that runs OMPVV_TEST_OFFLOADING, and without it, the test returned 'ran on host' 
2. Made sure that checks for get_num_threads are called only when get_thread_num == 0


tob2 (approved):
LGTM.

(Minor aesthetic issue: I think `, "` (comma, space, opening quotation) looks nicer than the `," ` (comma,opening quotation, space) in the last OMPVV_WARNING_IF)

spophale (approved):

nolanbaker31 (approved):

PR #512, opened 2022-04-13
Add new_test/test_loop_reduction_add_device.F90

seyonglee:


tob2:
Compiles & runs successfully with GCC 12 (mainline). [I only tried it with host fallback.]

spophale (approved):

nolanbaker31 (approved):

tmh97 (approved):

PR #513, opened 2022-04-15
Adding amdclang, amdclang++, and amdflang flags to make.def file

tmh97:
Tested a few different .c, .cpp, and .f90 tests on Spock with offloading.


jrreap (approved):
Looks good to go

nolanbaker31 (approved):

spophale (approved):

PR #515, opened 2022-04-19
ompvv/ompvv.F90: Guard a PARAMETER by #ifdef

tob2:
Silences "Unused parameter 'msg' declared" with unset VERBOSE_MODE in this always-included  file.

In gfortran, this warning only triggers with `-Wunused-parameter` (enabled by `-Wextra`) – if (and only if) VERBOSE_MODE is undefined. — However, since it is localized in ompvv/ompvv.F90, which is included in all Fortran test cases and is the only file where VERBOSE_MODE is processed, I think it still is sensible to silence this warning.


nolanbaker31 (approved):

spophale:
Thanks @tob2 !

PR #517, opened 2022-04-22
Add new_test/test_loop_reduction_subtract_device.F90

seyonglee:
Fails with GCC/11.1.0, XL/16.1.1-10, NVHPC/21.11


tob2:
Works with GCC 12 on my laptop (with host fallback):
```
[OMPVV_INFOMSG test_loop_reduction_subtract_device.F90:21] Test is running on host
[OMPVV_RESULT test_loop_reduction_subtract_device.F90] Test passed on the host.
```

spophale (approved):

nolanbaker31 (approved):

PR #519, opened 2022-04-22
Add new_test/test_loop_reduction_multiply.F90

seyonglee:

### DIFF
--- a/tests/5.0/loop/test_loop_reduction_multiply.F90
+++ b/tests/5.0/loop/test_loop_reduction_multiply.F90
@@ -60,7 +60,7 @@ CONTAINS
       END DO
       !$omp end loop
       !$omp do
-      DO y = 1, 16
+      DO y = 0, 15
          num_threads(x + y) = omp_get_num_threads()
       END DO
       !$omp end do
--- a/tests/5.0/loop/test_loop_reduction_multiply.F90
+++ b/tests/5.0/loop/test_loop_reduction_multiply.F90
@@ -1,0 +1,85 @@
+!===--- test_loop_reduction_multiply.F90 -------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test uses the reduction clause on a loop directive, testing that the
+! variable in the reduction clause is properly reduced using the multiply
+! operator.
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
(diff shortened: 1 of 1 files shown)


tests/5.0/loop/test_loop_reduction_multiply.F90
```diff
@@ -0,0 +1,85 @@
...
+      END DO
+      !$omp end loop
+      !$omp do
+      DO y = 1, 16
+         num_threads(x + y) = omp_get_num_threads()
+      END DO
+      !$omp end do
+      !$omp end parallel
+      host_result = 1
+      DO y = 1, 16
```
tob2, line 69:
```suggestion
      DO y = 0, 15
```

tests/5.0/loop/test_loop_reduction_multiply.F90
```diff
@@ -0,0 +1,85 @@
...
+    DO x = 1, N, 16
+      device_result = 1
+      !$omp parallel num_threads(OMPVV_NUM_THREADS_HOST)
+      !$omp loop reduction(*:device_result)
+      DO y = 1, 16
+         device_result = device_result * a(x + y)
+      END DO
+      !$omp end loop
+      !$omp do
+      DO y = 1, 16
```
tob2, line 63:
```suggestion
      DO y = 0, 15
```

tests/5.0/loop/test_loop_reduction_multiply.F90
```diff
@@ -0,0 +1,85 @@
...
+       num_threads(x) = -1 * x
+    END DO
+
+    device_result = 1
+
+    DO x = 1, N, 16
+      device_result = 1
+      !$omp parallel num_threads(OMPVV_NUM_THREADS_HOST)
+      !$omp loop reduction(*:device_result)
+      DO y = 1, 16
```
tob2, line 58: Works with GCC 12 after applying the `y` range fixes. With host fallback, the result is then:
```
[OMPVV_RESULT test_loop_reduction_multiply.F90] Test passed on the host.
```

```suggestion
      DO y = 0, 15
```

seyonglee: @tob2 Thank you for the nice catch!

spophale (approved):

nolanbaker31 (approved):

PR #521, opened 2022-04-25
New test/test parallel for order reproducible

jrreap:
Initial test to check if the new order(reproducible) clause works for a parallel for construct.

### DIFF
--- a/tests/5.1/order/test_parallel_for_order_reproducible.c
+++ b/tests/5.1/order/test_parallel_for_order_reproducible.c
@@ -22,24 +22,39 @@
 int main() {
 	int errors = 0;
 	int arr[N];
+	int correct[N];
+	int shared = 0;
 
 	for (int i = 0; i < N; i++) {
-		arr[i] = i;
+		arr[i] = 0;
   	}
+
+	for (int i = 0; i < N; i++) {
+		correct[i] = shared;
+		shared = i;
+		printf("Correct %d: %d, Shared: %d\n", i, correct[i], shared);
+	}
+
+	shared = 0;
 
 	OMPVV_TEST_OFFLOADING;
 
-	#pragma omp target map(tofrom:arr,errors)
+	#pragma omp target map(tofrom:arr,errors,correct,shared)
 	{
 		#pragma parallel for thread_limit(OMPVV_NUM_THREADS_DEVICE) //order(reproducible)
-		for (int i = 0; i < N; i++) {
+		{
+			for (int i = 0; i < N; i++) {
+				arr[i] = shared;
+				shared = i;	
+			}
 			
+			for (int i = 0; i < N; i++) {
+				OMPVV_TEST_AND_SET(errors, arr[i] != correct[i]);
+			}
+
 		}
 	
 	}
-
-	OMPVV_WARNING_IF(1, "The number of teams was unexpected, the test results are likely inconcuslive")
-	OMPVV_TEST_AND_SET(errors, 1 == 2);
 
 	OMPVV_REPORT_AND_RETURN(errors);
 }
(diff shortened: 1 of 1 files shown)


tests/5.1/order/test_parallel_for_order_reproducible.c
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	OMPVV_TEST_OFFLOADING;
+
+	#pragma omp target map(tofrom: x,y)
+	{
+
+		#pragma omp parallel for order(concurrent)
+		for (int i = 0; i < N; i++) {
+			x[i] = omp_get_thread_num();
```
tob2, line 36: This fails with GCC 12 here and below with:
```
test_parallel_for_order_reproducible.c:36:32: error: OpenMP runtime API call ‘omp_get_thread_num’ in a region with ‘order(concurrent)’ clause
   36 |                         x[i] = omp_get_thread_num();
      |                                ^~~~~~~~~~~~~~~~~~~~
```

tob2: Namely:
>  Restrictions Restrictions to the **order** clause are as follows:
> ...
> *  A region that corresponds to a construct with an **order** clause that specifies **concurrent** may not contain calls to procedures that contain OpenMP directives.

→ https://www.openmp.org/spec-html/5.1/openmpsu47.html#x72-720002.11.3

tests/5.1/order/test_parallel_for_order_reproducible.c
```diff
@@ -0,0 +1,51 @@
...
+		x[i] = i;
+		y[i] = i;
+	}
+
+	OMPVV_TEST_OFFLOADING;
+
+	#pragma omp target map(tofrom: x,y)
+	{
+
+		#pragma omp parallel for order(reproducible)
```
tob2, line 34: Fails with GCC 12 with:
```
test_parallel_for_order_reproducible.c:34:48: error: expected ‘concurrent’ before ‘reproducible’
   34 |                 #pragma omp parallel for order(reproducible)
```
Cf. https://www.openmp.org/spec-html/5.1/openmpsu47.html#x72-720002.11.3

But it works with GCC 12 after applying the change
```suggestion
		#pragma omp parallel for order(reproducible:concurrent)
```
and ...

tests/5.1/order/test_parallel_for_order_reproducible.c
```diff
@@ -0,0 +1,51 @@
...
+		for (int i = 1; i < N; i++) {
+			x[i] = x[i-1] + x[i];	
+		}
+
+		#pragma omp parallel for order(reproducible)
```
tob2, line 39:
```suggestion
		#pragma omp parallel for order(reproducible:concurrent)
```

tests/5.1/order/test_parallel_for_order_reproducible.c
```diff
@@ -0,0 +1,51 @@
+//===------------ test_target_parallel_for_order_reproducible.c -------------===//
+//
+// OpenMP API Version 5.1 Nov 2020
+//
+// This test checks that the order(reproducible) clause is properly handled.
```
tob2, line 5:
```suggestion
// This test checks that the order(reproducible:concurrent) clause is properly handled.
```

tests/5.1/order/test_parallel_for_order_reproducible.c
jrreap: Good catch! I was just looking at the spec and realized I had forgotten that. Also glad to see it's working with GCC 12

spophale:
The test looks good. It is structured differently than most of our recent tests (main calls test_fun). NOT a show stopper.

spophale:
See #544

spophale (approved):

PR #527, opened 2022-05-01
Testing OMP_NUM_TEAMS and OMP_TEAMS_THREAD_LIMIT env vars

mjcarr458:


tob2:
I think this testcase just checks that `setenv` and `getenv` work – which is POSIX but not OpenMP. I think you want to set the environment variable before the program is started and then use an OpenMP API routine to check the values.

Have a look at, e.g. https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.0/program_control/test_omp_target_offload_env_DEFAULT.c – that testcase checks that the OMP_TARGET_OFFLOAD environment variable is correctly handled, which is similar.

Crucial for the example above is the file name: The run script causes that for a test-case file name of `test_omp_foo_env_bar.suffix` the environment variable `OMP_FOO` is set to the value `bar` before running that testcase.

(That ENV support is in https://github.com/SOLLVE/sollve_vv/blob/master/sys/scripts/run_test.sh – This has to be done before the program is started as at that point the ICV is setup; therefore, it is in a run script and not done via setenv in the code.)

* The `OMP_TEAMS_THREAD_LIMIT` env var initializes the _teams-thread-limit-var_ ICV, which can then be queried using `omp_get_teams_thread_limit`
* The `OMP_NUM_TEAMS` env var is used to initialize the_nteams-var_ ICV, which can then be queried via `omp_get_max_teams`

_It could also be checked by using a **teams** construct (instead or – better – in addition) but then the fineprint has to be taken into account as the implementation can be valid also when not using the specified number of those ICVs. (Well, that's already implied by the names 'max' and 'limit'.)_

PR #528, opened 2022-05-04
New test/test loop reduction and device.f90

seyonglee:


tob2:
Works with GCC 12 – to be released on Friday (in 2 days).

Side remark: I get with -Wall for line 40: "Warning: Unused variable 'randominteger' declared"

tmh97 (approved):
This tests looks logically sound. Removing loop directive and running with GCC 12 produces correct results.

PR #531, opened 2022-05-04
Add new_test/test_loop_reduction_or.F90

seyonglee:


tob2:
Works with GCC 12 – and shows with -Wall for line 40: "Warning: Unused variable 'randominteger' declared"

tmh97 (approved):
Good - to -go.

spophale (approved):

PR #533, opened 2022-05-10
Add new_test/test_loop_reduction_bitand.F90

seyonglee:
Add explicit initialization statement to both C and Fortran versions.


tests/5.0/loop/test_loop_reduction_bitand.F90
```diff
@@ -0,0 +1,111 @@
...
+       b = b + ISHFT(1, x)
+    END DO
+
+    !$omp parallel num_threads(OMPVV_NUM_THREADS_HOST)
+    !$omp loop reduction(iand:b)
+    DO x = 1, N
+       b = IAND(b, a(x))
+    END DO
+    !$omp end loop
+    !$omp do
```
spophale, line 87: needs omp *parallel* do

spophale (approved):

PR #536, opened 2022-05-10
Add new_test/new_test/test_loop_reduction_bitor_device.F90

seyonglee:
Add explicit initialization statement to both C and Fortran versions.

### DIFF
--- a/tests/5.0/loop/test_loop_reduction_bitor_device.c
+++ b/tests/5.0/loop/test_loop_reduction_bitor_device.c
@@ -31,6 +31,9 @@ int test_bitor() {
     have_true = 0;
     have_false = 0;
     for (int x = 0; x < N; ++x) {
+      if( num_attempts == 0 ) {
+        a[x] = 0;
+      }
       for (int y = 0; y < 16; ++y) {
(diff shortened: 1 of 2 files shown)


tmh97:
Fortran versions works with CCE 13, gives following warning 

`Line = 54 : An OpenMP parallel construct in a target region is limited to a single thread`

Thus, the test operates with one thread, and so the reduction clause is not properly tested with this compiler.

tests/5.0/loop/test_loop_reduction_bitor_device.c
```diff
@@ -31,6 +31,9 @@ int test_bitor() {
     have_true = 0;
     have_false = 0;
     for (int x = 0; x < N; ++x) {
+      if( num_attempts == 0 ) {
```
spophale, line 34: Why not initialize a to 0s outside the while ?

seyonglee: Where to initialize the array `a` is just an implementation choice, but I agree that initializing the array before the `while` loop will be more efficient; I will change the initialization code as suggested.

spophale (approved):

PR #537, opened 2022-05-11
Add new_test/test_loop_reduction_bitxor_device.F90

seyonglee:

### DIFF
--- a/tests/5.0/loop/test_loop_reduction_bitxor_device.F90
+++ b/tests/5.0/loop/test_loop_reduction_bitxor_device.F90
@@ -1,0 +1,85 @@
+!===--- test_loop_reduction_bitxor_device.F90 -------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test uses the reduction clause on a loop directive, testing that the
+! variable in the reduction clause is properly reduced using the bitxor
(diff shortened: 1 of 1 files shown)


tob2:
Works with GCC mainline (but only tried with host callback). 

* "Warning: Unused variable ‘y’ declared" (line 32)
* As GCC 12 has been released, you could ask the HPC admin to make it available
  * GCC 12 announcement: https://gcc.gnu.org/pipermail/gcc/2022-May/238653.html
  * Linux distros have early access additional packages of it (and Fedora 36 and openSUSE Tumbleweed have it as default)

tmh97 (approved):

tmh97:
Passes with Cray (CCE 13) on spock. 

Noticing that the Cray ftn compiler issues the following warning 

`ftn-7256 ftn: WARNING TEST_BITXOR, File = ../../../autofs/nccs-svm1_home1/tmh97/sollve_vv/tests/5.0/loop/test_loop_reduction_bitxor_device.F90, Line = 54 
   An OpenMP parallel construct in a target region is limited to a single thread.`

Thus, the test operates with one thread, and so the reduction clause is not properly tested with this compiler.

spophale (approved):

PR #539, opened 2022-05-11
Add new_test/test_loop_private_device.F90

seyonglee:
LLVM 15.0.0 crashes when compiling the C version.

### DIFF
--- a/tests/5.0/loop/test_loop_private_device.F90
+++ b/tests/5.0/loop/test_loop_private_device.F90
@@ -44,11 +44,11 @@ PROGRAM test_loop_private_device
         privatized = privatized + 1 
      END DO
      d(x) = c(x) * privatized
-     IF (omp_get_thread_num() .eq. 0 ) THEN
-        num_threads = omp_get_num_threads()
-     END IF
   END DO
   !$omp end loop
+  IF (omp_get_thread_num() .eq. 0 ) THEN
+     num_threads = omp_get_num_threads()
+  END IF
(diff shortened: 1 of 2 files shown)


tob2:
New testcase fails with GCC mainline (GCC 13) with:
```
test_loop_private_device.F90:47:9:

   47 |      IF (omp_get_thread_num() .eq. 0 ) THEN
      |         ^
Error: OpenMP runtime API call ‘omp_get_thread_num’ in a region with ‘order(concurrent)’ clause
test_loop_private_device.F90:48:43:

   48 |         num_threads = omp_get_num_threads()
      |                                           ^
Error: OpenMP runtime API call ‘omp_get_num_threads’ in a region with ‘order(concurrent)’ clause
```

OpenMP 5.1 has for **order(concurrent)**, https://www.openmp.org/spec-html/5.1/openmpsu47.html#x72-720002.11.3
* "A region that corresponds to a construct with an **order** clause that specifies **concurrent** may not contain calls to the OpenMP Runtime API. "

And for the **loop** construct, https://www.openmp.org/spec-html/5.1/openmpsu51.html#x76-840002.11.7
* "Description [...] If the **order** clause is not present, the behavior is as if an **order** clause that specifies **concurrent** appeared on the construct."

seyonglee:
@tob2 Thank you for catching this error; fixed.
(OpenMP V5.0 also has the same restriction on the loop construct.)

spophale (approved):

PR #541, opened 2022-05-12
Fix mapping in target/test_target_defaultmap_to_from_tofrom.F90

tob2:
This testcase did spurious fail on some host/gpu arch combinations when the Fortran pointer ptr wasn't nullified, either only for 'to' or for 'from'. This commit does:
* Actually map the pointer (i.e. apply 'defaultmap(to/from/fromto) also to category 'pointer') instead of just using it inside the target.
* Initialize all variables to /= 0 to make checking work more properly.
* For 'tofrom', actually use the passed value to distinguish 'from' from 'tofrom'.

Note: Regarding the pointer handling, see also the "restrictions to the target construct" (last three bullet points in OpenMP 5.1/5.2). In OpenMP 5.0, the third-last item describes this with looser words.

Namely, OpenMP 5.0 has, https://www.openmp.org/spec-html/5.0/openmpsu60.html#x86-2880002.12.5
>  * An attached pointer that is associated with a given pointer target must not become associated with a different pointer target in a target region.

And OpenMP 5.1 (likewise 5.2), https://www.openmp.org/spec-html/5.1/openmpsu68.html#Q1-96-392
>  * An attached pointer that is associated with a given pointer target must not become associated with a different pointer target in a target region.
> * [...] (4 other items)
> * If the association status of a list item with the POINTER attribute that appears in a map clause on the construct is associated upon entry to the target region, the list item must be associated with the same pointer target upon exit from the region.
> * If the association status of a list item with the POINTER attribute that appears in a map clause on the construct is disassociated upon entry to the target region, the list item must be disassociated upon exit from the region.
> * If the association status of a list item with the POINTER attribute that appears in a map clause on the construct is disassociated or undefined on entry to the target region and if the list item is associated with a pointer target inside the target region, the pointer association status must become disassociated before the end the region.

@spophale @nolanbaker31 @tmh97 @seyonglee @mjcarr458 @krishols @jrreap – please review.

### DIFF
--- a/tests/5.0/target/test_target_defaultmap_to_from_tofrom.F90
+++ b/tests/5.0/target/test_target_defaultmap_to_from_tofrom.F90
@@ -40,15 +40,16 @@ CONTAINS
 
 
     TYPE(structure) :: new_struct !aggregate
-    ALLOCATE(B(N))
+    ALLOCATE(B(N), ptr(N))
 
     scalar = 1
     new_struct%s = 1
 
     DO i = 1, N
-       A(i) = 0
-       B(i) = 0
-       new_struct%SA(i) = 0
+       A(i) = 1
+       B(i) = 1
+       ptr(i) = 1
+       new_struct%SA(i) = 1
     END DO
 
     errors = 0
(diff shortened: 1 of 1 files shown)


tob2:
Cross ref in Pull Request #496, I had fixed some issues before, including avoiding one undefined-pointer issue by using `ptr=>null()`.
_(But only in 'tofrom', I missed 'to' and that fixed half of the fails, probably because the properly initialized stack memory got reused on some systems; in any case, the new fix is better as it uses pointers more properly and actually tests them with regards to `defaultmap`.)_

seyonglee (approved):
@tob2 thank you for this fix; I agree with these changes.

spophale (approved):

PR #547, opened 2022-05-21
Add new_test/test_loop_lastprivate.F90

seyonglee:

### DIFF
--- a/tests/5.0/loop/test_loop_lastprivate.F90
+++ b/tests/5.0/loop/test_loop_lastprivate.F90
@@ -89,12 +89,12 @@ CONTAINS
     !$omp end loop
     !$omp end parallel
 
-    OMPVV_TEST_AND_SET_VERBOSE(lp_errors_x, x .ne. NSIZE)
-    OMPVV_TEST_AND_SET_VERBOSE(lp_errors_y, y .ne. NSIZE2)
+    OMPVV_TEST_AND_SET_VERBOSE(lp_errors_x, x .ne. (NSIZE + 1))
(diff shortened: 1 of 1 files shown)


tests/5.0/loop/test_loop_lastprivate.F90
```diff
@@ -0,0 +1,107 @@
...
+    !$omp loop lastprivate(x)
+    DO x = 1, NSIZE
+      a(x) = a(x) + b(x)
+    END DO
+    !$omp end loop
+    !$omp end parallel
+
+    OMPVV_TEST_AND_SET_VERBOSE(lp_errors, x .ne. NSIZE)
+    WRITE(msgHelper, *) "Loop iteration variable in loop construct ended with invalid value."
+    OMPVV_ERROR_IF(x .ne. NSIZE, msgHelper)
```
tob2, line 55: Compiles and runs with GCC mainline (13; should likewise run with released GCC 12). However:
* `lastprivate` value is NSIZE + 1 after the loop.  (This matches Fortran's "11.1.7.4.3 The execution cycle" and the OpenMP rules).
* The test in line line 53 and 55 is effectively the same. Is this extra check needed? Additionally, instead of an internal write, an assignment would do in line 54 as it is currently just a string. Or  it is changed to `"with invalid value: ",x` then it has an added value (both direct and idiomatic meaning).

Likewise below – and also for `y` and `NSIZE2` .

seyonglee:
@tob2 Thank you for finding the error in the last private value checking; now fixed.
The duplicated error checkings are simply because the C version is written in that way; the Fortran version is written to mimic the behavior of the reference C version.

tmh97 (approved):
good-to-go

jrreap (approved):

PR #549, opened 2022-05-21
Add new_test/test_loop_collapse_device.F90

seyonglee:

### DIFF
--- a/tests/5.0/loop/test_loop_collapse_device.F90
+++ b/tests/5.0/loop/test_loop_collapse_device.F90
@@ -59,10 +59,10 @@ CONTAINS
        temp_total = 0
        DO y = 1, N+1
           OMPVV_TEST_AND_SET(errors, (temp_total - b(y,x)) .ne. 0)
+          IF ( y .ne. N+1 ) THEN
+             temp_total = temp_total + a(y,x)
+          END IF
        END DO
-       IF ( y .ne. N+1 ) THEN
-          temp_total = temp_total + a(y,x)
-       END IF
     END DO
 
     test_collapse1 = errors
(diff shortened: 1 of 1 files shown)


tests/5.0/loop/test_loop_collapse_device.F90
```diff
@@ -0,0 +1,122 @@
...
+    !$omp end loop
+    !$omp end target parallel
+
+    DO x = 1, N
+       temp_total = 0
+       DO y = 1, N+1
+          OMPVV_TEST_AND_SET(errors, (temp_total - b(y,x)) .ne. 0)
+       END DO
+       IF ( y .ne. N+1 ) THEN
+          temp_total = temp_total + a(y,x)
```
tob2, line 64: This leads to out of bounds: The loop goes from `y = 1` to `N+1` – and after the loop, `y = N + 2`. However, `a(1:N)` and thus, `y == 130` is out of bounds (`N = 128`).

In the second function, `test_collapse2`, it is inside the DO loop, which avoids the out of bounds.

Still: It is unclear to me what that `temp_total` is good for – it is not used for any error checking, but it also does not show up in an OpenMP context. Additionally, it is not initialized. (Those issues also apply to the second function.)

seyonglee: @tob2 thank you for finding this error; the IF-statement should be moved into the DO-loop as you suggested.
`temp_total` is used to check whether the innermost loop that is out of the scope of the collapsed loops was parallelized or not by checking possible race condition (race condition may occur to `b(y,x)` if the innermost loop was parallelized.).

seyonglee: The code was changed to properly initialize `temp_total`.

seyonglee:
[Status update on July 20, 2022]
- XL 16.1.1-10: C test passes but run on host, Fortran test failed.
- LLVM 15.0.0: C test passes.
- GCC 12.1.1: both C and Fortran tests pass.

spophale (approved):

krishols (approved):

Issue #507, opened 2022-03-29
C and Fortran test that utilize I/O in a target region


tmh97:
Overall,  we try and avoid using printf and our own OMPVV_TEST_AND_SET or OMPVV_TEST_AND_SET_VERBOSE inside of a target region. 

Of course, many users will be looking for printf inside of target region for help with debugging. We should add a .c and .f90 test to the suite that tests exactly this.

tmh97:
References recommendation from @tob2 in #505

PR #555, opened 2022-06-09
Adding print in target region test to address issue #507

tmh97:
Runs with latest llvm trunk.

There's no error checking here as I am simply looking to test that printf in target does not produce a compile/runtime error.

- [ ] Add .f90 version


tmh97:
Closes issue #507

jrreap (approved):

nolanbaker31 (approved):
looks good

PR #557, opened 2022-06-11
New test for the order reproducable clause with the taskloop construct

jrreap:
A modified test to check the order clause on a taskloop construct

### DIFF
--- a/tests/5.1/order/test_taskloop_order_reproducible_device.c
+++ b/tests/5.1/order/test_taskloop_order_reproducible_device.c
@@ -1,4 +1,4 @@
-//===------------ test_target_taskloop_order_reproducible.c -------------===//
+//===------------ test_target_taskloop_order_reproducible_device.c ----------===//
 //
 // OpenMP API Version 5.1 Nov 2020
 //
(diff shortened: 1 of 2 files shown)


spophale (approved):

tmh97 (approved):

tests/5.1/order/test_taskloop_order_reproducible_device.c
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	OMPVV_TEST_OFFLOADING;
+
+	#pragma omp target map(tofrom: x,y)
+	{
+
+		#pragma omp parallel
+		{
+		   #pragma omp taskloop order(reproducible:concurrent)
```
tob2, line 36: This testcase fails with GCC as follows:
```
error: ‘order’ is not valid for ‘#pragma omp taskloop’
```

If I look at OpenMP 5.2, I see:
> 12.6 `taskloop` Construct
> ...
> **Clauses**
> allocate, collapse, default, final, firstprivate, grainsize, if, in_reduction, lastprivate, mergeable, nogroup, num_tasks, priority,1 private, reduction, shared, untied

which does not list  the `order` clause. That clause in turn is only permitted for:

> 10.3 `order` Clause
> ...
> **Directives**
> distribute, do, for, loop, simd

Unsurprisingly, OpenMP 5.1 also does not permit it: https://www.openmp.org/spec-html/5.1/openmpsu55.html#x81-910002.12.2

PR #561, opened 2022-06-15
Test for Masked construct w/ & w/o filter clause

nolanbaker31:
There might be a more exhaustive way to test this construct, but this should work. LMK if anyone has any ideas.


tests/5.1/masked/test_masked.c
```diff
@@ -0,0 +1,51 @@
...
+  int total = 10;
+  for(i=0; i<N; i++){
+    x[i] = 0;
+  }
+  int threads = OMPVV_NUM_THREADS_HOST;
+
+#pragma omp parallel num_threads(threads)
+while(total > 0){
+    #pragma omp for
+    for(i=0; i<N; i++){ // give threads work
```
spophale, line 32: This is not useful if you are not checking value of x.

tests/5.1/masked/test_masked.F90
```diff
@@ -0,0 +1,54 @@
...
+    threads = omp_get_num_threads()
+
+    !$omp parallel num_threads(threads)
+    DO WHILE (total>0)
+        !$omp masked
+            OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_thread_num() .ne. 0) ! primary thread
+            ct = ct + 1
+            total = total - 1
+        !$omp end masked
+    END DO
```
tob2, line 49: With `parallel` either a `BLOCK ... END BLOCK` or an end-directive is required, e.g.:
```suggestion
    END DO
    !$omp end parallel
```

tests/5.1/masked/test_masked.F90
```diff
@@ -0,0 +1,54 @@
...
+    INTEGER:: i
+
+    OMPVV_INFOMSG("test_masked")
+
+    DO i = 0, N
+       x(i) = 0
+    END DO
+    threads = omp_get_num_threads()
+
+    !$omp parallel num_threads(threads)
```
tob2, line 42:
```
Error: Symbol ‘threads’ at (1) has no IMPLICIT type
```

tests/5.1/masked/test_masked.F90
```diff
@@ -0,0 +1,54 @@
...
+    INTEGER:: errors = 0
+    INTEGER:: total = 10
+    INTEGER:: ct = 0
+    INTEGER, DIMENSION(N):: x
+    INTEGER:: i
+
+    OMPVV_INFOMSG("test_masked")
+
+    DO i = 0, N
+       x(i) = 0
```
tob2, line 38:
```
Warning: Array reference at (1) out of bounds (0 < 1) in loop beginning at (2)
```
Thus, either change the bounds to `DIMENSION(0:N)` or the loop beginning to `1`.

Additionally, `x` assigned to outside any OpenMP construct but otherwise not used. (I think @spophale did remark this already.)

tests/5.1/masked/test_masked.c
```diff
@@ -0,0 +1,47 @@
...
+#define N 1024
+
+int test_masked() {
+  int errors = 0;
+  int i;
+  int x[N];
+  int ct = 0;
+  int total = 10;
+  for(i=0; i<N; i++){
+    x[i] = 0;
```
tob2, line 25: Additionally, x assigned to outside any OpenMP construct but otherwise not used. (I think @spophale did remark this already.)

tests/5.1/masked/test_masked.F90
```diff
@@ -0,0 +1,54 @@
...
+    END DO
+    threads = omp_get_num_threads()
```
tob2, line 40: This very likely gives: `threads = 1`. Thus, the loop is run as `omp parallel num_treads(1)` — which is fine, but odd.

However, using `OMPVV_NUM_THREADS_HOST` as in C seems to make more sense.

tests/5.1/masked/test_masked_filter.F90
```diff
@@ -0,0 +1,54 @@
...
+    INTEGER:: ct = 0
+    INTEGER, DIMENSION(N):: x
+    INTEGER:: i
+
+    OMPVV_INFOMSG("test_masked")
+
+    DO i = 0, N
+       x(i) = 0
+    END DO
+    threads = omp_get_num_threads()
```
tob2, line 40: See comments for the other Fortran file – all comments should also apply here.

However, with `num_threads(1)`, the testcase will deadlock as `filter(3)` will never become true. Thus, here, `num_threads(OMPVV_NUM_THREADS_HOST)` (directly or via assignment to `threads`) is required.

tmh97 (approved):

PR #562, opened 2022-06-22
scope reduction construct test

mjcarr458:
first iteration of scope construct, based off of 5.1 example sheet. scope construct currently not implemented


mjcarr458:
Updated tests with suggested comments. The test should be designed to reduce an array using scope reduction clause based off example in OpenMP example sheet.

tests/5.1/scope/test_scope_construct.c
```diff
@@ -0,0 +1,59 @@
...
+	int s = 0;
+	int errors = 0;
+	for (int i; i < N; i++){
+		a[i] = 1;
+	}
+	OMPVV_TEST_OFFLOADING;
+	OMPVV_TEST_AND_SET_VERBOSE(errors, test_scope() != 0);
+	OMPVV_REPORT_AND_RETURN(errors);
+	test_scope(N,a,s);
+	return s;
```
tob2, line 55:
```suggestion
	OMPVV_TEST_AND_SET_VERBOSE(errors, test_scope(N,a,s) != 0);
	OMPVV_REPORT_AND_RETURN(errors);
```

That is: Pass the arguments to the function – and remove the now unreachable second function call and return.

tests/5.1/scope/test_scope_reduction_construct.c
```diff
@@ -0,0 +1,58 @@
...
+	int a[N];
+	int s = 0;
+	int errors = 0;
+	for (int i; i < N; i++){
+		a[i] = 1;
+	}
+	OMPVV_TEST_OFFLOADING;
+	OMPVV_TEST_AND_SET_VERBOSE(errors, test_scope() != 0);
+	OMPVV_REPORT_AND_RETURN(errors);
+	return s;
```
tob2, line 54:
```suggestion
	OMPVV_TEST_AND_SET_VERBOSE(errors, test_scope(N,a,s) != 0);
	OMPVV_REPORT_AND_RETURN(errors);
```

tob2 (commented):
Question: What should be the difference between those two files, except for the file name?

Both files do a reduction (hence, the now suggested 'firstprivate→shared' change) – and the only difference 'diff' shows is that one file contains a second 'test_scope' call (which is unreachable as placed after the return macro).

tests/5.1/scope/test_scope_reduction_construct.c
mjcarr458: > 

Sorry about this confusion

> Question: What should be the difference between those two files, except for the file name?
> 
> Both files do a reduction (hence, the now suggested 'firstprivate→shared' change) – and the only difference 'diff' shows is that one file contains a second 'test_scope' call (which is unreachable as placed after the return macro).

Sorry about this confusion, test_scope_construct.c was left over from a previous test iteration before we decided to split scope tests. It should be removed now.

spophale (approved):

krishols (approved):

PR #563, opened 2022-06-22
New test for the order reproducable clause with the loop construct

jrreap:
Similar to previous tests, utilizes a different construct to ensure the order clause works on this construct as well.

### DIFF
--- a/tests/5.1/order/test_loop_order_reproducible.c
+++ b/tests/5.1/order/test_loop_order_reproducible.c
@@ -1,0 +1,53 @@
+//===------------ test_loop_for_order_reproducible.c -------------===//
+//
+// OpenMP API Version 5.1 Nov 2020
+//
+// This test checks that the order(reproducible:concurrent) clause is properly handled.
+// Leverages the previous array item to determine if the same threads executed
+// in the same order over the array. If both arrays match the test passes, 
+// if not it fails as the reproducible clause failed.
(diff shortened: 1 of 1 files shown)


tests/5.1/order/test_loop_order_reproducible.c
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	OMPVV_TEST_OFFLOADING;
+
+	#pragma omp target map(tofrom: x,y)
+	{
+
+	   #pragma omp parallel 
+	   {
+		#pragma omp loop order(reproducible:concurrent) nowait
```
tmh97, line 36:
```suggestion
		#pragma omp loop order(reproducible:concurrent) nowait
```
```suggestion
		#pragma omp loop order(reproducible:concurrent) 
```

tmh97, line 36:
```suggestion
		#pragma omp loop order(reproducible:concurrent) 
```
I do not believe that the nowait clause is valid on the loop construct

jrreap: Yes it is not, good catch! Must've slipped through

tmh97 (approved):

tmh97 (approved):

spophale (approved):

PR #567, opened 2022-06-27
Fix test_taskloop_order_reproducible_device.c

tob2:
Follow-up to Pull Request #557

* Use 'taskloop simd' as 'taskloop' itself does not take the 'order' clause.
* Rename tests/5.1/order/test_taskloop_order_reproducible_device.c to tests/5.1/order/test_taskloop_simd_order_reproducible_device.c
* Use 'omp single' + 'omp taskloop' to avoid running the taskloop num_thread times, ensuring that the result is as expected.


spophale (approved):
Thanks @tob2 .

jrreap (approved):
Good catch, thanks!

PR #568, opened 2022-06-29
omp_target_memcpy_rect_async test

nolanbaker31:
https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5-1.pdf#page=446

Unsure if my 2D arrays are allocated properly, and also unsure of what the "dst/src_dimensions" should be, as well as what volume should be. If it was a 3x5 array, then what would volume be? Seems it could only work for squres.

### DIFF
--- a/tests/5.1/target/test_target_memcpy_async_depobj.c
+++ b/tests/5.1/target/test_target_memcpy_async_depobj.c
@@ -64,6 +64,7 @@ int test_target_memcpy_async_depobj() {
         OMPVV_TEST_AND_SET(errors, mem[i]!=i*2);
     }
     // free resources
+    free(mem);
     omp_target_free(mem_dev_cpy, t);
     #pragma omp depobj(obj) destroy
     return errors;
--- a/tests/5.1/target/test_target_memcpy_async_no_obj.c
+++ b/tests/5.1/target/test_target_memcpy_async_no_obj.c
@@ -55,10 +55,11 @@ int test_target_memcpy_async_no_obj() {
                                 0,          NULL);
 
     #pragma omp taskwait
-    
     for(i = 0; i < N; i++){
         OMPVV_TEST_AND_SET(errors, mem[i]!=i*2);
     }
+    // free resources
+    free(mem);
     omp_target_free(mem_dev_cpy, t);
     return errors;
 }
--- a/tests/5.1/target/test_target_memcpy_async_no_obj.c
+++ b/tests/5.1/target/test_target_memcpy_async_no_obj.c
@@ -45,7 +45,7 @@ int test_target_memcpy_async_no_obj() {
 
     #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
     #pragma omp teams distribute parallel for
-    for(i = 0;i<N;i++){
+    for(i = 0; i < N; i++){
         mem_dev_cpy[i] = i*2; // initialize data on device
     }
     /* copy to host */
@@ -56,7 +56,7 @@ int test_target_memcpy_async_no_obj() {
 
     #pragma omp taskwait
     
-    for(int i=0;i<N;i++){
+    for(i = 0; i < N; i++){
         OMPVV_TEST_AND_SET(errors, mem[i]!=i*2);
     }
     omp_target_free(mem_dev_cpy, t);
--- a/tests/5.1/target/test_target_memcpy_rect_async_depobj.c
+++ b/tests/5.1/target/test_target_memcpy_rect_async_depobj.c
@@ -2,7 +2,7 @@
 //
 //  Inspired from OpenMP 5.1 Examples Doc, 5.16.4 & 8.9
 //  This test utilizes the omp_target_memcpy_rect_async construct to
-//  allocate memory on the device asynchronously. The construct
+//  allocate 2D memory on the device asynchronously. The construct
 //  uses 'obj' for dependency, so that memory is only copied once
 //  the variable listed in the depend clause is changed.
 //
@@ -27,22 +27,17 @@ int test_target_memcpy_async_depobj() {
 
     int h, t;
     errors = 0;
-    //double **devRect;
     h = omp_get_initial_device();
     t = omp_get_default_device();
 
     double hostRect[N][M]; // 5x10 2D array
-    //double devRect = (double **)omp_target_alloc( sizeof(double)*N, t);
-    /*for(i = 0; i < N; i++){
-        devRect[i] = (double **)omp_target_alloc( sizeof(double)*M, t);
-    }*/
     double *devRect = (double *)omp_target_alloc(sizeof(double)*N*M, t);
 
     OMPVV_TEST_AND_SET_VERBOSE(errors, devRect == NULL);
 
     for(i = 0; i < N; i++){             //each index is set to number of their row
         for (j = 0; j < M; j++){
-            hostRect[i][j] = i;
+            hostRect[i][j] = i + j;
         }
     }
     omp_depend_t obj;
@@ -63,7 +58,7 @@ int test_target_memcpy_async_depobj() {
     {
         for(i = 0; i < N; i++){
             for (j = 0; j < M; j++){
-                devRect[i*N + j] = devRect[i*N + j]*2; // initialize data
+                devRect[i*M + j] = devRect[i*M + j]*2; // initialize data
             }
         }
     }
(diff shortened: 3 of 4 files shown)


tests/5.1/target/test_target_memcpy_rect_async_depobj.c
```diff
@@ -0,0 +1,92 @@
...
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+#define N 5
```
spophale, line 17: #define M 10

tests/5.1/target/test_target_memcpy_rect_async_depobj.c
```diff
@@ -0,0 +1,92 @@
...
+
+int errors, i, j;
+
+int test_target_memcpy_async_depobj() {
+
+    int h, t;
+    errors = 0;
+    double *devRect;
```
tob2, line 25:
```suggestion
    double **devRect;
```
otherwise it will later fail with:
```
test_target_memcpy_rect_async_depobj.c:35:22: error: incompatible types when assigning to type 'double' from type 'double *'
   35 |         devRect[i] = (double *)omp_target_alloc( sizeof(double)*N, t);
```

tests/5.1/target/test_target_memcpy_rect_async_depobj.c
```diff
@@ -0,0 +1,92 @@
...
+    h = omp_get_initial_device();
+    t = omp_get_default_device();
+
+    double* hostRect[N];
+    for(i = 0; i < N; i++){
+        hostRect[i] = (double *)malloc( sizeof(double)*N); 
+    } // 5x5 2D array
+    devRect = (double *)omp_target_alloc( sizeof(double)*N, t);
```
tob2, line 33: Due to the change above:
```suggestion
    devRect = (double **)omp_target_alloc( sizeof(double)*N, t);
```

tests/5.1/target/test_target_memcpy_rect_async_depobj.c
```diff
@@ -0,0 +1,92 @@
...
+    for(i = 0; i < N; i++){
+        devRect[i] = (double *)omp_target_alloc( sizeof(double)*N, t);
```
tob2, line 35: However, with the changes above, this will fail (except with shared memory) as `devRect[i]` will then dereference device memory on the host!

Additionally, I think `omp_target_memcpy_rect_async` assumes a single stream of contiguous memory. Thus, I believe that you need to use something like:

```C
// Either:
     double hostRect[N][M];
// Or
    double *hostRect;
    hostRect = (double*) malloc(sizeof(double)*N*M);
// And:
   double *devRect = (double *)omp_target_alloc( sizeof(double)*N*M, t);
// ...
    for(i = 0; i < N; i++){
        for (j = 0; j < M; j++){
            devRect[i*N + j] *= 2;
// Likewise for hostRect in the 'malloc' case – if defined as 'hostRect[N][M]', using hostRect[i][j] is fine.
```

tests/5.1/target/test_target_memcpy_rect_async_depobj.c
```diff
@@ -0,0 +1,92 @@
...
+    #pragma omp depobj(obj) depend(inout: devRect)
+    omp_depend_t obj_arr[1] = {obj};
+
+    /* copy to device memory */
+    omp_target_memcpy_rect_async(devRect, hostRect, 
+                                32 /*number of bits??*/, 2, 5, //5 by 5
+                                0,          0,
+                                sizeof(double)*N*N, sizeof(double)*N*N,
+                                t,          h,
+                                1,          obj_arr);
```
tob2, line 55: Fails here with:
```
test_target_memcpy_rect_async_depobj.c:51:61: warning: passing argument 5 of 'omp_target_memcpy_rect_async' makes pointer from integer without a cast [-Wint-conversion]
   51 |                                 32 /*number of bits??*/, 2, 5, //5 by 5
      |                                                             ^
```
The prototype is:
```C
int omp_target_memcpy_rect_async(
  void *dst,
  const void *src,
  size_t element_size,
  int num_dims,
  const size_t *volume,
  const size_t *dst_offsets,
  const size_t *src_offsets,
  const size_t *dst_dimensions,
  const size_t *src_dimensions,
  int dst_device_num,
  int src_device_num,
  int depobj_count,
  omp_depend_t *depobj_list
);
```
I think it should be something like the following:
```suggestion
    const size_t volume[2] = {5, 5};
    const size_t offsets[2] = {0, 0};
    const size_t dimensions[2] = {N, N};
    omp_target_memcpy_rect_async(devRect, hostRect,
                                 sizeof (double), 2,
                                 volume,
                                 offsets,    offsets,
                                 dimensions, dimensions,
                                 t,          h,
                                 1,          obj_arr);
```
And likewise below.

_Disclaimer: I have not made any attempt to check the bounds of the to-be transferred rectangle and whether it makes sense with the rest of the program._
_I have also not checked whether I interpreted the arguments correctly, but glancing at the spec, those should be okay._

tests/5.1/target/test_target_memcpy_rect_async_depobj.c
```diff
@@ -0,0 +1,92 @@
...
+
+    #pragma omp taskwait depend(depobj: obj)
+    for(i = 0; i < N; i++){
+        for(j = 0; j < N; j++){
+            OMPVV_TEST_AND_SET(errors, hostRect[i][j]!=i*2);
+        }
+    }
+    // free resources
+    omp_target_free(devRect, t);
```
tob2, line 82: If going for the 'malloc'ed variant, there should be also a `free(hostRect)`.

_(Note: with the current variant, in order to free all memory, a loop to `free()` the array elements of hostRect and devRect would be required. However, following my suggestion to use a single `(omp_target_)malloc` above, using a single `(omp_target_)free`  call per variable is sufficient.)_

tests/5.1/target/test_target_memcpy_rect_async_depobj.c
nolanbaker31: This seems correct, the use of arrays confused me at first i.e. "unsure of what the "dst/src_dimensions" should be, as well as what volume should be". I will use this w/ Dimensions as `{N, M}`

tests/5.1/target/test_target_memcpy_async_depobj.c
```diff
@@ -48,18 +48,19 @@ int test_target_memcpy_async_depobj() {
...
                                 0,          0,
                                 h,          t,
                                 1,          obj_arr);
 
     #pragma omp taskwait depend(depobj: obj)
-    for(int i=0; i < N; i++){
+    for(i = 0; i < N; i++){
         OMPVV_TEST_AND_SET(errors, mem[i]!=i*2);
     }
     // free resources
```
tob2, line 66:
```suggestion
    // free resources
    free (mem);
```

tests/5.1/target/test_target_memcpy_async_no_obj.c
```diff
@@ -56,7 +56,7 @@ int test_target_memcpy_async_no_obj() {
 
     #pragma omp taskwait
     
-    for(int i=0;i<N;i++){
+    for(i = 0; i < N; i++){
         OMPVV_TEST_AND_SET(errors, mem[i]!=i*2);
     }
```
tob2, line 61:
```suggestion
    }
    free (mem);
```

tests/5.1/target/test_target_memcpy_rect_async_depobj.c
```diff
@@ -0,0 +1,97 @@
...
+                                dimensions, dimensions,
+                                t,          h,
+                                1,          obj_arr);
+
+    #pragma omp taskwait depend(depobj: obj)
+    #pragma omp target is_device_ptr(devRect) device(t) depend(depobj: obj)
+    {
+        for(i = 0; i < N; i++){
+            for (j = 0; j < M; j++){
+                devRect[i*N + j] = devRect[i*N + j]*2; // initialize data
```
tob2, line 66: Sorry, made a thinko – this should be:
```suggestion
                devRect[i*M + j] = devRect[i*M + j]*2; // initialize data
```

spophale (approved):

mjcarr458 (commented):
looks good.

mjcarr458 (approved):

PR #569, opened 2022-06-30
Fix 5.1/target/test_target_is_accessible.c

tob2:
* tests/5.1/target/test_target_is_accessible.c: Use omp_target_malloc instead of malloc to make it work also on systems that do not share memory between host and device.

While the testcase is based on the OpenMP example, the latter uses the following to make it work with both (unified) shared memory and unshared memory:
```C
#pragma omp metadirective \
            when(user={condition(accessible)}: target firstprivate(ptr) ) \
            otherwise( target map(ptr[:n]) )
```

### DIFF
--- a/tests/5.1/target/test_target_is_accessible.c
+++ b/tests/5.1/target/test_target_is_accessible.c
@@ -4,8 +4,8 @@
 //
 // This test checks that the omp_target_is_accessible device routine.
 // In this test the output of the target_is_accessible call should return
-// true because the storage indicated by the first and second arguements
-// is accessible by the targeted device. This test is closely adapdted
+// true because the storage indicated by the first and second arguments
+// is accessible by the targeted device. This test is closely adapted
 // from the 5.1 OpenMP example sheet.
 //-----------------------------------------------------------------------//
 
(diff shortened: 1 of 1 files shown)


spophale:
@tob2 thanks for the explanation.

What really throws me off is that the description says "whether *host* memory is accessible from a given device". Normally host memory is stack or heap variables. It would be clearer if it just said " memory" since the spec does not tie the memory being accessible to requires unified shared memory (since that is the only case where device ptr and host ptr are identical.)

tob2:
> What really throws me off is that the description says "whether _host_ memory is accessible from a given device". Normally host memory is stack or heap variables. It would be clearer if it just said " memory" since the spec does not tie the memory being accessible to requires unified shared memory (since that is the only case where device ptr and host ptr are identical.)

Hmm – I think my patch is then wrong: with  `omp_target_alloc`, it does not create _host memory_ that is accessible from the device but some _device pointer_ – and except with unified_shared_address – passing it to `test_target_is_accessible` won't work.

On the other hand, the test case as is will not work as it implicitly assumes shared memory. Thus, I think we need
* A check whether the memory is shared – and use it as flag whether it is accessible or not
* Add 'omp requires unified_shared_memory'
* ?

spophale:
@tob2 I think we all agree that omp_target_alloc is not the way to go. But this is a great discussion and I would certainly want to see the change in the spec IF the "host memory" was a mistake.

With your blessings I would like to close this PR.

tob2:
> @tob2 I think we all agree that omp_target_alloc is not the way to go. But this is a great discussion

... which is continued in Pull Request #575 ...

> and I would certainly want to see the change in the spec IF the "host memory" was a mistake.

I think the "host memory" makes sense – at least as used in the OpenMP example document and given that on non-unified-shared-address systems, 'omp_target_alloc' might return some special object, which is not directly a dereferenceable pointer. _(Example is on a 32bit host and 64bit target, where a host 'void *' cannot prepresent all device pointers.)_

> With your blessings I would like to close this PR.

I did so myself. — Thanks!

PR #573, opened 2022-07-06
Add new_test/test_requires_unified_shared_memory_stack.F90

seyonglee:
Test passed when compiled with NVHPC 21.11

### DIFF
--- a/tests/5.0/requires/test_requires_unified_shared_memory_stack.F90
+++ b/tests/5.0/requires/test_requires_unified_shared_memory_stack.F90
@@ -1,0 +1,84 @@
+!===--- test_requires_unified_shared_memory_stack.F90 --------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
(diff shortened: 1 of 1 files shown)


spophale:
Thinking about:
`The requires directive must appear in the specification part of a program unit, after any USE statement, any IMPORT statement, and any IMPLICIT statement, unless the directive appears by referencing a module and each clause already appeared with the same arguments in the specification part of the program unit.`

Does this mean that we need a requires usm inside CONTAINS ?

seyonglee:
Fortran OpenMP spec says that the requires directive should be included in all Fortran program units, but program units do not include `internal subprogram`:

**Fortran 90 Spec (ISO/IEC 1539 : 1991 (E)) Section 2.2 Program unit concepts:** "A program unit may be a main program, an external subprogram, a module, or a block data program unit."

spophale (approved):

mjcarr458 (approved):
Good to go

PR #575, opened 2022-07-07
Omp_is_accessible new test version

mjcarr458:


tob2:
This replicates _my_ mistake of (pending) Pull Request #569. Namely:

_(First, had I tested it with actual offloading, it would have failed. Not doing a proper test is one fault of mine. Another is not reading the spec properly:)_

@spophale pointed out in that PR, `omp_is_accessible` takes a _host pointer_. That's not a _device pointer_ or some generic pointer. _(With my thinko, replicated here, `omp_target_alloc`is used – which returns a device pointer.)_

Why this is different:
* That difference is obvious on a host with 32bit pointers – and a device with 64bit pointers. In that case, 'void *' on the host cannot store the result of `omp_target_alloc`; in that case, the implementation would return not a bare pointer – but e.g. an integer (e.g. uint64_t) or a struct or a pointer to some struct. — Obviously, in this case, passing such a pointer to 'omp_is_accessible' does not really work.
* But also if the host and device have the same bit size but a separate addressing (i.e. not unified_address), a given address cannot properly associated to host or device.

_(Granted: In either case, the implementation could unconditionally return 'false' for omp_target_is_accessible, but still it is exceptionally wrong. – And the change to omp_target_alloc is unconditionally wrong.)_

 * * *

Thus, something different is needed. I see three options – there might be more and I am not sure which one is the best:

* Do what the OpenMP example document does – using a metadirective. Downside: convolutes two features, an API call and metadirectives. The example document has:
```C
#pragma omp metadirective \
            when(user={condition(accessible)}: target firstprivate(ptr) ) \
            otherwise( target map(ptr[:n]) )
```

* Simply add `#omp requires unified_shared_memory` – which is the simplest solution, but requires USM (or uses host fallback).

* Using `OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT`, e.g. something like:
```diff
diff --git a/tests/5.1/target/test_target_is_accessible.c b/tests/5.1/target/test_target_is_accessible.c
index bcaa1a5..d0d97f5 100644
--- a/tests/5.1/target/test_target_is_accessible.c
+++ b/tests/5.1/target/test_target_is_accessible.c
@@ -19,2 +19,3 @@ int check_device(){
 	int errors = 0;
+        int isSharedMemory = 0;
 	int check_test = 2;
@@ -26,6 +27,19 @@ int check_device(){
 
+        /* Assumes that on shared-memory systems, no copy is done. */
+	#pragma omp target map(to: isSharedMemory)
+	  isSharedMemory = 1;
+
 	check_test = omp_target_is_accessible(ptr, buf_size, dev);
-	
+
+	if (check_test)
+          {
+	    #pragma omp target firstprivate(ptr)
+	      for (int i = 0; i < N; i++)
+	       ptr[i] = 5*i;
+	    for (int i = 0; i < N; i++)
+	     OMPVV_TEST_AND_SET (errors, ptr[i] != 5*i);
+          }
+
 	free(ptr);
-	OMPVV_TEST_AND_SET_VERBOSE(errors, check_test != 1);
+	OMPVV_TEST_AND_SET_VERBOSE(errors, check_test != isSharedMemory);
 	OMPVV_INFOMSG_IF(check_test == 0, "Omp_target_is_accessible is 0");
```

Note to the diff above:
* I only tested the change with host fallback.
* I do my own isSharedMemory test as `OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT` uses `var2set = (_ompvv_isOffloadingOn && _ompvv_isSharedEnv == 0);` which is not what I want to test for. – The host fallback is already taken care of by the "Test is running on host." / "Test passed on the host." diagnostics. And the warning printed do not make sense, either ("running on a shared data environment ... may cause errors").
* _Note that `_ompvv_isSharedEnv == 0` implies shared memory and `_ompvv_isSharedEnv == 1` not shared memory. I think the value/name is reversed, but it works as it is handled consistently in ompvv/ompvv.h._

spophale:
@mjcarr458 , @tob2 Lets have a separate test case for USM. 
For the original test, lets make sure that it FAILS. Since this is clearly indicated in the spec we will not be checking for undefined behavior.

I propose we close this PR.

mjcarr458:
Thank you Tobias for your edit recommendations. Tom and I went through your suggestions and went ahead with your last suggest for checking shared memory.

tests/5.1/target/test_target_is_accessible.c
```diff
@@ -17,17 +17,32 @@
...
 	const int N = 100;
 	const int buf_size = sizeof(int) * N;
 	const int dev = omp_get_default_device();
+	
+	/*Assumes taht one shared-memory systems, no copy is done*/
+	#pragma omp target map(to: isSharedMemory)
+		isSharedMemory = 1;
 
-	int *ptr = (int *) malloc(buf_size);
+	int *ptr = (int *) omp_target_malloc(buf_size,dev);
```
tob2, line 30: I think this one should remain `malloc`.

tests/5.1/target/test_target_is_accessible.c
```diff
@@ -17,17 +17,32 @@
...
 
 	check_test = omp_target_is_accessible(ptr, buf_size, dev);
 	
+	if(check_test)
+		{
+		#pragma omp target firstprivate(ptr)
+			for (int i=0; i<N; i++)
+				ptr[i] = 5*i
```
tob2, line 38:
```suggestion
				ptr[i] = 5*i;
```

tests/5.1/target/test_target_is_accessible.c
```diff
@@ -17,17 +17,32 @@
 
 int check_device(){
 	int errors = 0;
+	int isSharedMemory = 0;
 	int check_test = 2;
 	const int N = 100;
 	const int buf_size = sizeof(int) * N;
 	const int dev = omp_get_default_device();
+	
+	/*Assumes taht one shared-memory systems, no copy is done*/
```
tob2, line 26:
```suggestion
	/* Assumes that on shared-memory systems, no copy is done */
```

tob2 (commented):
I think this patch goes into the right direction. Thanks for working on that – _and sorry again for the nonsense-suggestion in my pull request._

spophale:
Aah, I see where this is going (and I like it!). Thanks @tob2 . @mjcarr458 I am creating the other scenario under requires USM in #586 .

spophale (approved):

PR #576, opened 2022-07-07
scope construct private test

mjcarr458:


tests/5.1/scope/test_scope_private_construct.c
```diff
@@ -0,0 +1,38 @@
...
+
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+int test_scope(){
+	int errors;
```
tob2, line 18:
```suggestion
	int errors = 0;
```

tests/5.1/scope/test_scope_private_construct.c
```diff
@@ -0,0 +1,38 @@
...
+	int test_int = 1;
+	#pragma omp parallel firstprivate(test_int)
+	{
+		#pragma omp scope private(test_int)
+		{
+			test_int += 1;
```
tob2, line 24: GCC rightly complains:
```
test_scope_private_construct.c:24:34: warning: ‘test_int’ is used uninitialized [-Wuninitialized]
   24 |                         test_int += 1;
      |                         ~~~~~~~~~^~~~
```
as `i` is private and not `firstprivate`. Should this be an assignment instead?

tests/5.1/scope/test_scope_private_construct.c
```diff
@@ -0,0 +1,38 @@
...
+		}
+	}
+	OMPVV_TEST_AND_SET_VERBOSE(errors,test_int != 1);
+	OMPVV_INFOMSG_IF(test_int == 2, "test int was not private");
```
tob2, line 28: I am not sure how 'test_int' can become '2':

* the parallel is 'firstprivate', thus, anything inside 'parallel' should not have any effect on the outside. → Use 'shared()' instead? — Currently, I always expect 1 on the outside.
* And with shared (and not working 'private' on scope: Shouldn't it be between 2 and (num-threads + 1), given that the code runs for num-threads (and the update is not done atomically).

tests/5.1/scope/test_scope_private_construct.c
```diff
@@ -0,0 +1,38 @@
...
+	return errors;
+}
+
+int main(){
+	int errors;
```
tob2, line 33:
```suggestion
	int errors = 0;
```

tests/5.1/scope/test_scope_private_construct.c
```diff
@@ -0,0 +1,38 @@
...
+	OMPVV_TEST_OFFLOADING;
+	OMPVV_TEST_AND_SET_VERBOSE(errors, test_scope() != 0);
+	OMPVV_REPORT_AND_RETURN(errors);
+	return errors;
```
tob2, line 37: Unreachable as OMPVV_REPORT_AND_RETURN returns already.
```suggestion
```

mjcarr458:
Thank you for recommendations I went ahead and made some changes I believe should fix the issues.

tests/5.1/scope/test_scope_private_construct.c
mjcarr458: Your comments make a lot of sense, I went ahead and tried to fix this by first using shared() like you suggested as well as wrapping a task region instead a parallel one around the scope pragma in hopes on only having a single thread run the operations and negating the need to worry about any number of threads in the test check.

spophale (approved):

tests/5.1/scope/test_scope_private_construct.c
```diff
@@ -0,0 +1,37 @@
...
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+int test_scope(){
+	int errors = 0;
+	int test_int = 1;
+	#pragma omp task shared(test_int)
+	{
+		#pragma omp scope private(test_int)
```
tob2, line 22: _I know that this has already merged. – Still as remark to this pull request:_

This **fails** with GCC mainline with:
```
tests/5.1/scope/test_scope_private_construct.c: In function ‘test_scope’:
tests/5.1/scope/test_scope_private_construct.c:22:25: error: ‘scope’ region may not be closely nested
inside of work-sharing, ‘loop’, explicit ‘task’, ‘taskloop’, ‘critical’, ‘ordered’, ‘master’,
or ‘masked’ region
   22 |                 #pragma omp scope private(test_int)
      |                         ^~~
```

Cf. https://www.openmp.org/spec-html/5.1/openmpse30.html (last item)
> *  A **scope** region may not be closely nested inside a worksharing, **loop**, **task**, **taskloop**, **critical**, **ordered**, **atomic**, or **masked** region.

PR #577, opened 2022-07-07
Error Directive test

nolanbaker31:
Not sure if tests are exhaustive enough, LMK.


tests/5.1/error/test_error_at_runtime.c
```diff
@@ -0,0 +1,35 @@
...
+#include <math.h>
+
+int errors, i;
+
+int test_error_at_runtime() {
+    i = 0;
+    printf("Test should print a 'runtime' error message: \n");
+    #pragma omp parallel
+    {
+        #pragma omp error at(runtime)
```
tob2, line 24: This fails with:
```
test_error_at_runtime.c:24:30: error: expected ‘execution’ or ‘compilation’
   24 |         #pragma omp error at(runtime)
```
I think the keyword was changed late in the process – and the example document got it wrong. (Corrected for the next iteration but still to late ...)

BTW: With GCC (12 or devel version), the program stops at this directive with exit code 1 and the output:
```
libgomp: 
libgomp: fatal error: error directive encountered
```

tests/5.1/error/test_error_severity_warning.c
```diff
@@ -0,0 +1,32 @@
...
+
+int errors, i;
+
+int test_error_severity_warning() {
+    i = 0;
+    printf("Test should print a 'warning' error message")
+    #pragma omp error severity(warning)
+    i+=5;
+    OMPVV_WARNING_IF("Error directive caused runtime error", i!=5);
+}
```
tob2, line 25: GCC gives the following compile error and warning:
```
test_error_severity_warning.c:21:58: error: expected ‘;’ before ‘#pragma’
   21 |     printf("Test should print a 'warning' error message")
      |                                                          ^
      |                                                          ;
   22 |     #pragma omp error severity(warning)
      |             ~~~                                           
test_error_severity_warning.c:25:1: warning: control reaches end of non-void function [-Wreturn-type]
   25 | }
      | ^
```
Fixing the error, it shows the expected warning:
```
test_error_severity_warning.c:22:13: warning: ‘pragma omp error’ encountered
   22 |     #pragma omp error severity(warning)
      |             ^~~
```

tob2 (commented):
For the execution test, I personally think a "message", possibly with a variable as argument instead of a string literal and also a "severity(warning)" test would be useful.

 * * * 

For `test_error_message.c`, a compile-time error is shown. This implies that the compilation will fail.

However, the testsuite expects that compilations succeed and the code then runs. I think we have to tag this testcase somehow (via the file name?) and then:

* Filter it out from the "run c app rule" in the Makefile.
* Add some special rule for the compilation to expect the fail.

 * * *

Do we need some way to tell the testsuite that an exit status code of != 0 is the expected result? – For `test_error_at_runtime.c`, the implied severity is fatal – and, hence, the program will fail at execution!

Maybe something like the following works:

* Encode the should-fail state in the file name, similar to the environment-var handling. In the Makefile, we have have for those:
https://github.com/SOLLVE/sollve_vv/blob/8f4f9f56498498469f016547203d898479644b16/Makefile#L308-L313

* Plus handle the should-fail case in the run script, e.g. like the following (untested):
```
diff --git a/sys/scripts/run_test.sh b/sys/scripts/run_test.sh
index 1d00151..c0f8674 100755
--- a/sys/scripts/run_test.sh
+++ b/sys/scripts/run_test.sh
@@ -2,3 +2,4 @@
 
-#  USAGE: ./run_test.sh [--env name val ...] <APP> [<LOG>].
+#  USAGE: ./run_test.sh [--should-fail] [--env name val ...] <APP> [<LOG>].
+#   --should-fail: expect exit code of != 0
 #   --env name val: set environment variable name to the given value
@@ -10,17 +11,18 @@ function report ()
 {
-  # $1= app, $2=status, $3=output
+  # $1= app, $2=should_fail $3=status, $4=output
   msg="FAIL"
-  if [ $2 -eq 0 ]; then
-    msg="PASS"
-  fi
-  if [ $2 -eq 124 ]; then
+  if [ $3 -eq 124 ]; then
     msg="FAIL: TEST HAS TIMEOUT"
+  elif [ $2 -eq 0 -a $3 -eq 0 ]; then
+    msg="PASS"
+  elif [ $2 -eq 1 -a $3 -ne 0 ]; then
+    msg="PASS"
   fi
-  if [ -n "$3" ]; then
-    echo "$1: $msg. exit code: $2"
+  if [ -n "$4" ]; then
+    echo "$1: $msg. exit code: $3"
     RED='\033[0;31m'
     NC='\033[0m' # No Color
-    echo -e "${RED}$1:\n${3}${NC}" 1>&2
+    echo -e "${RED}$1:\n${4}${NC}" 1>&2
   else
-    echo "$1: $msg. exit code: $2"
+    echo "$1: $msg. exit code: $3"
   fi
@@ -28,10 +30,29 @@ function report ()
 
+should_fail=0
 declare -a env_data
-while [[ "$1" = "--env" ]]; do
-  if [[ -z "$2" ]]; then
-    echo 'ERROR: Expected 'name val' arguments with --env' 1>&2
-    exit -1
-  fi
-  env_data+=("$2" "$3")
-  shift 3
+while [ $# -gt 0 ]; do
+  case "$1" in
+    --env)
+      if [[ -z "$2" ]]; then
+        echo 'ERROR: Expected 'name val' arguments with --env' 1>&2
+        exit -1
+      fi
+      env_data+=("$2" "$3")
+      shift 3
+      ;;
+    --should-fail)
+      should_fail=1
+      shift
+      ;;
+    --)
+      break
+      ;;
+    -*)
+      echo "ERROR: Unknown argument $1" 1>&2
+      exit 1
+      ;;
+    *)
+      break
+      ;;
+  esac
 done
@@ -53,5 +74,5 @@ output=$(printf '%s\n' "${output}" | uniq)
 if [ -z $2 ]; then
-  report $(basename $app) $status
+  report $(basename $app) $should_fail $status
 else
-  report $(basename $app) $status "$output"
+  report $(basename $app) $should_fail $status "$output"
 fi
```

tests/5.1/error/test_error_severity_warning.c
nolanbaker31: @tob2 do you think it would make more sense to just use severity(warning) for each test, instead of changing the makefile? Many tests, not just this one, use multiple directive clauses to ensure one clause works properly, so maybe it makes more sense to just use the warning so that this process doesn't need to be done.

This would cause 2 other issues though,
1. We could still not test for runtime errors, i.e. no way to see if the error warning properly worked
2. Obviously this would not be the case for a severity(fatal) test.

tob2: I think modifying `run.sh` or the Makefile isn't that bad. There are already some special cases. Besides the environment special cases, there is:
```
# Special rule for test that needs OMPVV lib
$(CURDIR)/tests/4.5/application_kernels/qmcpack_target_static_lib.c.o:
```

If code is added to handle should-fail-at-runtime, based on e.g. a directory name (probably better) or on the file name, it should handle setting the env var and the should-run-time-fail at the same time. (Future proof.)

* * *

I think in the long run, it is not bad to have some support for "should fail". In OpenMP 5.2, quite often the following appears:  "If..., runtime error termination is performed."

In order to test this, having some Makefile support for run-time error terminations would be useful (i.e. testing that it fails with an error).

In OpenMP 5.2, I see the ".., runtime error termination is performed" with:

* If a **map** clause with a **present** _map-type-modifier_ ... list item is not present ...
* .... **device_num.** If the _device-description_ evaluates to **omp_invalid_device**, ...
* If the _device_num,_ _src_device_num,_ or _dst_device_num_ argument of a device memory routine has the value
* **omp_invalid_device,** ...
  *  Likewise for the **omp_pause_resource** routine
* _(And the OpenMP development version has one additional case.)_

And also `OMP_TARGET_OFFLOAD` with mandatory value:
> The **mandatory** value specifies that the effect of any device construct or device memory routine that uses a device that is unavailable or not supported by the implementation, or uses a non-conforming device number, is as if the **omp_invalid_device** device number was used.

which goes alongside with
> [OMP_DEFAULT_DEVICE] ... If _target-offload-var_ is **mandatory** and the number of non-host devices is zero then the
_default-device-var_ is initialized to **omp_invalid_device.** Otherwise ...

I think the last one is not directly testable as it is not known beforehand whether there is an offload device available, but it could be tested in two testcases:
* Some tests, which uses `omp_set_default_device(omp_invalid_device)` – and check for error termination.
* Another test with `OMP_TARGET_OFFLOAD=mandatory` + a check that `if omp_num_devices() > 0)`, omp_get_default_device() is >= 0 and < num_devices – while otherwise, `omp_get_default_device() == omp_invalid_device`.

tob2: > We could still not test for runtime errors, i.e. no way to see if the error warning properly worked

With a special case in the `Makefile`, there could be a `grep` on the output, searching for the expected output, both for the compile-time and the run-time check, i.e.

*  `2>&1 |grep '...'` and then use the exit status of the grep for fail/pass
* Or, alternatively, to keep the full output:  2>&1 |tee tmp.log;  grep '...' tmp.log && echo PASS || echo FAIL

And just adding an explicit case for the compilation (similar to qmcpack_target_static_lib.c.o) + filtering the testcase from the 'run:' make target — and a similar line for the run-time check. – The only downside if that the explicit grep is needed – but that's only true for the **error** directive.

_For the other cases (... error termination is performed), the run.sh change with a simple should-fail and run.sh changes would be better, as we do not know the content of the error message. – But that's outside of the scope of this pull request._

Alternatively, we could start with some 'special-tag: value' inside the source-code file – and parse those, e.g., by run.sh and honor them (should fail, match a particular string, etc.). That's more flexible, but requires a grep of all source files.

Thoughts?

nolanbaker31: I like the first strategy better, will look into the makefile changes and push a PR. Thank's so much for your helpful advice!

tests/5.1/error/test_error_at_execution.c
```diff
@@ -0,0 +1,36 @@
...
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+int errors, i;
```
spophale, line 17: "errors" needs to be initialized to 0 and needs to be set when the error condition is met.

tests/5.1/error/test_error_message.c
```diff
@@ -0,0 +1,34 @@
...
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+int errors, i;
```
spophale, line 17: initialize errors, set errors.

tests/5.1/error/test_error_severity_warning.c
```diff
@@ -0,0 +1,33 @@
...
+//
+////===----------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+int errors, i;
```
spophale, line 17: Initialize and set errors.

tests/5.1/error/test_error_message.c
```diff
@@ -0,0 +1,34 @@
...
+
+int test_error_message() {
+    char msg[] = "Success! Error message works properly"
+    i = 0;
+    printf("Test should print a 'Success!' error message: \n");
+    #pragma omp error message(msg)
+    i+=5;
+    OMPVV_WARNING_IF("Error directive caused runtime error", i!=5);
```
spophale, line 25: I am not sure what this test is doing. If there is a RT error the test will terminate at 23 and not reach 25.

tests/5.1/error/test_error_message.c
```diff
@@ -0,0 +1,34 @@
...
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+int errors, i;
+
+int test_error_message() {
+    char msg[] = "Success! Error message works properly"
+    i = 0;
+    printf("Test should print a 'Success!' error message: \n");
```
nolanbaker31, line 22: use info message

tests/5.1/error/test_error_at_execution.c
```diff
@@ -0,0 +1,37 @@
...
+
+int test_error_at_execution() {
+    i = 0;
+    OMPVV_INFOMSG("Test should print a 'execution' error message: \n");
+    #pragma omp parallel
+    {
+        #pragma omp error at(execution) severity(warning)
+        i+=5;
```
nolanbaker31, line 25: fix race condition

tests/5.1/error/test_error_message.c
```diff
@@ -0,0 +1,38 @@
...
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+int errors, i;
+
+int test_error_message() {
+    char msg[] = "Success! Error message works properly"
+    i = 0;
+    OMPVV_INFOMSG("Test should print a 'Success!' error message: \n");
```
spophale, line 22: remove.

tests/5.1/error/test_error_message.c
```diff
@@ -17,12 +17,11 @@
 int errors, i;
 
 int test_error_message() {
-    char msg[] = "error message"
     i = 0;
-    OMPVV_INFOMSG("If successful, should print "error message": ");
-    #pragma omp parallel
+    OMPVV_INFOMSG("If successful, test should print an \"error message success\" at the beginning of the test");
+    #pragma omp single
```
spophale, line 22: @nolanbaker31 
Use omp single inside of omp parallel (not instead of). We want to create a team of threads and then have just 1 thread modify the value.

spophale (approved):

spophale:
@jrreap will give 2nd approval.

jrreap (approved):
Looks good to go! Nice work

PR #579, opened 2022-07-09
Added order unconstrained clause on parallel loop test

jrreap:
Currently not implemented in LLVM. Not sure if GCC supports this yet either


tob2:
> Currently not implemented in LLVM. Not sure if GCC supports this yet either

Compiles & runs with/since GCC 12 (and on GCC mainline).

spophale (approved):

PR #581, opened 2022-07-11
Add new_test/test_requires_unified_shared_memory_stack_is_device_ptr.F90

seyonglee:
- C test passed with NVHPC 22.5
        - C test passed with LLVM 15.0.0-latest
        - C test failed with GCC 11.2.0
        - C test passed with XL 16.1.1-10 but on the host
        - Fortran test passed with NVHPC 22.5
        - Fortran test failed with GCC 11.2.0
        - Fortran test failed with XL 16.1.1-10

### DIFF
--- a/tests/5.0/requires/test_requires_unified_shared_memory_stack_is_device_ptr.F90
+++ b/tests/5.0/requires/test_requires_unified_shared_memory_stack_is_device_ptr.F90
@@ -1,0 +1,83 @@
+!===--- test_requires_unified_shared_memory_stack_is_device_ptr.F90 --------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! Testing memory access from host and device to a variable allocated in the stack
+! The variable is accessed through a pointer to guarantee that there is no default
+! mapping tofrom: of the stack array
+!
+! We use the is_device_ptr to guarantee using the host pointer is used in the device
+!
+!===------------------------------------------------------------------------------===//
+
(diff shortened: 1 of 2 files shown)


tests/5.0/requires/test_requires_unified_shared_memory_stack_is_device_ptr.F90
```diff
@@ -0,0 +1,83 @@
...
+    errors = 0
+    aPtr => anArray
+
+    DO i = 1, N
+      anArray(i) = i
+      anArrayCopy(i) = 0
+    END DO
+
+    ! Test for writes to this varriable from device
+    !$omp target is_device_ptr(aPtr)
```
tob2, line 53: Same comment as in #585 applies here: This violates the 5.0 restrictions for **is_device_ptr,** but it is valid in 5.1. Thus, either it should be moved to directory 5.1 – or the argument must be changed into a dummy argument without POINTER attribute.

seyonglee: Updated the Fortran version to use dummy argument in the is_device_ptr clause.

seyonglee:
[Update on August 16, 2022]
- Update the Fortran test so that dummy argument is used in the is_device_ptr clause. 
- NVHPC 22.5: Both C and Fortran tests passed.
- LLVM 15.0.0: C test passed.
- GCC 12.1.1: C test passed but Fortran test failed (Segmentation fault - invalid memory reference.).

krishols (approved):

PR #586, opened 2022-07-13
Adding new test to check  host memory is accessible when usm is required

spophale:
Related to #575 .

### DIFF
--- a/tests/5.1/requires/test_target_is_accessible_with_usm.c
+++ b/tests/5.1/requires/test_target_is_accessible_with_usm.c
@@ -19,7 +19,7 @@
 
 int check_device(){
   int errors = 0;
-  int check_test = -57;
+  int check_test = 0;
   const int buf_size = sizeof(int) * N;
   const int dev = omp_get_default_device();
 
@@ -46,9 +46,11 @@ int check_device(){
   }
 
   for(int i = 0;i < N; i++)
-   OMPVV_ERROR_IF(ptr[i] != i * i, "Operations on the host memory that was reported as accessible produced incorrect results");
+   OMPVV_ERROR_IF(ptr[i] != i * i, "Operations on the host memory that was reported as accessible produced incorrect results.");
 
-}
+  }
+  else
+  OMPVV_ERROR ("Host memory was reported inaccessible despite of 'requires unified_shared_memory' specified.");
 
   free(ptr);
 
--- a/tests/5.1/requires/test_target_is_accessible_with_usm.c
+++ b/tests/5.1/requires/test_target_is_accessible_with_usm.c
@@ -18,8 +18,8 @@
 #pragma omp requires unified_shared_memory
 
 int check_device(){
-	int errors = -57;
-	int check_test = 0;
+	int errors = 0;
+	int check_test = -57;
 	const int buf_size = sizeof(int) * N;
 	const int dev = omp_get_default_device();
 
--- a/tests/5.1/requires/test_target_is_accessible_with_usm.c
+++ b/tests/5.1/requires/test_target_is_accessible_with_usm.c
@@ -2,7 +2,7 @@
 //
 // OpenMP API Version 5.1 Aug 2021
 //
-// This test checks that the omp_target_is_accessible device routine.
+// This test checks the omp_target_is_accessible device routine.
 // In this test the output of the target_is_accessible call should return
 // true because the requires unified shared memory is used.
 //-----------------------------------------------------------------------//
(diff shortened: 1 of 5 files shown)


tmh97 (approved):

tests/5.1/requires/test_target_is_accessible_with_usm.c
```diff
@@ -0,0 +1,50 @@
...
+
+	check_test = omp_target_is_accessible(ptr, buf_size, dev);
+	
+	free(ptr);
+
+	OMPVV_TEST_AND_SET_VERBOSE(errors, check_test == 0);
+
+	OMPVV_ERROR_IF(check_test == -57, "Call to omp_target_is_accessible did not return true or false");
+	OMPVV_ERROR_IF(check_test == 0, "Call to omp_target_is_accessible did not return true when using USM.");
+	OMPVV_INFOMSG_IF((check_test != 0 && check_test != -57), "The host memory is accessible from the default device.");
```
tmh97, line 36: maybe switch the OMPVV_INFOMSG_IF to `check_test == 1`, if check_test returns 100 for example, that is still wrong behavior

tob2: The `==/!= -57` checks looks odd. I think this testcase mangles two things:

* Checking the return value of `omp_target_is_accessible`. This, I think would require something like the following:
```diff
- 	OMPVV_ERROR_IF(check_test == -57, "Call to omp_target_is_accessible did not return true or false");
+ 	OMPVV_ERROR_IF(check_test != 0 && check_test != 1, "Call to omp_target_is_accessible did not return true or false");
```

* Checking whether the memory was really available on the device, I mean something like the following:
```C
for (int i = 0; i < N; i++)
   ptr[i] = -42;
if (check_test)
  {
    !$omp target is_device_ptr(ptr)
       for (int i = 0; i < N; i++)
         ptr[i] = 2*i;
     for (int i = 0; i < N; i++)
       if (ptr[i] /= 2*i)
         ...
  }
OMPVV_INFOMSG_IF(check_test && ptr[1] == 2, "The host memory is accessible from the default device.");
```
or something along this line.

spophale: @tmh97 @tob2 The first suggestion assumes true == 1, which (maybe is the norm) but not standard . Bool true macro evaluates to 1 and false to 0, but the return value here is an integer.

Even the spec says : "This document uses true as a generic term
for a non-zero integer value and false as a generic term for an integer value of zero."

In essence an implementation could return any non-zero value to indicate true. -57 (a random choice ) is chosen to avoid the more regular values for (int) true.

Your second suggestion is a cross-verification that when the call returns true (as it should for USM) the memory is actually  accessible. IMO it goes slightly beyond the scope of the test, but I am happy to include it.

tob2: > @tmh97 @tob2 The first suggestion assumes true == 1, which (maybe is the norm) but not standard . Bool true macro evaluates to 1 and false to 0, but the return value here is an integer.

I concur – but have a still a minor remark:

I think `true == 1` matches the C standard (C17 in Sect. 7.18) – but I concur that the `omp_target_is_...` functions do not return `true` but the OpenMP-defined _true_, which can be any nonzero integer. Thus, the suggested `true`/`false` alias `1`/`0` check does not make sense – but likewise, previously used `check_test == -57, "... did not return true or false");` suffers from the same issue.

_Side remark: With Fortran, **omp_target_is_accessible** returns an integer (`integer(c_int)` plus `bind(C)`), hence, it does not return `.true.`/`.false.` – but it neither returns_ true _/_ false _. — That's issue now tracked in OpenMP spec issue 3327._

tests/5.1/requires/test_target_is_accessible_with_usm.c
```diff
@@ -0,0 +1,68 @@
...
+  }
+  for(int i = 0;i < N; i++)
+    ptr[i] = i;
+
+  check_test = omp_target_is_accessible(ptr, buf_size, dev);
+
+  OMPVV_TEST_AND_SET_VERBOSE(errors, check_test == 0);
+
+  OMPVV_ERROR_IF(check_test == -57, "Call to omp_target_is_accessible did not return true or false");
+  OMPVV_INFOMSG_IF((check_test != 0 && check_test != -57), "The host memory is accessible from the default device.");
```
tob2, line 40: The first line lines actually check whether the assignment `check_test = omp_target_is_accessible(...)` on the host works – and _true_ is not `-57`.

I think there is no reason to assume that the call to the API function returned to the caller but that the assignment did not work – and as the assignment is not done done by reference but as return value, even a `return;` instead of `return res;` in the function will not keep the original value. — Additionally, -57 matches the OpenMP definition of _true_.

Likewise for the second line. Thus, I would do:

```suggestion
  OMPVV_INFOMSG_IF(check_test != 0, "The host memory is accessible from the default device.");
```

tests/5.1/requires/test_target_is_accessible_with_usm.c
```diff
@@ -0,0 +1,68 @@
...
+  {
+    for(int i = 0;i < N; i++)
+      ptr[i] = ptr[i] * i;
+  }
+
+  for(int i = 0;i < N; i++)
+   OMPVV_ERROR_IF(ptr[i] != i * i, "Operations on the host memory that was reported as accessible produced incorrect results");
+
+}
+
```
tob2, line 53: how about adding:
```C
else
  OMPVV_ERROR ("Host memory was reported inaccessible despite 'requires unified_shared_memory'");
```

tob2 (commented):
LGTM except for the two comments.

tests/5.1/requires/test_target_is_accessible_with_usm.c
spophale: I like it. Thanks @tob2 ! 

Added.

spophale:
@nolanbaker31 can you give it a look?

PR #587, opened 2022-07-14
Assume Directive Tests

nolanbaker31:
Both the begin_end_assumes_contains & assumes_contains tests pass on most recent LLVM. Does not look like they have implemented the "assume" syntax, unless I am using it incorrectly.

I am unsure if my use of the assume_hold makes any sense, LMK.


tests/5.1/assume/test_assume_contains.c
```diff
@@ -0,0 +1,49 @@
...
+#define N 1024
+
+int errors, i;
+
+int test_assume_contains() {
+    int arr[N];
+    for(i = 0; i < N; i++){
+        arr[i] = i;
+    }
+    #pragma omp assume contains(parallel)
```
spophale, line 28: Why not contains(target, parallel, for) ?

nolanbaker31: Good call, originally had them as the combined directive but spec states:

> An encountered construct matches the directive name if it is a combined or composite construct for which a constituent construct has the same name as one of the specified directive names

As well as

>  A directive-name that appears in an absent or contains clause may not be a combined or composite directive name.

So I am unsure whether this means a combined construct would work... I'm assuming this means that if a combined construct is explicity listed in the spec (such as #omp target parallel for), then it is allowed.

I ended up just splitting up just using parallel, but **either way, your recommendation would work**

tests/5.1/assume/test_assume_no_openmp.c
```diff
@@ -0,0 +1,44 @@
...
+#include <math.h>
+
+#define N 1024
+
+int errors, i;
+
+int test_assume_no_openmp() {
+    #pragma omp assume no_openmp
+    {
+        int arr[N];
```
spophale, line 23: It may work, but can you pull the array declaration to before the ump pragma ?

tests/5.1/assume/test_assume_no_parallelism.c
```diff
@@ -0,0 +1,44 @@
...
+    }
+    return errors;
+}
+
+int main() {
+   errors = 0;
+   OMPVV_TEST_OFFLOADING;
+   OMPVV_TEST_AND_SET_VERBOSE(errors, test_assume_no_parallelism() != 0);
+   OMPVV_REPORT_AND_RETURN(errors);
+}
```
spophale, line 44: Do you see that symbol after the bracket ? Noticed it for the for the first time. Can we get rid of it ?

tests/5.1/assume/test_assume_no_parallelism.c
```diff
@@ -0,0 +1,44 @@
...
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+#define N 1024
+
+int errors, i;
+
+int test_assume_no_parallelism() {
+    #pragma omp assume no_parallelism
```
spophale, line 21: I am trying to distinguish the cases where there is no_openmp vs. no_parallelism. How about we add a omp rt routine here and report the value ?

Something like:
int arr[N];

#pragma omp assume no_parallelism
{
         int x = omp_get_thread_num(); // should be 0
         for(i = 0; i < N; i++){
             arr[i] = i + x; 
         }
}

spophale:
Excellent coverage @nolanbaker31 ! 
I think you are using assume_hold correctly. Even if is not a stellar use case, it does the job.

spophale (approved):

tests/5.1/assume/test_assume_no_parallelism.c
spophale: I see it only in the Flies Changed tab.

krishols (approved):
Looks good - I agree that coverage looks great. Logically the use of the assume syntax makes sense to me.

PR #588, opened 2022-07-15
Update Makefile

nolanbaker31:
Often .mod files don't go in the "ompvv" directory. Quick fix to remove these w/ make tidy.

Also, using a slurm bash script creates a slurm-(numbers) file, so adding this to delete these as well.


mjcarr458 (commented):
Looks good to go.

mjcarr458 (approved):

PR #589, opened 2022-07-19
Test scope nowait construct.c

mjcarr458:
Test for new scope construct with nowait clause. Not sure of the best way to approach this, any recommendations of use case appreciated.

### DIFF
--- a/tests/5.1/scope/test_scope_nowait_construct.c
+++ b/tests/5.1/scope/test_scope_nowait_construct.c
@@ -36,7 +36,6 @@ int test_scope_nowait(){
 			}
 		}
 		#pragma omp scope nowait
-		#pragma omp scope
 		{
 			#pragma omp for
 			for (int i = 0; i < N; i++){
--- a/tests/5.1/scope/test_scope_nowait_construct.c
+++ b/tests/5.1/scope/test_scope_nowait_construct.c
@@ -19,13 +19,9 @@
 int test_scope_nowait(){
 	int errors = 0;
 	int test_scalar = 1;
-	for (int i=0; i<N; i++){
-		test_arr[i] = 1;
-		sum += 1;
-	}
 	#pragma omp task shared(test_int)
 	{
-		#pragma omp scope privatee(test_int) nowait
+		#pragma omp scope private(test_int) nowait
 		{
 			test_int += 1;
 		}
--- a/tests/5.1/scope/test_scope_nowait_construct.c
+++ b/tests/5.1/scope/test_scope_nowait_construct.c
@@ -16,38 +16,44 @@
 
 #define N 1024
 
-int a = 1;
-int b = 2;
-int c = 3;
-int d = 4;
-int sum1 = 0;
-int sum2 = 0;
+int test1[N];
+int test2[N];
 
-void test_func1(){
-	sum1 = a+b;
-	return;
-}
-
-void test_func2(){
-	sum2 = c+d;
-	return;
-}
 
 int test_scope_nowait(){
 	int errors = 0;
-	#pragma omp scope
-	{
-		test_func1();
+	for (int i = 0; i <N; i++){
+		test1[i] = 0;
+		test2[i] = 0;
 	}
-	#pragma omp scope nowait
-	#pragma omp scope
+	#pragma omp parallel shared(test1, test2)
 	{
-		test_func2();
+		#pragma omp scope
+		{
+			#pragma omp for
+			for (int i = 0; i < N; i++){
+				test1[i] = 1;
+			}
+		}
+		#pragma omp scope nowait
+		#pragma omp scope
+		{
+			#pragma omp for
+			for (int i = 0; i < N; i++){
+				test2[i] = 1;
+			}
+		}
 	}
-	int total = sum1 + sum2
-	OMPVV_TEST_AND_SET_VERBOSE(errors, total != 10);
-        OMPVV_INFOMSG_IF(sum1 == 0, "first function was not called");
-	OMPVV_INFOMSG_IF(sum2 == 0, "second function was not called");
+	int sum1 = 0;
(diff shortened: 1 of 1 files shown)


tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,39 @@
...
+#define N 1024
+
+int test_scope_nowait(){
+	int errors = 0;
+	int test_scalar = 1;
+	#pragma omp task shared(test_int)
+	{
+		#pragma omp scope private(test_int) nowait
+		{
+			test_int += 1;
```
nolanbaker31, line 26: should be test_scalar

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,39 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+#include <unistd.h>
+
+#define N 1024
+
+int test_scope_nowait(){
+	int errors = 0;
```
nolanbaker31, line 20: errors set to 0 twice (line 35)

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,39 @@
...
+	int test_scalar = 1;
```
nolanbaker31, line 21: should be test_int

nolanbaker31:
Not sure if there's a way to explicitly test nowait, but I would remove the private clause (if private doesn't work properly but nowait does, test would fail)

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,49 @@
...
+#include "ompvv.h"
+#include <math.h>
+#include <unistd.h>
+
+#define N 1024
+
+int test_scope_nowait(){
+	int errors = 0;
+	int test_arr[N];
+	for (int i=0; i<N, i++){
```
tob2, line 22:
```suggestion
	for (int i=0; i<N; i++){
```
(semicolon not colon)

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,49 @@
...
+		test_arr[i] = 1;
+		sum += i;
```
tob2, line 24:
```
test_scope_nowait_construct.c:24:17: error: ‘sum’ undeclared (first use in this function)
   24 |                 sum += i;
```

Note: It is declared in Line 35 – i.e. the decl needs to be moved up.

And please decide whether you want to use `sum` – or `test_int`.

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,49 @@
...
+	}
+	#pragma omp parallel shared(test_int)
```
tob2, line 26:
```
test_scope_nowait_construct.c:26:37: error: ‘test_int’ undeclared (first use in this function)
   26 |         #pragma omp parallel shared(test_int)
```
Should either be `sum` – or `int sum = 0;` should be `int test_int = 0;`

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,49 @@
...
+				test_arr[i] += 1;
+			}
+		}
+	}
+	int sum = 0;
+	for (int i = 0; i < N; i++){
+		sum += test_arr[i];
+	}
+	OMPVV_TEST_AND_SET_VERBOSE(errors, sum != 2048);
+        OMPVV_INFOMSG_IF(test_int == 1024, "scope region was not accessed");
```
tob2, line 40: Same issue as above: `test_int` not declared. – Maybe `test_int` is supposed to be `test_arr[i]` – but then a loop is missing.

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,49 @@
...
+			for (int i=0; i<N; i++){
+				test_arr[i] += 1;
+			}
+		}
+	}
+	int sum = 0;
+	for (int i = 0; i < N; i++){
+		sum += test_arr[i];
+	}
+	OMPVV_TEST_AND_SET_VERBOSE(errors, sum != 2048);
```
tob2, line 39: The expected value is not quite right. In the first loop, it is set to `sum += i` – which is not `N` (1024) but 523776 – should this be `sum += 1` instead?

However, even with the latter, it would not 2*N but 3*N as `test_arr[i] = 1` is followed by `test_arr[i] += 1` – such that `sum += test_arr[i]` adds 2*N – yielding `2*N + <value of sum of the first loop>` (which is either `2*N + 523776` or, after the 'sum += 1' change, `3*N`).

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,59 @@
...
+}
+
+int test_scope_nowait(){
+	int errors = 0;
+	#pragma omp scope
+	{
+		test_func1();
+	}
+	#pragma omp scope nowait
+	#pragma omp scope
```
spophale, line 43: extra?

tests/5.1/scope/test_scope_nowait_construct.c
```diff
@@ -0,0 +1,59 @@
...
+	{
+		test_func2();
+	}
+	int total = sum1 + sum2
```
spophale, line 47: missing ; at the end of line.

spophale:
Change test to use parallel and work sharing loops inside scope.

spophale (approved):

krishols (approved):
Looks good!

PR #590, opened 2022-07-22
Dispatch w/ novariants, nowait, nocontext & is_device_ptr clauses

nolanbaker31:
Getting the simpler clauses out of the way first for dispatch, still need device, depend & is_device_ptr clauses.

GCC 11.2 does not recognize directive, llvm just reports errors for `dispatch w/ ____ false is not working`
I'm not sure if I am using declare variant improperly (with the naming schemes, etc) or its not implemented properly.

### DIFF
--- a/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
+++ b/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
@@ -34,7 +34,7 @@ void add(int *arr){
 }
 
 void add_dev(int *arr){
-    #pragma omp target
+    #pragma omp target is_device_ptr(arr)
     for (int i = 0; i < N; i++){
         arr[i] = arr[i]+4; // Variant function adds 4 to array values
     }
--- a/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
+++ b/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
@@ -63,7 +63,6 @@ int test_wrapper() {
     for(i = 0; i < N; i++){
         OMPVV_TEST_AND_SET(errors, (arr[i] != i+4) && (arr[i] != i+2) );
     }
-    OMPVV_ERROR_IF(errors > 0, "Dispatch w/ novariants true is not working properly");
     OMPVV_INFOMSG_IF(errors > 0 || arr[0] == 2,
                    "Dispatch is either not working or was not considered"
                    " by the implementation as part of the context selector.");
--- a/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
+++ b/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
@@ -55,7 +55,7 @@ int test_wrapper() {
     #pragma omp dispatch is_device_ptr(arr)
         add(arr);
     
-    #pragma omp target
+    #pragma omp target map(tofrom: errors)
     for(i = 0; i < N; i++){
         OMPVV_TEST_AND_SET_VERBOSE(errors, arr[i] != i+4);
         if(i == 5) {
--- a/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
+++ b/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
@@ -25,10 +25,10 @@ int *arr;
 
 void add_dev(int *arr);
 
-#pragma omp declare variant(add_two) match(construct={dispatch}) // NOTE: necessary? append_params(interop: use_device_ptr)
+#pragma omp declare variant(add_dev) match(construct={dispatch}) 
 void add(int *arr){
     #pragma omp for
-    for (int i = 0; i < N; i++){ // Base function adds 1 to array values
+    for (int i = 0; i < N; i++){ // Base function adds 2 to array values
         arr[i] = arr[i]+2;
     }
 }
@@ -36,7 +36,7 @@ void add(int *arr){
 void add_dev(int *arr){
     #pragma omp target
     for (int i = 0; i < N; i++){
-        arr[i] = arr[i]+4; // Variant function adds 2 to array values
+        arr[i] = arr[i]+4; // Variant function adds 4 to array values
     }
 }
 
@@ -57,7 +57,7 @@ int test_wrapper() {
     
     #pragma omp target
     for(i = 0; i < N; i++){
-        OMPVV_TEST_AND_SET_VERBOSE(errors, arr[i] != i+1);
+        OMPVV_TEST_AND_SET_VERBOSE(errors, arr[i] != i+4);
         if(i == 5) {
             OMPVV_ERROR_IF(arr[i] == 5, "No function called or error in mapping");
             OMPVV_ERROR_IF(arr[i] == 7, "Non-target function was called");
--- a/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
+++ b/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
@@ -45,7 +45,11 @@ int test_wrapper() {
     errors = 0;
     t = omp_get_default_device();
     arr = (int *)omp_target_alloc( sizeof(int)*N, t);
-    #pragma omp target
+    if(arr == NULL){
+        OMPVV_ERROR("Can't properly create device pointer");
+        return 1;
+    }
+    #pragma omp target is_device_ptr(arr)
     {
         #pragma omp parallel for
         for(int i = 0; i < N; i++){
@@ -57,11 +61,7 @@ int test_wrapper() {
     
     #pragma omp target map(tofrom: errors)
     for(i = 0; i < N; i++){
-        OMPVV_TEST_AND_SET(errors, arr[i] != i+4);
-        if(i == 5) {
-            OMPVV_ERROR_IF(arr[i] == 5, "No function called or error in mapping");
-            OMPVV_ERROR_IF(arr[i] == 7, "Non-target function was called");
-        }
+        OMPVV_TEST_AND_SET(errors, (arr[i] != i+4) || (arr[i] != i+2) );
     }
     OMPVV_ERROR_IF(errors > 0, "Dispatch w/ novariants true is not working properly");
     return errors;
--- a/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
+++ b/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
@@ -27,7 +27,7 @@ void add_dev(int *arr);
 
 #pragma omp declare variant(add_dev) match(construct={dispatch}) 
 void add(int *arr){
-    #pragma omp for
+    #pragma omp parallel for
     for (int i = 0; i < N; i++){ // Base function adds 2 to array values
         arr[i] = arr[i]+2;
     }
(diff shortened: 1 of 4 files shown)


seyonglee:
Fixed some minor errors in the error checking code and messages, but no compilers on Summit can compile them.

seyonglee (approved):
It looks good to me, even though no compiler can compile on Summit.

spophale:
We are missing a simple dispatch example (no clauses) to show the basic functionality.

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
```diff
@@ -0,0 +1,74 @@
...
+
+int errors;
+int i = 0;
+int *arr;
+
+void add_dev(int *arr);
+
+#pragma omp declare variant(add_dev) match(construct={dispatch}) 
+void add(int *arr){
+    #pragma omp for
```
spophale, line 30: Any particular reason for not doing it in parallel ?

nolanbaker31: No. Will update to use parallel for

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
```diff
@@ -0,0 +1,74 @@
...
+#pragma omp declare variant(add_dev) match(construct={dispatch}) 
+void add(int *arr){
+    #pragma omp parallel for
+    for (int i = 0; i < N; i++){ // Base function adds 2 to array values
+        arr[i] = arr[i]+2;
+    }
+}
+
+void add_dev(int *arr){
+    #pragma omp target
```
nolanbaker31, line 37: add is_device_ptr

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
```diff
@@ -0,0 +1,74 @@
...
+    for (int i = 0; i < N; i++){
+        arr[i] = arr[i]+4; // Variant function adds 4 to array values
+    }
+}
+
+int test_wrapper() { 
+    int t;
+    errors = 0;
+    t = omp_get_default_device();
+    arr = (int *)omp_target_alloc( sizeof(int)*N, t);
```
nolanbaker31, line 47: check if its null; also, find a way to use host fallback

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
```diff
@@ -0,0 +1,74 @@
...
+    #pragma omp target
+    {
+        #pragma omp parallel for
+        for(int i = 0; i < N; i++){
+            arr[i] = i;
+        }
+    }
+    #pragma omp dispatch is_device_ptr(arr)
```
nolanbaker31, line 55: refer to page 54 line 6-7; we should accept both variant functions, use infomsg to print which version was used

andrewkallai:
For test_dispatch_is_device_ptr.c
https://godbolt.org/z/8YPT35fe3
clang 17.0.0 (https://github.com/llvm/llvm-project.git ba7cc56782dbf4a26c0a043dd33c7949366e2b0d)
compiler crashes
https://github.com/llvm/llvm-project/issues/62389
Indication that LLVM developers have not fully implemented dispatch construct, but it is unclear if they are aware that a segmentation fault occurrs when assertion is turned on.

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
```diff
@@ -0,0 +1,74 @@
...
+        add(arr);
+    
+    #pragma omp target map(tofrom: errors)
+    for(i = 0; i < N; i++){
+        OMPVV_TEST_AND_SET(errors, (arr[i] != i+4) || (arr[i] != i+2) );
```
spophale, line 64: Here we can add a INFOMSG here if (arr[i] == i+2) that the dispatch construct in the context-selector was ignored.

tests/5.1/dispatch/test_dispatch_nocontext.c
```diff
@@ -0,0 +1,74 @@
...
+      OMPVV_TEST_AND_SET(errors, arr[i] != i+1);
+   }
+   OMPVV_ERROR_IF(errors > 0, "Dispatch w/ nocontext true is not working properly");
+
+   nocontext_arg = false;
+   #pragma omp dispatch nocontext(nocontext_arg)
+      add(arr);
+
+   for(i = 0; i < N; i++){
+      OMPVV_TEST_AND_SET(errors, arr[i] != i+2);
```
spophale, line 64: Error condition has to change, both values are correct. Provide INFOMSG to indicate if dispatch was considered.

tests/5.1/dispatch/test_dispatch_nocontext.c
```diff
@@ -0,0 +1,74 @@
...
+}
+
+void add_two(int *arr){
+   for (int i = 0; i < N; i++){
+      arr[i] = i+2; // Variant function adds 2 to array values
+   }
+}
+
+int test_wrapper() { 
+   errors = 0;
```
spophale, line 43: Make this an array of size 2.

tests/5.1/dispatch/test_dispatch_nowait.c
```diff
@@ -0,0 +1,62 @@
...
+   errors = 0;
+   add(arr);
+   for(i = 0; i < N; i++){
+      OMPVV_TEST_AND_SET(errors, arr[i] != i+1);
+   } 
+   OMPVV_ERROR_IF(errors > 0, "Base function is not working properly");
+
+   #pragma omp dispatch nowait
+      add(arr);
+
```
spophale, line 50: Add #pragma omp barrier

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
```diff
@@ -0,0 +1,74 @@
...
+    }
+    OMPVV_ERROR_IF(errors > 0, "Dispatch w/ novariants true is not working properly");
```
tob2, line 66:
```suggestion
    OMPVV_ERROR_IF(errors > 0, "Dispatch w/o novariants true is not working properly");
```

tests/5.1/dispatch/test_dispatch_nocontext.c
tob2: Shouldn't this be size 3? That is: no dispatch, dispatch with nocontext(true) and with nocontext(false)?

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
```diff
@@ -0,0 +1,74 @@
...
+int errors;
+int i = 0;
+int *arr;
+
+void add_dev(int *arr);
+
+#pragma omp declare variant(add_dev) match(construct={dispatch}) 
+void add(int *arr){
+    #pragma omp parallel for
+    for (int i = 0; i < N; i++){ // Base function adds 2 to array values
```
tob2, line 31: Maybe not the best wording but you get the idea — likewise for the other add/add_two function comments in this pull request:
```suggestion
    for (int i = 0; i < N; i++){ // Base function sets array values to i+2
```

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
tob2: BTW: It would be good if the INFOMSG is only printed once and not for every of the N=1024 array elements.

Regarding the code: If 'add' is called, i.e. the base function: That's only valid if omp_get_default_device() is the initial device **or** there is unified shared memory.  – Because 'add' is called on the host and 'arr' is allocated on the device with 'omp_target_alloc(...)'.

IMHO: I think that the code requires a comment why both results are valid. In my understanding, it is an oversight in the spec that it is implementation defined which one is invoked – fixed in 6.0, isn't it?
In any case, I think a comment is required here – and in the other files – stating why "== i +2" is okay.

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
```diff
@@ -0,0 +1,74 @@
...
+
+int errors;
+int i = 0;
+int *arr;
+
+void add_dev(int *arr);
+
+#pragma omp declare variant(add_dev) match(construct={dispatch}) 
+void add(int *arr){
+    #pragma omp parallel for
```
tob2, line 30: If the implementation is permitted to call the base function inside the 'dispatch' construct – then the following must  be executed in a **target** region. _See comment further below._
```suggestion
    #pragma omp target parallel for is_device(arr)
```

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
seyonglee: Should `||` be changed to `&&`?

tests/5.1/dispatch/test_dispatch_nocontext.c
```diff
@@ -0,0 +1,79 @@
...
+   }
+   errors += err_ar[0];
+   OMPVV_ERROR_IF(errors > 0, "Dispatch w/ nocontext true is not working properly");
+
+   nocontext_arg = false;
+   #pragma omp dispatch nocontext(nocontext_arg)
+      add(arr);
+
+   for(i = 0; i < N; i++){
+      OMPVV_TEST_AND_SET(err_ar[1], (arr[i] != i+1) || (arr[i] != i+2));
```
seyonglee, line 66: Change `||` to `&&`

tests/5.1/dispatch/test_dispatch_nocontext.c
```diff
@@ -0,0 +1,79 @@
...
+   for(i = 0; i < N; i++){
+      OMPVV_TEST_AND_SET(errors, arr[i] != i+1);
+   } 
+   OMPVV_ERROR_IF(errors > 0, "Base function is not working properly");
+   nocontext_arg = true;
+   #pragma omp dispatch nocontext(nocontext_arg)
+      add(arr);
+   
+   for(i = 0; i < N; i++){
+      OMPVV_TEST_AND_SET(err_ar[0], (arr[i] != i+1) || (arr[i] != i+2));
```
seyonglee, line 56: Change `||` to `&&`

spophale: Don't check for add_two.

tests/5.1/dispatch/test_dispatch_nocontext.c
andrewkallai: If there is no dispatch considered, that is not an error, so there is no need to keep track of it.

tests/5.1/dispatch/test_dispatch_is_device_ptr.c
spophale: @tob2 , yes, you  are correct. This behavior is allowed in 5.1 but will change in 6.0.

tests/5.1/dispatch/test_dispatch_nocontext.c
spophale: Add comment: // OMP 5.1, Pg54:6 -Whether the dispatch construct is added to the construct set is implementation defined.

andrewkallai (approved):

PR #591, opened 2022-07-27
omp_places extended syntax test

mjcarr458:


mjcarr458:
Development guide requested a test for "extended OMP_PLACES syntax", but cannot find any reference to extended syntax in documentation.

tob2:
> Development guide requested a test for "extended OMP_PLACES syntax", but cannot find any reference to extended syntax in documentation.

It is not quite clear what you mean by 'extended syntax' means – a more complex/advanced/extended syntax – or the extensions to the syntax that were added in OpenMP 5.1. In case of the latter:

* Abstract names ll_caches and numa_domains added (besides preexisting threads, cores and sockets)
* In many cases the curly braces can be removed. To quote the related GCC commit:
```
    in particular that in the grammar place non-terminal is now
    not only { res-list } but also res (i.e. a non-negative integer),
    which stands as a shortcut for { res }
    So, one can specify OMP_PLACES=0,4,8,12 with the meaning
    OMP_PLACES={0},{4},{8},{12} or OMP_PLACES=0:4 instead of OMP_PLACES={0}:4
    or OMP_PLACES={0},{1},{2},{3} etc.
```

spophale:
Split into 2 --numa domains and ll_cache.

mjcarr458:
Split into  separate tests

PR #592, opened 2022-07-27
Add new_test/test_requires_reverse_offload.F90

seyonglee:
Update both C and Fortran versions to print a warning if the test cannot detect errors because the target system supports a shared memory.
    - LLVM 15.0.0: C test failed.
    - XL 16.1.1-10: Both C and Fortran tests failed.

### DIFF
--- a/tests/5.0/requires/test_requires_reverse_offload.F90
+++ b/tests/5.0/requires/test_requires_reverse_offload.F90
@@ -39,7 +39,7 @@ PROGRAM test_requires_reverse_offload
      A(i) = i
   END DO
 
-  OMPVV_WARNING_IF(device_num .le. 0, "Cannot properly properly test reverse offload if no devices are available")
+  OMPVV_WARNING_IF(device_num .le. 0, "Cannot properly test reverse offload if no devices are available")
   OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env)
   OMPVV_WARNING_IF(is_shared_env, "[WARNING] May not be able to detect errors if the target system supports shared memory.")
 
--- a/tests/5.0/requires/test_requires_reverse_offload.F90
+++ b/tests/5.0/requires/test_requires_reverse_offload.F90
@@ -26,7 +26,7 @@ PROGRAM test_requires_reverse_offload
 
   INTEGER, DIMENSION(N) :: A
   LOGICAL :: isOffloading
-  INTEGER :: device_num, i
+  INTEGER :: device_num, i, errors2
   LOGICAL :: is_shared_env
 
   OMPVV_TEST_AND_SET_OFFLOADING(isOffloading)
@@ -43,15 +43,19 @@ PROGRAM test_requires_reverse_offload
   OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env)
   OMPVV_WARNING_IF(is_shared_env, "[WARNING] May not be able to detect errors if the target system supports shared memory.")
 
-  !$omp target enter data map(to: A)
+  errors2 = 0
 
-  !$omp target   ! Run on the default device, which is the host for device_num = 0
-    !$omp target device(ancestor: 1) map(always, to: A)
+  !$omp target map(tofrom: errors2) map(to: A, is_shared_env) ! Run on the default device, which is the host for device_num = 0
+    !$omp target device(ancestor: 1) map(to: A)
     DO i = 1, N
        A(i) = 2*i
     END DO
     !$omp end target
+    if ((.NOT. omp_is_initial_device()) .AND. (.NOT. is_shared_env)) then
+      if (any (A /= [(i, i = 1, N)])) errors2 = errors2 + 1
+    end if  
   !$omp end target
+  OMPVV_TEST(errors2 /= 0)
 
   DO i = 1, N
      OMPVV_TEST(A(i) .ne. 2*i)
--- a/tests/5.0/requires/test_requires_reverse_offload.F90
+++ b/tests/5.0/requires/test_requires_reverse_offload.F90
@@ -1,0 +1,61 @@
+!===--- test_requires_reverse_offload.F90 -------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test checks to see that the reverse_offload clause on a requires directive 
+! is supported, and if so, the function host_function() will made available as a 
+! procedure only on the host. By specificying the 'ancestor' modifier with device 
+! number of 1, we are indicating to compiler that execution is to be performed on
+! the immediate parent, the host. The omp declare target statement ensures that the
+! host_function will only be available on host.  
+!
+! Based on OpenMP 5.0 Example: target_reverse_offload.7.c
+! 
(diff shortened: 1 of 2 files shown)


tests/5.0/requires/test_requires_reverse_offload.F90
```diff
@@ -0,0 +1,61 @@
...
+  OMPVV_WARNING_IF(.not. isOffloading, "Without offloading enabled, host execution is already guaranteed")
+
+  device_num = omp_get_num_devices()
+
+  DO i = 1, N
+     A(i) = i
+  END DO
+
+  OMPVV_WARNING_IF(device_num .le. 0, "Cannot properly properly test reverse offload if no devices are available")
+  OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env)
```
tob2, line 43: GCC shows:
```
Warning: Extension: Conversion from LOGICAL(4) to INTEGER(4) at (1)
```
As the macro (or rather the "function test_and_set_shared_environment" it calls), returns a logical, I think the variable should be changed to a logical.

tests/5.0/requires/test_requires_reverse_offload.F90
```diff
@@ -0,0 +1,61 @@
...
+
+  !$omp target enter data map(to: A)
+
+  IF (device_num .gt. 0) THEN
+    !$omp target device(ancestor: 1) map(always, to: A)
+    DO i = 1, N
+       A(i) = 2*i
+    END DO
+    !$omp end target
+  END IF
```
tob2, line 54: This code up to the IF clause is run on the _initial_/host device. The OpenMP spec states for that case:

> If **ancestor** is specified as the _device-modifier_ ... If the construct on which the device clause appears is not encountered in a target region, the current device is treated as the parent device

Thus, this testcase checks that doing reverse offload on the initial device works, by running on the initial device – but it only does so when the _available-devices-var_ consists of more than only the initial/host device.

I think you really want to do the following: replace the 'IF' ... 'END IF' by 'omp target' ... 'omp end target'.

That way, it will do real reverse offloading (if device_num > 0) and additionally checks that host-fallback reverse offload works (if device_num == 0). Namely:

```suggestion
  !$omp target   ! Run on the default device, which is the host for device_num = 0
    !$omp target device(ancestor: 1) map(always, to: A)
    DO i = 1, N
       A(i) = 2*i
    END DO
    !$omp end target
  !$omp end target
```

tests/5.0/requires/test_requires_reverse_offload.c
```diff
@@ -41,6 +42,8 @@ int main()
     }
 
     OMPVV_WARNING_IF(device_num <= 0, "Cannot properly properly test reverse offload if no devices are available");
+    OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env);
+    OMPVV_WARNING_IF(is_shared_env != 0, "[WARNING] May not be able to detect errors if the target system supports shared memory.")
```
tob2, line 46: The same applies for the C testcase as to the Fortran testcase: You need an the reverse-offload/ancestor:1 target construct needs to be inside a target region to be effective.

tob2 (changes requested):
See comments.

I only want to add that GCC mainline will very soon support reverse offloading (host fallback only for now as 'omp requires reverse_offload' in GCC (still) implies available devices = only the host). Soon = patch posted but still under review. — For true reverse offload, a WIP patch exists, sufficient for hello world but still taking a while until it is ready.

tests/5.0/requires/test_requires_reverse_offload.F90
seyonglee: @tob2 Thank you for catching this error.

tests/5.0/requires/test_requires_reverse_offload.F90
```diff
@@ -0,0 +1,61 @@
...
+
+  DO i = 1, N
+     OMPVV_TEST(A[i] .ne. 2*i)
```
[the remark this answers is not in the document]
seyonglee: Updated; thanks!

tests/5.0/requires/test_requires_reverse_offload.F90
spophale: I think it makes sense. @seyonglee thoughts ?

seyonglee: I need some help to correctly understand the expected behavior of the above suggested code change (nested target regions).

OpenMP V5.0, Section 2.12.5 target Construct, page 172, line 20~23 says the following:

> 20: If a device clause in which the ancestor device-modifier appears is present on the target 
21: construct and the device clause expression evaluates to 1, execution of the target region occurs 
22: on the parent device of the enclosing target region. If the target construct is not encountered 
23: in a target region, the current device is treated as the parent device. 

If we apply these texts to the codes in the above suggested change , who is the `target region` in line 21 of the above texts? My interpretation is that it should be the inner target region in line 49 of the above code change.
And who is the `enclosing target region` in line 22 of the above texts? I think it should be the outer target region in line 48 of the above code change. If my interpretation is correct, the inner target region will be executed on the device executing the outer target region. In this case, if the target system has only one target device, the inner target region will be executed on the same target device regardless of whether `ancestor` device-modifier is observed or not. 
If my interpretation is wrong, can anyone explain how to interpret `target region` in line 21 and `enclosing target region` in line 22 of the above texts? For me, the most confusing part is how to interpret `enclosing target region`; is it with respect to the inner target region? Or is it with respect to the outer target region that executes the inner target directive?

seyonglee: A key to correctly interpret the above `OpenMP V5.0, Section 2.12.5 target Construct, page 172, line 20~23` seems to be on the correct interpretation of the `parent device`, which is defined in OpenMP V5.0, Section 1.2.1 Threading Concepts, page 2, line 20~21, as the following:

> 20: **parent device** For a given target region, the device on which the corresponding target 
21:               construct was encountered.

In the above  `OpenMP V5.0, Section 2.12.5 target Construct, page 172, line 20~23`, `target region` in line 21 refers to the inner target region in line 49 of the above code change, `enclosing target region` in line 22 refers to the outer target region in line 48 of the above code change, and `parent device` of the enclosing target region refers to the device on which the outer target construct in line 48 of the above code change was encountered. Therefore, `parent device of the enclosing target region` will be always the host device in the above code change.
Then, with the code change, 1) if there is no target device, both the outer and inner target regions will be executed on the host device. 2) If there is at least one target device, the outer target region will be executed on the default target device, and the inner target region will be executed on the host device.

Please let me know if there is anything wrong in the above interpretation; if there is no error, I will update this PR and  PR #461 as suggested.

tests/5.0/requires/test_requires_reverse_offload.c
seyonglee: Updated.

seyonglee:
[Update on September 11, 2022]
        - LLVM 15.0.0: C test failed.

spophale:
@jrreap will review.

spophale (approved):

krishols (approved):

Issue #595, opened 2022-08-09
sollve_vv/tests/5.0/declare_mapper/test_declare_mapper_target_struct.F90 need to be updated.


juntaoma-intel:
**Describe the bug**
The subroutine "init" needs to have the directive
`!$omp declare target
`added to it.

The OpenMP spec dictates that external references seen inside a TARGET/ENDTARGET region should be treated as if it had the 'DECLARE TARGET' attribute, but that does not guarantee that the attribute will be applied when the actual external reference itself is compiled.   

In particular, please note that Fortran's scoping is on a program-unit basis, not on a per-file basis.  That means there is no information available when SUBROUTINE INIT is ultimately compiled that it had been referenced inside a TARGET/ENDTARGET region.     This behavior would be identical in C/C++ if the two routines were in separate files; the user would have had to add the #pragma to any functions referenced in different files.

**To Reproduce**
```
bash-4.4$ ifx  -fiopenmp -fopenmp-targets=spir64   test_declare_mapper_target_struct.F90 -o test_declare_mapper_target_struct.exe
bash-4.4$ ./test_declare_mapper_target_struct.exe
...
test_declare_mapp  000000000040609B  Unknown               Unknown  Unknown
test_declare_mapp  000000000040527D  Unknown               Unknown  Unknown
libc-2.31.so       00007F55BA942083  __libc_start_main     Unknown  Unknown
test_declare_mapp  000000000040519E  Unknown               Unknown  Unknown
Aborted (core dumped)

```

**Expected behavior**
```
bash-4.4$ diff -u test_declare_mapper_target_struct.F90 test_declare_mapper_target_struct_new.F90
--- test_declare_mapper_target_struct.F90       2022-08-08 16:35:22.901116802 +0800
+++ test_declare_mapper_target_struct_new.F90   2022-08-08 16:36:21.680757971 +0800
@@ -68,5 +68,6 @@
     use my_struct
     type(newvec) :: s

+    !$omp declare target
     s%data = [(i, i = 1, s%len)]
 END SUBROUTINE

bash-4.4$ ifx  -fiopenmp -fopenmp-targets=spir64   test_declare_mapper_target_struct_new.F90 -o test_declare_mapper_target_struct.exe ; ./test_declare_mapper_target_struct.exe
[OMPVV_RESULT test_declare_mapper_target_struct_new.F90] Test passed on the device.

```

**Compiler**

**Accelerator hardware**

PR #596, opened 2022-08-10
fix_issue_595/test_declare_mapper_target_struct.F90

seyonglee:
Fix issue #595 by adding a missing declare target directive to the SUBROUTINE init(s) in the Fortran version.
    - NVHPC 22.5: C test failed but Fortran test passed.
    - LLVM 15.0.0: C test passed.

Related OpenMP spec rules:
- OpenMP V5.0, Section 2.12.5 target Construct, page 173, line 1~3:
    - If a function (C, C++, Fortran) or subroutine (Fortran) is referenced in a target construct then that function or subroutine is treated as if its name had appeared in a to clause on a declare target directive.
- OpenMP V5.0, Section 2.12.7 declare target Directve, page 182, line 10~11:
    - If a function appears in a to clause in the same translation unit in which the definition of the function occurs then a device-specific version of the function is created.
- OpenMP V5.0, Section 2.12.7 declare target Directve, page 184, line 7~8:
    - Any declare target directive with clauses must appear in a specification part of a subroutine subprogram, function subprogram, program or module.

### DIFF
--- a/tests/5.0/declare_mapper/test_declare_mapper_target_struct.F90
+++ b/tests/5.0/declare_mapper/test_declare_mapper_target_struct.F90
@@ -68,5 +68,7 @@ SUBROUTINE init(s)
     use my_struct
     type(newvec) :: s
 
+    !$omp declare target
+
     s%data = [(i, i = 1, s%len)]
 END SUBROUTINE


spophale (approved):

spophale:
@seyonglee we might have to do the same check for other related Fortran tests.

spophale:
@nolanbaker31 , @jrreap , @krishols or @mjcarr458  - can someone do a second review ?

spophale:
@krishols to review.

krishols (approved):
All good to me. Solves issue 595 and I see Swaroop has linked this PR to that issue.

PR #601, opened 2022-08-17
Dispatch Depend test

nolanbaker31:
Compiles with LLVM 16.0.0, but gives 2048 errors. I don't believe the variant function is ever being reached.

I am not sure if my use of **depend** is extensive enough
I am also not sure if I am using variants properly... OpenMP 5.1 examples doc uses it similarly to the way I am doing in this test: https://www.openmp.org/wp-content/uploads/openmp-examples-5.1.pdf#page=458

`void p_vxv(int *v1,int *v2,int *v3,int n);
void t_vxv(int *v1,int *v2,int *v3,int n);

#pragma omp declare variant( p_vxv ) match( construct={parallel} )
#pragma omp declare variant( t_vxv ) match( construct={target} )
void vxv(int *v1,int *v2,int *v3,int n) // base function
{
for (int i= 0; i< n; i++) v3[i] = v1[i] * v2[i];
 }

void p_vxv(int *v1,int *v2,int *v3,int n) // function variant
{
 #pragma omp for
for (int i= 0; i< n; i++) v3[i] = v1[i] * v2[i]*3;
}

#pragma omp declare target
void t_vxv(int *v1,int *v2,int *v3,int n) // function variant
{
 #pragma omp distribute simd
 for (int i= 0; i< n; i++) v3[i] = v1[i] * v2[i]*2;
}
#pragma omp end declare target`

### DIFF
--- a/tests/5.1/dispatch/test_dispatch_depend.c
+++ b/tests/5.1/dispatch/test_dispatch_depend.c
@@ -25,7 +25,7 @@ int a = 0;
 
 void add_a(int *arr);
 
-#pragma omp declare variant(add_two) match(construct={dispatch})
+#pragma omp declare variant(add_a) match(construct={dispatch})
 void add(int *arr){
     for (int i = 0; i < N; i++){ // Base function adds 1 to array values
         arr[i] = i+1;
--- a/tests/5.1/dispatch/test_dispatch_depend.c
+++ b/tests/5.1/dispatch/test_dispatch_depend.c
@@ -18,7 +18,7 @@
 
 #define N 1024
 
-int arr[N]; // implicit map array 
+int arr[N]; 
 int errors;
 int i = 0;
 int a = 0;
@@ -34,7 +34,7 @@ void add(int *arr){
 }
 
 void add_two(int *arr){
-    OMPVV_TEST_AND_SET_VERBOSE(errors, a != 3);
+    OMPVV_TEST_AND_SET(errors, a != 3);
     OMPVV_ERROR_IF(errors > 0, "Depend clause on dispatch directive not working properly");
     for (int i = 0; i < N; i++){
         arr[i] = i+a; // Variant function adds 3 to array values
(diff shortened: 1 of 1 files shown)


andrewkallai:
Frontier
clang version 17.0.0 (https://github.com/llvm/llvm-project.git 8b4a27f410ba42b2e19651ddafc60cd878d0963c)
CRASH
https://clang.llvm.org/docs/OpenMPSupport.html#openmp-5-1-implementation-details
https://github.com/llvm/llvm-project/issues/62389
Indication that LLVM developers have not fully implemented dispatch construct.
https://github.com/llvm/llvm-project/issues/63871
Potential bug may prevent testing any existing implementation.

tests/5.1/dispatch/test_dispatch_depend.c
```diff
@@ -0,0 +1,66 @@
...
+    for(i = 0; i < N; i++){
+        OMPVV_TEST_AND_SET_VERBOSE(errors, arr[i] != i+1);
+    } 
+    OMPVV_ERROR_IF(errors > 0, "Base function is not working properly");
+
+    #pragma omp dispatch depend(in: a)
+        add(arr);
+
+    for(i = 0; i < N; i++){
+        OMPVV_TEST_AND_SET_VERBOSE(errors, arr[i] != i+3);
```
spophale, line 56: check for both functions here.

tests/5.1/dispatch/test_dispatch_depend.c
```diff
@@ -0,0 +1,66 @@
...
+#pragma omp declare variant(add_two) match(construct={dispatch})
+void add(int *arr){
+    for (int i = 0; i < N; i++){ // Base function adds 1 to array values
+        arr[i] = i+1;
+    }
+    a = 3;
+}
+
+void add_two(int *arr){
+    OMPVV_TEST_AND_SET_VERBOSE(errors, a != 3);
```
spophale, line 37: remove verbose

tests/5.1/dispatch/test_dispatch_depend.c
```diff
@@ -0,0 +1,66 @@
...
+    for (int i = 0; i < N; i++){
+        arr[i] = i+a; // Variant function adds 3 to array values
+    }
+}
+
+int test_wrapper() { 
+    errors = 0;
+    add(arr);
+    for(i = 0; i < N; i++){
+        OMPVV_TEST_AND_SET_VERBOSE(errors, arr[i] != i+1);
```
spophale, line 48: remove verbose

tests/5.1/dispatch/test_dispatch_depend.c
andrewkallai, line 37: Remove Verbose

tests/5.1/dispatch/test_dispatch_depend.c
```diff
@@ -0,0 +1,66 @@
...
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <math.h>
+#include "ompvv.h"
+
+#define N 1024
+
+int arr[N]; // implicit map array
```
tob2, line 21: The comment does not really make sense as there is no target region.
```suggestion
int arr[N];
```

tests/5.1/dispatch/test_dispatch_depend.c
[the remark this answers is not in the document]
tob2: _Side remark: The 'depend(in: a)' case needs to call something else but 'add' – as the nowait 'add_two' checks for 'arr[i] == i+1' but sets it to 'arr[i] == i+3'; thus, calling 'add_two' again will fail._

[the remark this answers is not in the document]
spophale: @tob2 , shouldn't the first call to add call add and not the variant add_two ?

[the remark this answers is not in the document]
spophale: Thanks @tob2. I see your concern. The task test is going to replace the base test since we want to ensure that the dependencies are evaluated correctly (and are only considered for the sibling tasks).

andrewkallai (approved):

spophale (approved):

PR #602, opened 2022-08-17
add ompx F90 test code

krishols:
Preliminary test for ompx sentinel. 
Testing `ompx test_nonexistant`. Code performs as expected and seems to ignore `$!ompx`.

### DIFF
--- a/tests/5.2/implementation_defined/test_ompx_free.F90
+++ b/tests/5.2/implementation_defined/test_ompx_free.F90
@@ -20,11 +20,11 @@ PROGRAM test_ompx
 
         OMPVV_TEST_OFFLOADING
 
-        OMPVV_TEST_VERBOSE(test_fixed_ompx() .ne. 2)
+        OMPVV_TEST_VERBOSE(test_free_ompx() .ne. 2)
 
         OMPVV_REPORT_AND_RETURN()
 CONTAINS
-        INTEGER FUNCTION test_fixed_ompx() 
+        INTEGER FUNCTION test_free_ompx() 
                 INTEGER :: i
                 INTEGER :: n
                 INTEGER :: errors
@@ -39,6 +39,6 @@ CONTAINS
                 do n = 1,2
                         errors = errors + ARR_ERR(n)
                 end do
-                test_fixed_ompx = errors
-        END FUNCTION test_fixed_ompx
+                test_free_ompx = errors
+        END FUNCTION test_free_ompx
 END PROGRAM test_ompx
--- a/tests/5.2/implementation_defined/test_ompx_free.F90
+++ b/tests/5.2/implementation_defined/test_ompx_free.F90
@@ -7,7 +7,7 @@
 !
 ! 'ompx' sentinel is reserved for
 ! implementation-defined extensions to
-! fixed source form OpenMP directives.
+! free source form OpenMP directives.
 !===-------------------------------===//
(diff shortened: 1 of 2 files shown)


tob2:
> Preliminary test for ompx sentinel. Unsure if I am understanding ompx functionality correctly

The idea is that a compiler vendor could add a new feature (compiler extension) without colliding with future OpenMP versions or with non-OpenMP compilers. For instance, they could come up with:
   `!$omp unroll_simd partial(2) linear(i:1)`
writing this like that will work – but if the OpenMP committee adds later their own "unroll_simd" directive with a different syntax, this will clash. – On the other hand, it also gets confusing if one vendor uses "!$omp_ext" the other uses "!$fancy" and the third ...

Thus, having a common prefix helps to get this in order. The expectation is that compilers that do not implement that feature will silently (or with some warning) ignore '!$ompx' or '#pragma ompx' or '[[ ompx :: ...]]'.

I am not sure what's the best way to test this — adding '!$ompx this_is_a_non_existing_vendor_directive'?

krishols:
> Ah, perfect timing! This explanation makes a lot of help, thanks. I talked to Swaroop and she suggested the same change as you, so I've implemented that.

tests/5.2/implementation_defined/test_ompx_free.F90
```diff
@@ -0,0 +1,40 @@
...
+        OMPVV_REPORT_AND_RETURN()
+CONTAINS
+        INTEGER FUNCTION test_fixed_ompx() 
+                INTEGER :: i
+                INTEGER :: errors
+                INTEGER, DIMENSION(10) :: A
+                INTEGER, DIMENSION(10) :: B
+                INTEGER, DIMENSION(10) :: D
+                errors = 0
+                !$omp parallel num_threads(2)
```
krishols, line 34: array of two, update index

tests/5.2/implementation_defined/test_ompx_free.F90
```diff
@@ -0,0 +1,40 @@
...
+                !$ompx test_nonexistant
+                        errors = errors + 1
```
spophale, line 36: make it an array.

tob2:
FYI – I think this testcase if fine. 

_And as side node – and only for your interest and for for completeness, I want to add what GCC does:_

From https://gcc.gnu.org/onlinedocs/libgomp/OpenMP-5_002e2.html
> The `ompx` sentinel as C/C++ pragma and C++ attributes are warned for with `-Wunknown-pragmas` (implied by `-Wall`) and `-Wattributes` (enabled by default), respectively; for Fortran free-source code, there is a warning enabled by default and, for fixed-source code, the `omx` sentinel is warned for with with `-Wsurprising` (enabled by `-Wall`). Unknown clauses are always rejected with an error.

_This just documents (pre)existing behavior, only the fixed-form Fortran diagnostic is new in GCC 13/mainline. For free-form Fortran, the diagnostic is the warning that only `!$omp<space>` is valid as first line and without space, it is only valid as continuation lines. This warning is then also shown for `!$ompx` – as it could also indicate a missing `&` on the previous line or a missing space for the first line: `Warning: !$OMP at (1) starts a commented line as it neither is followed by a space nor is a continuation line`_

spophale (approved):

mjcarr458 (approved):

PR #606, opened 2022-08-24
tests/5.1/assume/test_assume_no_parallelism.c: Fix var scope issue

tob2:
In my understanding, variables declare inside the curly braces of
```C
#pragma omp assume ...
   { ... }
```
are not visible outside. Hence, the 'arr' and 'x' need to be declared before the assume construct.

_Side remark: While my older clang version errors out because of the unsupported 'assume' ("error: expected an OpenMP directive"), GCC only warns ("warning: ignoring ‘#pragma omp assume’ [-Wunknown-pragmas]")_

### DIFF
--- a/tests/5.1/assume/test_assume_holds.c
+++ b/tests/5.1/assume/test_assume_holds.c
@@ -19,9 +19,9 @@ int errors, i, N;
 
 int test_assume_holds() {
     N = 1024;
+    int arr[N];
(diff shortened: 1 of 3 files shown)


spophale (approved):
Thanks @tob2 .

nolanbaker31 (approved):

PR #608, opened 2022-08-24
test_get_mapped_ptr.c: Add cast to avoid conversion error

tob2:
With C++, it failed without this commit with:
```
error: invalid conversion from ‘void*’ to ‘int*’
```
It does compile with a C compiler.


spophale (approved):
LGTM

tob2:
PING - anyone who wants to add a second review?

tmh97 (approved):
This is good-to-go

PR #610, opened 2022-08-31
Fel01

fel-cab:
omp_set_num_teams runtime test added


sys/scripts/createSummary.py
```diff
@@ -178,7 +178,7 @@ def parseFile(log_file):
       current_state = "END"
       current_buffer = ""
       current_test = testResult()
-      for line in open(log_file,'r'):
+      for line in open(log_file,'r', encoding='utf-8', errors='ignore'):
```
tob2, line 181: I think you want to re-diff. This change has been already merged via commit 4c62ffca98d1929995139b335b6d204a385d9242 (Pull Request #609).

tests/5.1/runtime_calls/test_omp_set_num_teams.c
```diff
@@ -0,0 +1,63 @@
...
+
+  omp_set_num_teams(OMPVV_NUM_TEAMS_DEVICE);
+  #pragma omp teams
+  {
+    num_teams[omp_get_team_num()] = omp_get_num_teams();
+  }
+
+  if (num_teams[0] == 1) {
+    OMPVV_WARNING("Test operated with 1 team. num_teams requested were %d.", OMPVV_NUM_TEAMS_DEVICE);
+  } else if (num_teams[0] < 1) {
```
tob2, line 37: I wonder whether there shouldn't be additionally a check for `num_teams[0] > OMPVV_NUM_TEAMS_DEVICE`. That's extremely unlikely, but maybe it should be still added for completeness?

tests/5.1/runtime_calls/test_omp_set_num_teams.c
```diff
@@ -0,0 +1,63 @@
...
+}
+
+
+int main() {
+
+  //OMPVV_TEST_OFFLOADING;
+
+  int errors = 0;
+  //OMPVV_TEST_AND_SET_VERBOSE(errors, test_omp_set_num_teams() != 0);
+  test_omp_set_num_teams();
```
tob2, line 59: The `error` variable in 'main' is currently unused.

_I think it is more in line with the other testcases to end `test_omp_set_num_teams` with `return errors;` – remove the `OMPVV_REPORT_AND_RETURN` in that function and uncomment the OMPVV_ macros in main._

tests/5.1/runtime_calls/test_omp_set_num_teams.c
fel-cab: Agree, done!

spophale:
@fel-cab , this PR is pending on a few changes. The createSummary.py needs to be removed from this PR as the changes there are already in.

tests/5.1/runtime_calls/test_omp_set_num_teams.c
```diff
@@ -0,0 +1,64 @@
...
+{
+  int num_teams[OMPVV_NUM_TEAMS_DEVICE];
+  int errors = 0;
+  
+
+  for (int x = 0; x < OMPVV_NUM_TEAMS_DEVICE; ++x) {
+    num_teams[x] = -99;
+  }
+
+  omp_set_num_teams(OMPVV_NUM_TEAMS_DEVICE);
```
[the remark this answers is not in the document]
fel-cab: You are right, moreover, this test seems to be covered by https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.1/runtime_calls/test_teams_region_routines.c
So we probably can drop it

PR #611, opened 2022-09-02
first omp_places cache test iteration

mjcarr458:


mjcarr458:
Here is a very basic test that just sets the value of omp_places to ll_caches. I can't think of a way to actually confirm the cores are sharing a common cache as implied in the spec, particularly making it system agnostic as well. If anyone has a idea on how to implement that, it would be greatly appreciated.

tests/5.1/env_var/test_omp_places_env_ll_caches.c
```diff
@@ -0,0 +1,23 @@
+#include <omp.h>
```
nolanbaker31, line 1: Needs header w/ file name & test description https://crpl.cis.udel.edu/ompvvsollve/documentation/codingstandards/

spophale (approved):

jrreap (approved):
Looks good to go!

tests/5.1/env_var/test_omp_places_env_ll_caches.c
```diff
@@ -0,0 +1,33 @@
+//-----------------test_omp_places_ll_caches.c---------------//
```
krishols, line 1: Very minor, but test name does not match file name

PR #615, opened 2022-09-08
test_teams_region_routines test added

fel-cab:


tests/5.1/runtime_calls/test_teams_region_routines.c
```diff
@@ -0,0 +1,88 @@
...
+#include <omp.h>
+#include "ompvv.h"
+
+
+int test_teams_region_routines()
+{
+  int num_teams[OMPVV_NUM_TEAMS_DEVICE];
+  int num_threads[OMPVV_NUM_THREADS_DEVICE];
+  int errors[2] = {0,0};
+  int is_offloading;
```
tob2, line 27: (This variable is unused.)

tests/5.1/runtime_calls/test_teams_region_routines.c
```diff
@@ -0,0 +1,88 @@
...
+  OMPVV_TEST_AND_SET_VERBOSE(errors[0], omp_get_max_teams() != OMPVV_NUM_TEAMS_DEVICE);
+
+  //thread error if the max teams doesn't get the set value
+  OMPVV_TEST_AND_SET_VERBOSE(errors[1], omp_get_teams_thread_limit() != OMPVV_NUM_THREADS_DEVICE);
+
+  // The test will also check if the correct team and thread values are passed
+#pragma omp teams 
+  {
+    num_teams[omp_get_team_num()] = omp_get_num_teams();
+#pragma omp parallel master
```
spophale, line 52: change master to "masked".

fel-cab: When I make this change, I don't get the test to pass, because
in line 53: 
num_threads[omp_get_team_num()]= omp_get_num_threads();
do not get the value of the number of threads
Is it ok if we write it as:
    #pragma omp parallel
    if (omp_get_thread_num() == 0)
      num_threads[omp_get_team_num()]= omp_get_num_threads();

spophale: Which compiler are you testing with ? Maybe they do not have "masked" supported yet. Though in that case, I would think, there would be a compile time error. 

If while testing with master you are getting the correct answer, just replace with masked as it should be supported in Spec 5.1.

tests/5.1/runtime_calls/test_teams_region_routines.c
```diff
@@ -0,0 +1,96 @@
...
+
+  //thread error if the max teams doesn't get the set value
+  OMPVV_TEST_AND_SET_VERBOSE(errors[1], omp_get_teams_thread_limit() != OMPVV_NUM_THREADS_DEVICE);
+
+  // The test check if the correct team and thread numbers were created
+  
+  #pragma omp teams 
+  {
+    num_teams[omp_get_team_num()] = omp_get_num_teams();
+    #pragma omp parallel
```
spophale, line 60: needs a single, masked or if.

tests/5.1/runtime_calls/test_teams_region_routines.c
tob2: > Which compiler are you testing with ? Maybe they do not have "masked" supported yet. Though in that case, I would think, there would be a compile time error.

Two remarks: For GCC, 'masked' is supported since GCC 12 (released this spring). And regarding unknown pragmas, C compilers tend to simply ignore them – and only warn for them either by default (like with Clang) or only if some warning flag is specified (such as -Wunknown-pragmas, implies by -Wall, in case of GCC). – Otherwise, the variant as commited compiles and shows "Test passed on the host." with GCC 13/mainline.

tests/5.1/runtime_calls/test_teams_region_routines.c
```diff
@@ -0,0 +1,96 @@
+//===------------ test_teams_region_routines.c ----------------------------===//
+//
+// OpenMP API Version 5.1 Aug 2021
+//
+// Based on https://www.openmp.org/wp-content/uploads/OpenMP-API-Specification-5-1.pdf
```
spophale, line 5: Link not required. We do not do that for the other tests.

spophale (approved):

spophale:
@nolanbaker31 will review.

PR #616, opened 2022-09-09
Add omx fixed-source test

krishols:
Help wanted - my first time writing fixed source form Fortran. I tried to follow these [guidelines](https://people.cs.vt.edu/~asandu/Courses/MTU/CS2911/fortran_notes/node4.html#:~:text=For%20back%2Dcompatibility,the%20first%20column.) but am not receiving any output when I attempt to run the test.

### DIFF
--- a/tests/5.2/implementation_defined/test_omx_fixed.f
+++ b/tests/5.2/implementation_defined/test_omx_fixed.f
@@ -1,0 +1,44 @@
+C===-test_ompx.F90-===/
+C
+C
+C OpenMP API Version 5.2 July 2022
+C
+C Testing 'omx' sentinel
+C
+C 'omx' sentinel is reserved for
+C implementation-defined extensions to
+C fixed source form OpenMP directives.
+C===-------------------------------===//
+        #include "ompvv.F90"
+
+        #define N 1024
+        PROGRAM test_omx
+                use iso_fortran_env
+                use ompvv_lib
+                use omp_lib
+                implicit none
+
+                OMPVV_TEST_OFFLOADING
+
+                OMPVV_TEST_VERBOSE(test_fixed_omx() .ne. 2)
+
+                OMPVV_REPORT_AND_RETURN()
+        CONTAINS
+                INTEGER FUNCTION test_fixed_omx() 
+                        INTEGER :: i
+                        INTEGER :: n
+                        INTEGER :: errors
+                        INTEGER, DIMENSION(2) :: ARR_ERR
+                        errors = 0
+                        !$omp parallel shared(ARR_ERR) private(i)  num_threads(2)
+                        !$omx test_nonexistant_omx
+                                i = omp_get_thread_num()
+                                i = i + 1
+                                ARR_ERR(i) = 1
+                        !$omp end parallel
+                        do n = 1,2
+                                errors = errors + ARR_ERR(n)
+                        end do
+                        test_fixed_omx = errors
+                END FUNCTION test_fixed_omx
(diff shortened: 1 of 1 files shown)


tob2:
As this is the first fixed-source testcase, I think there are a lot of problems:
* File name: A common handling of Fortran file suffixes is that lower case is without C pre-processor while capital letters imply doing preprocessing. Thus, you should change `.f` to `.F`
* Accordingly, also the Makefile only searches for `grep "\(\.F90\|\.F95\|\.F03\|\.F\|\.FOR\)$$")` – i.e. assuming .F or .FOR, while F is more common.
* '#include <ompvv.F90>' does not work – the file is written to only work as free-form file. It can be rewritten to support fixed form, but that needs to be done. Alternatively, a ompvv.F has to be written. – I thin the former is simpler, but still requires some changes in formatting  in particular, using only 72 columns for code, placing for continuation lines the tailing '&' in column > 72 (for free from) and adding a '&' in the continued line (for fixed form, but also permitted for free from). Comments may not start in column 2 to 6 and '!' has to be used as comment character (not 'C').
* For GCC, '#include' and '#defined' must be in the first column - or at least the '#' must (then spaces can come before the 'include'/'define'). I think that's because GCC runs cpp in traditional mode. – As the preprocessor is run before the Fortran compiler, the indentation rules only apply for the result after preprocessing.

Good luck!

krishols:
> As this is the first fixed-source testcase, I think there are a lot of problems:
> 
> * File name: A common handling of Fortran file suffixes is that lower case is without C pre-processor while capital letters imply doing preprocessing. Thus, you should change `.f` to `.F`
> * Accordingly, also the Makefile only searches for `grep "\(\.F90\|\.F95\|\.F03\|\.F\|\.FOR\)$$")` – i.e. assuming .F or .FOR, while F is more common.
> * '#include <ompvv.F90>' does not work – the file is written to only work as free-form file. It can be rewritten to support fixed form, but that needs to be done. Alternatively, a ompvv.F has to be written. – I thin the former is simpler, but still requires some changes in formatting  in particular, using only 72 columns for code, placing for continuation lines the tailing '&' in column > 72 (for free from) and adding a '&' in the continued line (for fixed form, but also permitted for free from). Comments may not start in column 2 to 6 and '!' has to be used as comment character (not 'C').
> * For GCC, '#include' and '#defined' must be in the first column - or at least the '#' must (then spaces can come before the 'include'/'define'). I think that's because GCC runs cpp in traditional mode. – As the preprocessor is run before the Fortran compiler, the indentation rules only apply for the result after preprocessing.
> 
> Good luck!

Great - thanks for the insight and guidance! 
I will: 
- change .f --> .F
- edit include, define lines
Then circle back to see how framework changes to support fixed-form files are coming along

tests/5.2/implementation_defined/test_omx_fixed.F
```diff
@@ -0,0 +1,44 @@
...
+C
+C OpenMP API Version 5.2 July 2022
+C
+C Testing 'omx' sentinel
+C
+C 'omx' sentinel is reserved for
+C implementation-defined extensions to
+C fixed source form OpenMP directives.
+C===-------------------------------===//
+#        include "ompvv.F90"
```
spophale, line 12: Check if this is correct!

fel-cab: I have fixed this

spophale (approved):

fel-cab (approved):

PR #617, opened 2022-09-12
Add new_test/test_declare_target_device_type_any.F90

seyonglee:
- NVHPC 22.5: Both C and Fortran tests passed, but Fortran test ran on the host.
        - GCC 12.1.1: Both C and Fortran tests passed, but Fotran test ran on the host.
        - XL 16.1.1-10: C test passed but Fortran test failed.
            "line 32.29: 1514-480 (S) The entity update that appears in the DECLARE TARGET directive must be declared in the same scoping unit."
        - LLVM 15.0.0: C test passed.

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_device_type_any.F90
+++ b/tests/5.0/declare_target/test_declare_target_device_type_any.F90
@@ -1,0 +1,82 @@
+!===--- test_declare_target_device_type_any.F90 -----------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! The declare target directive specifies that variables, functions(C,C++ and Fortran),
+! and subroutines (Fortran) are mapped to a device. If a device_type
+! clause is present on the contained declare target directive, then its argument 
+! determines which versions are made available. When device_type(any) is specified 
+! both device and host versions of the procedure are made available.
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
+
+MODULE m_dat
+    !$omp declare target to(a, b, c, i)
+    INTEGER, DIMENSION(N) :: a, b, c
+    INTEGER :: i = 0
+END MODULE m_dat
+
+PROGRAM test_declare_target_device_type_any
+
+    USE iso_fortran_env
+    USE ompvv_lib
+    USE omp_lib
+    USE m_dat
+    implicit none
+    INTEGER :: errors = 0
+    
+    !$omp declare target to(update) device_type(any)
+    
+    ! initialize arrays on host
+    DO i = 1, N
+      a(i) = i
+      b(i) = i + 1
+      c(i) = i + 2
+    END DO
+     
+    OMPVV_TEST_VERBOSE(test_device_type_any() .ne. 0)
+
+    OMPVV_REPORT_AND_RETURN()
+
+CONTAINS  
+    SUBROUTINE update()
+      DO i = 1, N
+        a(i) = a(i) + 1
+        b(i) = b(i) + 2
+        c(i) = c(i) + 3
+      END DO
+    END SUBROUTINE update
+
+    INTEGER FUNCTION test_device_type_any() 
+      !$omp target update to(a, b, c)
+      !$omp target
+      CALL update()
+      !$omp end target
+
+      !$omp target update from(a, b, c)
+
+      ! check array values on host
+      DO i = 1, N
+        IF ( a(i) .NE. i+1 .OR. b(i) .NE. i+3 .OR. c(i) .NE. i+5  ) THEN
+          errors = errors + 1
+        END IF
+      END DO
+     
+      ! on host
+      CALL update()
+
+      ! check array values on host
+      DO i = 1, N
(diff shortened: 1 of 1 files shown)


tests/5.0/declare_target/test_declare_target_device_type_any.F90
```diff
@@ -0,0 +1,82 @@
...
+PROGRAM test_declare_target_device_type_any
+
+    USE iso_fortran_env
+    USE ompvv_lib
+    USE omp_lib
+    USE m_dat
+    implicit none
+    INTEGER :: errors = 0
+    
+    !$omp declare target to(update) device_type(any)
```
seyonglee, line 32: XL 16.1.1-10 Fortran compiler fails with the following error message: 
"/gpfs/alpine/csc383/scratch/f6l/sollve/tests/5.0/declare_target/test_declare_target_device_type_any.F90", line 32.29: 1514-480 (S) The entity update that appears in the DECLARE TARGET directive must be declared in the same scoping unit.
Does it mean that we should add a declaration of the internal procedure to be used for `declare target` directive? 
Or is it a bug of the XL Fortran compiler? (Other tested OpenMP Fortran compilers (GCC, NVHPC) do not complain this.)

seyonglee: OpenMP V5.0, Section 2.12.7 declare target Directve, page 183, line 24~26 says the following (OpenMP V5.2 also has the same texts):

This rule applies to our case since `update` is an internal procedure, which should not contain any declare target directive because of `!$omp declare target to(update) device_type(any)` in the main program unit.

tob2: Cf. discussion in Issues #618/#619 about this. I believe the latter does not apply and the '!$omp declare target to(update) device_type(any)' should be inside 'update' itself, following the OpenMP 5.1 bug fix/clarification that added a **that** for the **to**/**enter** clause (→ aforementioned issues).

seyonglee: Moved `!$omp declare target to(update) device_type(any)` to `update`, as suggested.

seyonglee:
[Update on Oct. 30, 2022]
- Move `!$omp declare target to(update) device_type(any)` to `update`
        - NVHPC 22.5: Both C and Fortran tests passed, but Fortran test ran on the host.
        - GCC 12.1.1: Both C and Fortran tests passed, but Fotran test ran on the host.
        - XL 16.1.1-10: Both C and Fortran tests passed, but Fortran test ran on the host.
        - LLVM 15.0.0: C test passed.

spophale:
set OMP_TARGET_OFFLOAD=MANDATORY and see if Fortran tests will run on device.

seyonglee:
[Update on Nov 27, 2022]
       - Use "export OMP_TARGET_OFFLOAD=MANDATORY"
        - NVHPC 22.5: Both C and Fortran tests passed, but Fortran test ran on the host (without any runtime error).
        - GCC 12.2.1: Both C and Fortran tests passed, but Fotran test ran on the host (without any runtime error).
        - XL 16.1.1-10: Both C and Fortran tests passed, but Fortran test ran on the host (without any runtime error).

seyonglee:
OpenMP Spec V5.0 Section "6.17 OMP_TARGET_OFFLOAD" (page 616, line 1~3) says "The MANDATORY value specifies that program execution is terminated if a device construct or device memory routine is encountered and the device is not available or is not supported by the implementation." 
Even with the above requirement, it is still OK that this test program does not exit with an error, since the `MANDATORY` value checks only the availability of a device in  a target system and whether the implementation supports the device or not.

mjcarr458 (approved):
Looks good

tob2:
> OpenMP Spec V5.0 Section "6.17 OMP_TARGET_OFFLOAD" (page 616, line 1~3) says "The MANDATORY value specifies that program execution is terminated if a device construct or device memory routine is encountered and the device is not available or is not supported by the implementation." Even with the above requirement, it is still OK that this test program does not exit with an error, since the `MANDATORY` value checks only the availability of a device in a target system and whether the implementation supports the device or not.

There was a lengthy discussion at the OpenMP specification meetings about OMP_TARGET_OFFLOAD=mandatory and what it should do if there is no device available. (Error termination vs. continuing to run with host fallback.)

The wished-for behavior is clear: If there is a device and should be used but isn't: fail. If there is no device and I wanted to run it with offloading, fail as well - but if there is no offload device and I am aware of this, don't fail. — Besides the problem about personal intent, the exact behavior also had to be codified. That happened after some lengthier discussions. Questions during the discussion included what should happen if `if(0)` or `device(omp_get_initial_device())` or `omp_set_default_device(omp_get_initial_device())` or ... had been used. In any case:

OpenMP 5.2 clarified this by adding a new term (_conforming device number_), two constants **omp_invalid_device** and **omp_initial_device**, some wording change here and there – and stating that the default-device-var ICV gets initialized to omp_invalid_device if OMP_TARGET_OFFLOAD=mandatory was specified but no non-host device is available.

(Side remark: The terms _accessible devices_, _supported devices_ and _available devices_ get more properly defined in Technical Report 11.)

spophale (approved):

PR #618, opened 2022-09-12
Add new_test/test_declare_target_device_type_host.F90

seyonglee:
- NVHPC 22.5: Both C and Fortran tests passed.
        - LLVM 15.0.0: C test passed.
        - GCC 12.1.1: Both C and Fortran tests passed.
        - XL 16.1.1-10: C test passed but Fortran test failed.
            "line 32.29: 1514-480 (S) The entity update that appears in the DECLARE TARGET directive must be declared in the same scoping unit."

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_device_type_host.F90
+++ b/tests/5.0/declare_target/test_declare_target_device_type_host.F90
@@ -1,0 +1,79 @@
+!===--- test_declare_target_device_type_host.F90 -----------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! The declare target directive specifies that variables, functions(C,C++ and Fortran),
+! and subroutines (Fortran) are mapped to a device. If a device_type
+! clause is present on the contained declare target directive, then its argument 
+! determines which versions are made available. If device_type(host) is present 
+! only a host version of the procedure is made available. 
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
+
+MODULE m_dat
+    !$omp declare target to(a, b, c, i)
+    INTEGER, DIMENSION(N) :: a, b, c
+    INTEGER :: i = 0
+END MODULE m_dat
+
+PROGRAM test_declare_target_device_type_host
+
+    USE iso_fortran_env
+    USE ompvv_lib
+    USE omp_lib
+    USE m_dat
+    implicit none
+    INTEGER :: errors = 0
+    
+    !$omp declare target to(update) device_type(host)
+    
+    ! initialize arrays on host
+    DO i = 1, N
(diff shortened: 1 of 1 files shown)


seyonglee:
This has the same issue (IBM XL compiler error) as #617

tests/5.0/declare_target/test_declare_target_device_type_host.F90
```diff
@@ -0,0 +1,79 @@
...
+PROGRAM test_declare_target_device_type_host
+
+    USE iso_fortran_env
+    USE ompvv_lib
+    USE omp_lib
+    USE m_dat
+    implicit none
+    INTEGER :: errors = 0
+    
+    !$omp declare target to(update) device_type(host)
```
tob2, line 32: I think the same applies as I wrote in Pull Request #619 (associated nohost variant). Namely: OpenMP 5.1/5.2 contain a change in the restriction wording, which can be interpreted as bug fix/corrected wording intended already to apply already to 5.0. Namely OpenMP 5.2 (identical to 5.1?) has:

>    If a list item is a procedure name, the directive must be in the specification part of _**that**_ subroutine or function subprogram or in the specification part of that subroutine or function in an interface body.

Note the "that". Thus, I would move it down to the specification part of 'subroutine update'.

seyonglee: As I said in #619, it seems that we should not put the declare target directive in the internal subroutine due to the following restriction (both OpenMP V5.0 and V5.2 have the same texts):

tob2: And as I said in #619, that I believe it does not apply – this restriction does not state that '!$omp declare target ... device_type(...)' may not be in internal procedures – it only states that it may not ***if*** the parent has a 'declare target ... device_type(...)'.

But as the 'declare target' only applies to the internal procedure, it just can be moved there. – That way, there is no declare target in the parent and, additionally, the 5.{1,2} restriction (to be regarded as bug fix) is also fulfilled. _The latter is violated, unless the line is moved to the internal procedure due to the **that**._

seyonglee: Move `!$omp declare target to(update) device_type(host)` into the internal procedure, `update`, as suggested.

seyonglee:
[Update on Oct. 30, 2022]
Move `!$omp declare target to(update) device_type(host)` into the internal procedure, `update`, as suggested.
        - NVHPC 22.5: Both C and Fortran tests passed.
        - LLVM 15.0.0: C test passed.
        - GCC 12.1.1: Both C and Fortran tests passed.
        - XL 16.1.1-10: Both C and Fortran tests passed.

krishols (approved):
Looks good to me!

PR #619, opened 2022-09-13
Add new_test/test_declare_target_device_type_nohost.F90

seyonglee:
- LLVM 15.0.0: C test passed.
        - GCC 12.1.1: C test passed, but Fortran test failed.
            "undefined reference to `update.1'"
        - XL 16.1.1-10: C test passed (ran on the host) but Fortran test failed.
            "line 33.29: 1514-480 (S) The entity target_function that appears in the DECLARE TARGET directive must be declared in the same scoping unit."
            "line 58.13: 1515-019 (S) Syntax is incorrect."
        - NVHPC 22.5:
            - C test failed.
                "line 54: error: this kind of pragma may not be used here
                #pragma omp target "
            - Fortran test failed
                "nvlink error   : Undefined reference to '_update_' in '/tmp/nvfortranXxzgotnDRs3Mi.o'"

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
+++ b/tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
@@ -1,0 +1,97 @@
+!===--- test_declare_target_device_type_nohost.F90 -----------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! The declare target directive specifies that variables, functions(C,C++ and Fortran),
+! and subroutines (Fortran) are mapped to a device. If a device_type
+! clause is present on the contained declare target directive, then its argument 
+! determines which versions are made available. If device_type(nohost) is present 
+! only a device version of the procedure is made available. 
+! The device version of the function is specified via declare variant.
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 10
+
+MODULE m_dat
+    !$omp declare target to(a, b, c, i)
+    INTEGER, DIMENSION(N) :: a, b, c
+    INTEGER :: i = 0
+END MODULE m_dat
+
+PROGRAM test_declare_target_device_type_nohost
+
+    USE iso_fortran_env
+    USE ompvv_lib
+    USE omp_lib
+    USE m_dat
+    implicit none
+    INTEGER :: errors = 0
+    
+    !$omp declare target to(target_function) device_type(nohost)
+    
+    OMPVV_TEST_OFFLOADING
+
+    ! initialize arrays on host
+    DO i = 1, N
+      a(i) = 0
+      b(i) = 0
+      c(i) = 0
+    END DO
+     
+    OMPVV_TEST_VERBOSE(test_device_type_nohost() .ne. 0)
+
+    OMPVV_REPORT_AND_RETURN()
+
+CONTAINS  
+    SUBROUTINE target_function()
+      DO i = 1, N
+        a(i) = a(i) + 1
+        b(i) = b(i) + 2
+        c(i) = c(i) + 3
+      END DO
+    END SUBROUTINE target_function
+
+    SUBROUTINE update()
+      !$omp declare variant(target_function) match(device={kind(nohost)})
+      DO i = 1, N
+        a(i) = a(i) + 5
+        b(i) = b(i) + 5
+        c(i) = c(i) + 5
+      END DO
+    END SUBROUTINE update
+
+    INTEGER FUNCTION test_device_type_nohost() 
+      CALL update()
+
+      !$omp target update to(a, b, c)
+
+      !$omp target
+      CALL update()
+      !$omp end target
+
+      !$omp target update from(a, b, c)
+
+      IF ( omp_get_default_device() .GE. 0 .AND. omp_get_default_device() .LT. omp_get_num_devices() ) THEN
+        ! check array values on host
+        DO i = 1, N
+          IF ( a(i) .NE. 6 .OR. b(i) .NE. 7 .OR. c(i) .NE. 8  ) THEN
+            errors = errors + 1
+          END IF
+        END DO
+      ELSE
+        OMPVV_WARNING("Default device is the host device. Thus, test only ran on the host")
+        ! check array values on host
+        DO i = 1, N
+          IF ( a(i) .NE. 10 .OR. b(i) .NE. 10 .OR. c(i) .NE. 10  ) THEN
+            errors = errors + 1
+          END IF
+        END DO
+      END IF
+     
+      test_device_type_nohost= errors
+    END FUNCTION test_device_type_nohost
+
+END PROGRAM test_declare_target_device_type_nohost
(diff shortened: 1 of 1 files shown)


tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
```diff
@@ -0,0 +1,97 @@
...
+PROGRAM test_declare_target_device_type_nohost
+
+    USE iso_fortran_env
+    USE ompvv_lib
+    USE omp_lib
+    USE m_dat
+    implicit none
+    INTEGER :: errors = 0
+    
+    !$omp declare target to(target_function) device_type(nohost)
```
tob2, line 33: I think the IBM XL compiler is rightly complaining about the program – but with the wrong message. The spec states (5.2 wording, I think 5.1 is identical):
> If a list item is a procedure name, the directive must be in the specification part of _**that**_ subroutine or function subprogram or in the specification part of that subroutine or function in an interface body.

Note the "that" that I highlighted. In your case. Your "declare target to(target_function)" is comes after USE/IMPORT/IMPLICIT (→ restriction is hidden in the glossary) and inside a specification part – but only in the specification part of "PROGRAM test_declare_target_device_type_nohost" and not in the specification part of "target_function".

Note that the restrictions have quite a bit changed in 5.1 – but as there was no deprecation of the old syntax, I think it should be read as bug fix of the and not as spec change.

In OpenMP 5.0, the wording did permit this – unless I missed some restriction elsewhere:
> Any declare target directive with clauses must appear in a specification part of a subroutine subprogram, function subprogram, program or module.

tob2 (changes requested):
> - GCC 12.1.1: C test passed, but Fortran test failed.
>         "undefined reference to `update.1'"

_Side note: I tried with GCC 13/mainline and it compiles + passes this test with nvptx offloading. I have not attempted to confirmed the fail on GCC 12 or on OG12 (devel/omp/gcc-12) ._

tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
```diff
@@ -0,0 +1,97 @@
...
+      c(i) = 0
+    END DO
+     
+    OMPVV_TEST_VERBOSE(test_device_type_nohost() .ne. 0)
+
+    OMPVV_REPORT_AND_RETURN()
+
+CONTAINS  
+    SUBROUTINE target_function()
+      DO i = 1, N
```
spophale, line 50: Add "!$omp declare target to(target_function) device_type(nohost)" before the DO loop.

tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
seyonglee: OpenMP V5.0, Section 2.12.7 declare target Directve, page 183, line 24~26 says the following (OpenMP V5.2 also has the same texts):

This rule applies to our case since `target_function` is an internal procedure, which should not contain any declare target directive because of `!$omp declare target to(target_function) device_type(nohost)` in the main program unit.

tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
seyonglee: It seems that we should not add the declare target directive here (see my comments above).

tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
tob2: > This rule applies to our case since target_function is an internal procedure, which should not contain any declare target directive because of !$omp declare target to(target_function) device_type(nohost) in the main program unit.

But that does not really apply: You can still have
  `!$omp declare target to(target_function) device_type(nohost)`
in the internal procedure. You just cannot have one `!$omp declare target ...device_type` in a program/subroutine/function – and then have another `declare target` in its internal procedure.

While moving your `declare target` code to the internal function is clearly permitted in OpenMP 5.0 to 5.2, placing it into the main program is invalid since 5.1 – and I regard this rather a bug fix and not something which was correct before and just got deprecated.

Thus, I still think you should move the 'declare target' into the internal subroutine, independent whether it a 5.0 (as intended) or a 5.{1,2} testcase.

seyonglee: > If a declare target directive has a device_type clause then any enclosed internal procedures cannot contain any declare target directives. The enclosing device_type clause implicitly applies to internal procedures.

The above restriction still exists in both OpenMP V5.1 and V5.2.
If `!$omp declare target to(target_function) device_type(nohost)` is moved into the internal procedure, does it violate the above restriction? (Even if we move this directive into the internal procedure, `target_function`, the enclosed internal procedure is still the same internal procedure, `target_function`.)

tob2: > The above restriction still exists in both OpenMP V5.1 and V5.2. If `!$omp declare target to(target_function) device_type(nohost)` is moved into the internal procedure, does it violate the above restriction?

The restriction still does exist – but it no longer applies: After moving the "declare target" inside the internal procedure, the "if" condition of the restriction is false (for the outer PROGRAM) and we can ignore the "then" part.

You wouldn't be permitted to add another '!$omp declare target' to an internal procedure of 'target_function' – but that function does not even have internal procedures.

seyonglee:
[Update on Oct. 30, 2022]
Moved `!$omp declare target to(target_function) device_type(nohost)` into the internal procedure, `target_function`, as suggested.
Tested compilers still have the same errors (except for XL Fortran compiler):
        - LLVM 15.0.0: C test passed.
        - GCC 12.1.1: C test passed, but Fortran test failed.
            "undefined reference to `update.1'"
        - XL 16.1.1-10: C test passed (ran on the host) but Fortran test failed (with different error messages)
            "line 57.13: 1515-019 (S) Syntax is incorrect."
        - NVHPC 22.5: 
            - C test failed.
                "line 54: error: this kind of pragma may not be used here
                #pragma omp target "
            - Fortran test failed
                nvlink error   : Undefined reference to '_update_' in '/tmp/nvfortrantdPYh59iBbPKq.o'

tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
seyonglee: Moved `!$omp declare target to(target_function) device_type(nohost)` into the internal procedure, `target_function`, as suggested.

tests/5.0/declare_target/test_declare_target_device_type_nohost.F90
seyonglee: Moved `!$omp declare target to(target_function) device_type(nohost)` into the internal procedure, `target_function`, as suggested.

fel-cab (approved):

seyonglee:
[Update on Nov 27, 2022]
        - LLVM 15.0.0: C test passed.
        - *GCC 12.2.1: C test passed, but Fortran test failed:
            "undefined reference to `update.1'"
        - XL 16.1.1-10: C test passed (ran on the host) but Fortran test failed:
            "line 57.13: 1515-019 (S) Syntax is incorrect."
        - NVHPC 22.5: 
            - C test failed.
                "line 54: error: this kind of pragma may not be used here
                #pragma omp target "
            - Fortran test failed
                nvlink error   : Undefined reference to '_update_' in '/tmp/nvfortrankw3LcEZVWZ7z0.o'

spophale (approved):

PR #621, opened 2022-09-19
Test Depend inoutset

nolanbaker31:
Confused as to the difference between this & mutexinoutset introduced in 5.0. It is clear that the spec states that mutexinoutset dependencies are mutually exclusive, and doesn't state this for outset, but the behavior/use of using inoutset is unclear to me.

Test based on 5.0/task/test_task_depend_mutexinoutset.c test

### DIFF
--- a/tests/5.1/depend/test_depend_inoutset.c
+++ b/tests/5.1/depend/test_depend_inoutset.c
@@ -20,6 +20,9 @@
 int errors;
 int arr[N];
 int depend_inoutset(){
+  for(int i = 0; i < N; i++){
+    arr[i] = 0;
+  }
   #pragma omp parallel
   #pragma omp single
   {
--- a/tests/5.1/depend/test_depend_inoutset.c
+++ b/tests/5.1/depend/test_depend_inoutset.c
@@ -13,32 +13,36 @@
 #include "ompvv.h"
 #include <math.h>
 
-int errors;
+#define N 1024
 
+int errors;
+int arr[N];
 int depend_inoutset(){
-    #pragma omp parallel
-#pragma omp single
+  #pragma omp parallel
+  #pragma omp single
   {
-#pragma omp task depend(out: c)
-   c = 1; /* Task T1 */
- #pragma omp task depend(out: a)
-   a = 2; /* Task T2 */
- #pragma omp task depend(out: b)
-   b = 3; /* Task T3 */
- #pragma omp task depend(in: a) depend(inoutset: c)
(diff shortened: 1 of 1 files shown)


nolanbaker31:
Passes on GCC 12

spophale:
> Confused as to the difference between this & mutexinoutset introduced in 5.0. It is clear that the spec states that mutexinoutset dependencies are mutually exclusive, and doesn't state this for outset, but the behavior/use of using inoutset is unclear to me.
> 
> Test based on 5.0/task/test_task_depend_mutexinoutset.c test

The only difference between mutexinoutset and inoutset, as you pointed out, is the requirement of mutual exclusion for dependence type mutexinoutset. The usefulness of inoutset is to separate two big tasks with dependence on same list item(s).

tests/5.1/depend/test_depend_inoutset.c
```diff
@@ -0,0 +1,44 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+int errors;
+
+int depend_inoutset(){
+    #pragma omp parallel
+#pragma omp single
```
spophale, line 20: Suggestion:
```
#pragma omp parallel
#pragma omp single
{
for (int i=0; i<n; i++) {
#pragma omp task depend(in: _list items_ )
 //some computation using  _list items_ 
}
for (int i=0; i<n; i++) {
#pragma omp task depend(inoutset:  _list items_ )
 //some computation using/updating   _list items_ 
}
```

nolanbaker31:
Test passes on most recent version of LLVM,
Fails on GCC 12.1.0 w/ message `error: invalid depend kind before 'inoutset'
#pragma omp task depend(inoutset: arr[i]),` 
but passes on GCC 12.2.1

tests/5.1/depend/test_depend_inoutset.c
```diff
@@ -0,0 +1,52 @@
...
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+#define N 1024
+
+int errors;
+int arr[N];
```
spophale, line 21: Initialize to 0.

spophale (approved):

spophale:
@mjcarr458  will give 2nd approval.

mjcarr458 (approved):
Looks good to go.

PR #626, opened 2022-10-05
Update test_target_uses_allocators_pteam.c

spophale:
Updating test to check parallelism across parallel region and not teams. Pteam access trait is defined as memory accessible by all threads that bind to the same parallel region of the thread that requested the allocation.

### DIFF
--- a/tests/5.0/target/test_target_uses_allocators_pteam.c
+++ b/tests/5.0/target/test_target_uses_allocators_pteam.c
@@ -23,12 +23,12 @@ int test_uses_allocators_pteam() {
   int result[N] = {0};
 
   for (int i = 0; i < N; i++) {
-    result[i] = 3 * i ;
+    result[i] = 2 * i ;
   }
 
-#pragma omp target teams distribute uses_allocators(omp_pteam_mem_alloc) allocate(omp_pteam_mem_alloc: x) private(x) map(from: device_result)
+#pragma omp target parallel for num_threads(8) uses_allocators(omp_pteam_mem_alloc) allocate(omp_pteam_mem_alloc: x) private(x) map(from: device_result)
   for (int i = 0; i < N; i++) {
-    x = 2 * i;
+    x = omp_get_thread_num();    
     device_result[i] = i + x;
   }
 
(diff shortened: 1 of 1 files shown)


spophale:
@fel-cab and @nolanbaker31 will review.

nolanbaker31 (approved):
lgtm

fel-cab (approved):

ronlieb:
Hi Swaroop, sorry for the slightly delayed feedback on this PR.
After it landed we noticed that our trunk build of LLVM/OpenMP and our local product build started failing this test.

The quick fix is as follows:  Use 64 for N, as its a multiple of wavesize for both AMD and NVPTX. Also use N in the pragma rather than 8.   

diff --git a/tests/5.0/target/test_target_uses_allocators_pteam.c b/tests/5.0/target/test_target_uses_allocators_pteam.c
index e26aa92..23359bf 100644
--- a/tests/5.0/target/test_target_uses_allocators_pteam.c
+++ b/tests/5.0/target/test_target_uses_allocators_pteam.c
@@ -14,7 +14,7 @@
 #include <stdlib.h>
 #include "ompvv.h"

-#define N 8
+#define N 64

 int test_uses_allocators_pteam() {
   int errors = 0;
@@ -26,7 +26,7 @@ int test_uses_allocators_pteam() {
     result[i] = 2 * i ;
   }

-#pragma omp target parallel for num_threads(8) uses_allocators(omp_pteam_mem_alloc) allocate(omp_pteam_mem_alloc: x) private(x) map(from: device_result)
+#pragma omp target parallel for num_threads(N) uses_allocators(omp_pteam_mem_alloc) allocate(omp_pteam_mem_alloc: x) private(x) map(from: device_result)
   for (int i = 0; i < N; i++) {
     x = omp_get_thread_num();
     device_result[i] = i + x;

Explanation from Carlo:
asking for < 64 threads on amdgpu on new device RTL means you get 1. For nvptx64, if you ask for < 32, you get 0. num_threads is only an upper bound on the number of threads that get recruited. This behavior is compliant with the OpenMP specifications.

Try changing num_threads(8) to 64  to see for yourself.

This is different from the old device RTL, where users would get the number they were asking for. We inherited this behavior from trunk, when we moved to the new device rtl.

spophale:
Thanks @ronlieb, I will be creating a patch for this test.

PR #627, opened 2022-10-15
New test for the omp_all_memory locator for the depend clause

jrreap:
Test to confirm if the omp_all_memory locator is implemented correctly. This is the initial rough test, though likely needs some polish.


spophale (approved):

spophale:
@jrreap We also need a test for depend (inout: omp_all_memory)

PR #628, opened 2022-10-16
Flush seq_cst test

nolanbaker31:
Passes on GCC 12 & most recent llvm build
Test is essentially the same as 5.0/tests/flush/test_flush_no_memory_oder_clause.c, as 5.1 states 

> If neither memory-order-clause nor a list appears on the flush construct then the behavior is as if memory-order-clause is seq_cst

### DIFF
--- a/tests/5.1/flush/test_flush_seq_cst.c
+++ b/tests/5.1/flush/test_flush_seq_cst.c
@@ -15,10 +15,38 @@
 #include <stdlib.h>
 #include "ompvv.h"
 
+int flush_default();
+int flush_seq_cst();
+
 int errors = 0;
+int flush_default_result = 0;
+int flush_seq_cst_result = 0;
 
-int main() {
+int flush_default() { // flush with no clause specified
+   int x = 0, y = 0;
+   #pragma omp parallel num_threads(2)
+   {
+      int thrd = omp_get_thread_num();
+      if (thrd == 0) {
+         x = 10;
+         #pragma omp flush
+         #pragma omp atomic write
+         y = 1;
+      } else {
+         int tmp = 0;
+         while (tmp == 0) {
+            #pragma omp atomic read
+	    tmp = y;
+         }
+         #pragma omp flush
+         OMPVV_TEST_AND_SET_VERBOSE(errors, x != 10);
+         flush_default_result = x;
+      }
+  }
+  return errors;
+}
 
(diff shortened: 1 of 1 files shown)


tests/5.1/flush/test_flush_seq_cst.c
```diff
@@ -0,0 +1,45 @@
+//===---- test_flush_seq_cst.c ----------------------------------------------===//
+// 
+// OpenMP API Version 5.1 Nov 2020
+// 
+// This is a test of the flush directive with no memory-order-clause specified.
```
spophale, line 5: Too verbose - don't need 5-8. If the test is based off another test in the V&V, you don't need to mention Examples doc.

tests/5.1/flush/test_flush_seq_cst.c
```diff
@@ -0,0 +1,45 @@
...
+
+int main() {
+
+   int x = 0, y = 0;
+   #pragma omp parallel num_threads(2)
+   {
+      int thrd = omp_get_thread_num();
+      if (thrd == 0) {
+         x = 10;
+         #pragma omp flush seq_cst
```
spophale, line 31: remove seq_cst

tests/5.1/flush/test_flush_seq_cst.c
```diff
@@ -0,0 +1,45 @@
...
+         #pragma omp atomic write
+         y = 1;
+      } else {
+         int tmp = 0;
+         while (tmp == 0) {
+            #pragma omp atomic read
+	    tmp = y;
+         }
+         #pragma omp flush seq_cst
```
spophale, line 40: remove seq_cst

tests/5.1/flush/test_flush_seq_cst.c
nolanbaker31: I'm a bit confused.. the test is for flush seq_cst

spophale: seq_cst  is the memory order clause. I thought you were checking if the default behavior is the same as if seq_cst was specified.

nolanbaker31:
Updated test now uses to functions, one with no clause specified & one with seq_cst to ensure results between the two are the same.

spophale (approved):

spophale:
@krishols could you do the 2nd review ?

krishols (approved):

PR #629, opened 2022-10-17
Fix test_metadirective_nothing.c

tob2:
After a loop-associated construct, a loop-nest must follow. In C/C++, that's 'for' and not a '{' ... '}' block with a contained 'for'. The metadirective uses besides 'nothing': 'parallel for'. This is a combined/composite construct with a worksharing-loop, which is loop-associated construct.

@nolanbaker31 @spophale @jrreap @krishols @mjcarr458  – please review

See also OpenMP 5.1:
* https://www.openmp.org/spec-html/5.1/openmpsu73.html (2.16.1  Parallel Worksharing-Loop Construct) 
* https://www.openmp.org/spec-html/5.1/openmpsu45.html ("loop-nest — One of the following:")


spophale:
@nolanbaker31 can you please review ?

nolanbaker31 (approved):
Looks good

PR #633, opened 2022-10-26
New test for depend inout omp_all_memory

jrreap:
Added additional test to check if `omp_all_memory` works in the `depend` clause with `inout`.

### DIFF
--- a/tests/5.1/depend/test_depend_inout_omp_all_memory.c
+++ b/tests/5.1/depend/test_depend_inout_omp_all_memory.c
@@ -5,7 +5,7 @@
 // This test checks that the special omp_all_memory locator keyword works
 // as intended. The test checks this by having two variables passed into the 
 // target parallel region, depending on both values. If both variables are
-// properly passed through the sections then it passes. If not, the test fails.
+// properly passed through the tasks then it passes. If not, the test fails.
 //
 ////===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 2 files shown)


tests/5.1/depend/test_depend_inout_omp_all_memory.c
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	OMPVV_TEST_OFFLOADING;
+
+	#pragma omp target map(tofrom: errors) map(to: x, y)
+	{
+	    #pragma omp parallel
+	    #pragma omp single
+	    {
+	      #pragma omp task shared(x, y, errors) depend(inout: x, y)
```
spophale, line 35: I don't see the use of omp_all_memory. 
Modify test to have tasks before and after the task with inout dep.

Something like:
#pragma omp parallel
   #pragma omp single
   {
     #pragma omp task depend (out: x){// modify x} //t1
     #pragma omp task depend (out: y){// modify y} //t2
     #pragma omp task depend (inout: omp_all_memory){// modify x,y --> this should see the modifications from t1,t2 }//t3
     #pragma omp task depend (out: x){// modify x} //t4
     #pragma omp task depend (out: y){// modify y} //t5
}
t3 should be scheduled after t1 and t2 and before t4 and t5.

jrreap: Yeah sorry forgot to remove my hard code (to make to compile so it could run and test the rest of the flow).

Good point about having the tasks on both sides though to fully test both variables. Will update!

spophale (approved):

spophale:
@nolanbaker31 could you please review ?

nolanbaker31 (approved):

PR #634, opened 2022-10-28
Create test_declare_target_device_type_nohost1.c

spophale:
This is different from the test_declare_target_device_type_nohost.c - here the default version is the target version and the variant is created for host fall back (match(device={kind(host)}).

This trips up the LLVM compiler.

### DIFF
--- a/tests/5.0/declare_target/test_declare_target_device_type_nohost1.c
+++ b/tests/5.0/declare_target/test_declare_target_device_type_nohost1.c
@@ -53,7 +53,7 @@ int test_declare_target_device_type_nohost() {
 
   #pragma omp target  
   {
-    update(); //Will call fun() on device OR host_function() in case of fall-back
+    fun(); //Will call fun() on device OR host_function() in case of fall-back
     dev = omp_get_device_num();
   }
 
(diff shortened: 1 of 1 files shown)


spophale:
Passes on device with gcc/12.2.1-20220926 on Summit.

FAILS with llvm/16.0.0-20221028 on Summit.

spophale:
@fel-cab and @mjcarr458 will review.

fel-cab (approved):

spophale:
@seyonglee will you be able to review ?

mjcarr458 (approved):
Looks good to me

seyonglee:
It looks good to me too.
[Summit results]
       - NVHPC 22.5: C test failed:
            line 54: error: this kind of pragma may not be used here
                #pragma omp target  
        - LLVM 15.0.0: C test failed:
            50:3: error: function with 'device_type(nohost)' is not available on host
                fun(); //should call host_function()
        - GCC 12.2.1: C test passed.
        - XL 16.1.1-10: C test failed
            undefined reference to `omp_get_device_num`

seyonglee (approved):

PR #639, opened 2022-11-07
Create test_declare_target_enter_device_type_host.c

spophale:
Similar to 5.0 example but with removal of begin-end declare target and enter clause instead of to.


spophale:
Passes with GCC 12.1.0

fel-cab (approved):

PR #640, opened 2022-11-07
Create test_declare_target_enter_device_type_nohost.c

spophale:
Similar to 5.0 test, removing deprecated features adding new clauses.

### DIFF
--- a/tests/5.2/declare_target/test_declare_target_enter_device_type_nohost.c
+++ b/tests/5.2/declare_target/test_declare_target_enter_device_type_nohost.c
@@ -21,11 +21,11 @@
 #define N 10
 int errors = 0;
 int a[N], b[N], c[N];  
-int i = 0;
+int dev=-9, i = 0;
 
 void target_function();
 #pragma omp declare target enter(target_function) device_type(nohost)
-#pragma omp declare target enter(a,b,c,i)
+#pragma omp declare target enter(a,b,c,i,dev)
 
 #pragma omp declare variant(target_function) match(device={kind(nohost)})
 void update() {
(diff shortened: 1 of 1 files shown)


spophale:
Passed on device with GCC 12.1.0 (without enter clause ).

tests/5.2/declare_target/test_declare_target_enter_device_type_nohost.c
```diff
@@ -0,0 +1,92 @@
+//===---- test_declare_target_enter_device_type_nohost.c  ---------------------------===//
+// 
+// OpenMP API Version 5.0
```
fel-cab, line 3: This is a version 5.2 test. Shouldn't this comment be ... Version 5.2?

tests/5.2/declare_target/test_declare_target_enter_device_type_nohost.c
```diff
@@ -0,0 +1,92 @@
...
+  for (i = 0; i < N; i++) {
+    a[i] += 1;
+    b[i] += 2;
+    c[i] += 3;
+  }
+} 
+
+int test_declare_target_enter_device_type_nohost() { 
+
+  update();
```
fel-cab, line 49: As in the test:  tests/5.2/declare_target/test_declare_target_enter_device_type_nohost1.c
Would be useful to comment
//should call host_function()

tests/5.2/declare_target/test_declare_target_enter_device_type_nohost.c
```diff
@@ -0,0 +1,92 @@
...
+
+  #pragma omp target update to(a,b,c)
+
+  #pragma omp target  
+  {
+    update();
```
fel-cab, line 55: As in the test:  tests/5.2/declare_target/test_declare_target_enter_device_type_nohost1.c
Would be useful to comment
//Will call target_function() on device OR update() in case of fall-back

nolanbaker31 (approved):
looks good

fel-cab (approved):

tests/5.2/declare_target/test_declare_target_enter_device_type_nohost.c
spophale: It's not the same as it calls update() which is effectively only for the host

PR #645, opened 2022-12-16
Add a new test (test/5.0/loop/test_loop_bind_device.F90)

seyonglee:
- NVHPC 22.11:
            - C test failed: NVC++-F-1196-OpenMP - 'omp parallel' in a 'loop' region is not supported yet.  (/gpfs/alpine/csc383/scratch/f6l/sollve/tests/5.0/loop/test_loop_bind_device.c: 46)
            - Fortran test failed: NVFORTRAN-S-1224-LOOP construct may not contain calls to the OpenMP Runtime API - . (/gpfs/alpine/csc383/scratch/f6l/sollve/tests/5.0/loop/test_loop_bind_device.F90: 60)
        - LLVM 15.0.0: C test passed but tested with one thread
        - GCC 12.2.1: C test passed; Fortran test passed but ran on the host
        - XL 16.1.1-10: C test passed, but tested with one team and one thread
            - Fortran test failed: line 52.10: 1515-019 (S) Syntax is incorrect.

### DIFF
--- a/tests/5.0/loop/test_loop_bind_device.F90
+++ b/tests/5.0/loop/test_loop_bind_device.F90
@@ -1,0 +1,246 @@
+!===--- test_loop_bind_device.F90 ------------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! This test checks the loop directive with the bind(binding) clause. The bind
+! clause indicates that the loop construct should apply in the context of the
+! given binding, one of teams, parallel, or thread. Each of these bindings
+! is tested in an appropriate context and the correctness of results of
+! array operations in the nested loop is checked. This test checks these
+! directives in a target context.
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 32
+
+PROGRAM test_loop_bind_device
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
+  OMPVV_TEST_VERBOSE(test_bind_teams() .NE. 0)
+  OMPVV_TEST_VERBOSE(test_bind_parallel() .NE. 0)
+  OMPVV_TEST_VERBOSE(test_bind_thread_teams() .NE. 0)
+  OMPVV_TEST_VERBOSE(test_bind_thread_parallel() .NE. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+
+CONTAINS
+  INTEGER FUNCTION test_bind_teams()
+    INTEGER,DIMENSION(N,N):: x
+    INTEGER,DIMENSION(N):: y
+    INTEGER,DIMENSION(N):: z
+    INTEGER:: errors, num_teams, i, j
+
+    OMPVV_INFOMSG("test_loop_bind_teams")
+
+    errors = 0
+    num_teams = -1
+
+    DO i = 1, N
+       DO j = 1, N
+          x(j,i) = 1
+       END DO
+       y(i) = i
+       z(i) = 2*i
+    END DO
+
+    !$omp target teams num_teams(OMPVV_NUM_TEAMS_DEVICE) thread_limit(OMPVV_NUM_THREADS_HOST) map(tofrom: x, y, z, num_teams)
+    !$omp loop bind(teams)
+    DO i = 1, N
+       DO j = 1, N
+          x(j,i) = x(j,i) + y(i)*z(i)
+       END DO
+    END DO
+    !$omp end loop
+    !$omp parallel if(.FALSE.)
+    IF ( omp_get_team_num() .EQ. 0 ) THEN
+       num_teams = omp_get_num_teams()
+    END IF
+    !$omp end parallel
+    !$omp end target teams
+
+    DO i = 1, N
+       DO j = 1, N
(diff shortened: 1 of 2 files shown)


tests/5.0/loop/test_loop_bind_device.F90
```diff
@@ -0,0 +1,246 @@
...
+
+    DO i = 1, N
+       DO j = 1, N
+          x(j,i) = 1
+       END DO
+       y(i) = i
+       z(i) = 2*i
+    END DO
+
+    !$omp target teams num_teams(OMPVV_NUM_TEAMS_DEVICE) thread_limit(OMPVV_NUM_THREADS_HOST) map(tofrom: x, y, z, num_teams)
```
spophale, line 51: thread_limit should use  OMPVV_NUM_THREADS_DEVICE

spophale (changes requested):
Use  OMPVV_NUM_THREADS_DEVICE

tests/5.0/loop/test_loop_bind_device.F90
seyonglee: Updated; thanks.

spophale (approved):

tests/5.0/loop/test_loop_bind_device.F90
```diff
@@ -0,0 +1,246 @@
...
+    !$omp parallel if(.FALSE.)
+    IF ( omp_get_team_num() .EQ. 0 ) THEN
+       num_teams = omp_get_num_teams()
+    END IF
+    !$omp end parallel
+    !$omp end target teams
+
+    DO i = 1, N
+       DO j = 1, N
+          OMPVV_TEST_AND_SET_VERBOSE(errors, x(j,i) .NE. (1 + y(i)*z(i)))
```
fel-cab, line 68: If the error fails, this VERBOSE would produce N*N error output lines, I think it is better to change OMPVV_TEST_AND_SET_VERBOSE for OMPVV_TEST_AND_SET. And you could add a OMPVV_ERROR_IF after the loop with a single message for the loop error.

tests/5.0/loop/test_loop_bind_device.F90
```diff
@@ -0,0 +1,246 @@
...
+    DO i = 1, N
+       DO j = 1, N
+          x(j,i) = 1
+       END DO
+       y(i) = i
+       z(i) = 2*i
+    END DO
+
+    !$omp target teams num_teams(OMPVV_NUM_TEAMS_DEVICE) thread_limit(OMPVV_NUM_THREADS_DEVICE) map(tofrom: x, y, z, num_teams)
+    !$omp loop bind(teams)
```
fel-cab, line 52: I don't really understand the bind, so I have these questions:
Are we interested on checking the behavior of the binding or just the compiler ability to compile correctly?
If the compiler or runtime ignores the bind(teams), would this produce the correct results?

seyonglee: This test just checks whether the compiler can compile correctly or not; even if the compiler ignores the `bind(team)` clause, the correct compiler should generate an output program producing the correct results since the `loop` construct guarantees that each logical iteration is executed once per instance of the `loop` region, no matter how the binding thread set is decided.

tests/5.0/loop/test_loop_bind_device.F90
seyonglee: Good suggestion; updated.

tests/5.0/loop/test_loop_bind_device.F90
```diff
@@ -0,0 +1,246 @@
...
+       DO j = 1, N
+          outData(omp_get_team_num()+1,j,i) = x(j,i)
+       END DO
+    END DO
+    !$omp end target teams
+
+    DO i = 1, N
+       DO j = 1, N
+          DO k = 1, num_teams
+             OMPVV_TEST_AND_SET_VERBOSE(errors, outData(k,j,i) .NE. (1 + y(i)*z(i)))
```
fel-cab, line 174: I suggest changing `OMPVV_TEST_AND_SET_VERBOSE` for `OMPVV_TEST_AND_SET`

tests/5.0/loop/test_loop_bind_device.F90
```diff
@@ -0,0 +1,246 @@
...
+       DO j = 1, N
+          outData(omp_get_thread_num()+1,j,i) = x(j,i)
+       END DO
+    END DO
+    !$omp end target parallel
+
+    DO i = 1, N
+       DO j = 1, N
+          DO k = 1, num_threads
+             OMPVV_TEST_AND_SET_VERBOSE(errors, outData(k,j,i) .NE. (1 + y(i)*z(i)))
```
fel-cab, line 234: I suggest changing `OMPVV_TEST_AND_SET_VERBOSE` for `OMPVV_TEST_AND_SET`

tests/5.0/loop/test_loop_bind_device.F90
seyonglee: Updated

tests/5.0/loop/test_loop_bind_device.F90
seyonglee: Updated

fel-cab (approved):

PR #652, opened 2023-01-09
Target Update Iterator test

nolanbaker31:
Not sure if this is the proper use of the 'iterator' clause, don't see how it fits with this directive. Why not just update the entire array? Or use A[:N], for example.

### DIFF
--- a/tests/5.1/target_update/test_target_update_iterator.c
+++ b/tests/5.1/target_update/test_target_update_iterator.c
@@ -1,0 +1,76 @@
+//===--- test_target_update_iterator.c --------------------------------------------------------------===//
+//
+//  OpenMP API Version 5.1 Aug 2021
+//
+//  This test usese the target update directive with the iterator clause
+//  to update the iterative values.
+//
+//  This example has been adapted from the 5.2 OpenMP Examples document,
+//  "Multidependences Using Iterators" and "Simple target data and target update
+//  constructs"
+//
+////===--------------------------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+#include <math.h>
+
+#define N 1024
+
+typedef struct test_struct{
+    size_t len;
+    int *data;
+} test_struct_t;
+
+void init(struct test_struct *s ){
+    for(size_t i = 0; i < s->len; i++){
+        s->data[i] = i;
+    }
+}
+
+void init_again(struct test_struct *s ){
+    for(size_t i = 0; i < s->len; i++){
+        s->data[i] = i+1;
+    }
+}
+
+int test_target_update_iterator() {
+    int errors = 0;
+    int A[N]; // array for testing
+
+    for(int i = 0; i < N; i++){
+        A[i] = 0;
+    }
+    test_struct_t new_struct;
+
+    init(&new_struct);
+    #pragma omp target map(to: new_struct) map(from: A[:N])
+    {
+        #pragma omp parallel for
+        for(int i = 0; i < N; i++){
+            A[i] = new_struct.data[i];
+        }
+    }
+    init_again(&new_struct);
+    // update with new values, i+i+1
+    #pragma omp target update to(iterator(it = 0:N), new_struct.data[it])
+    #pragma omp target map(from: A[:N])
+    for(int i = 0; i < N; i++){
+            A[i]+=new_struct.data[i];
+    }
+    for(int i = 0; i < N; i++){
+        OMPVV_TEST_AND_SET(errors, A[i] != (i*2)+1);
+    }
+    
+    return errors;
+}
+
+int main() {
+   int errors = 0;
+   OMPVV_TEST_OFFLOADING;
+   OMPVV_TEST_AND_SET_VERBOSE(errors, test_target_update_iterator() != 0);
+   OMPVV_REPORT_AND_RETURN(errors);
+}           
+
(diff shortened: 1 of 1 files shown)


tests/5.1/target_update/test_target_update_iterator.c
```diff
@@ -0,0 +1,76 @@
...
+    #pragma omp target map(to: new_struct) map(from: A[:N])
+    {
+        #pragma omp parallel for
+        for(int i = 0; i < N; i++){
+            A[i] = new_struct.data[i];
+        }
+    }
+    init_again(&new_struct);
+    // update with new values, i+i+1
+    #pragma omp target update to(iterator(it = 0:N), new_struct.data[it])
```
tob2, line 58: I think it should be:
```suggestion
    #pragma omp target update to(iterator(it = 0:N): new_struct.data[it])
```
_Disclaimer: I have not fully read the example and I don't have (tried) a compiler support iterators in map._

nolanbaker31: I believe the comma is correct, spec defines as:

> where iterators-definition is one of the following:
> `iterator-specifier [, iterators-definition ]`

tob2:
> Something interesting I've found is that **iterator** is not listed as a clause for **target update** in OpenMP 5.2 specification, but it is not listed as deprecated in the features history. If it was truly removed, I would be very surprised if any compiler vendors were to implement it.

It's not a _clause_ but a _(clause) modifier_. Thus, if I go to "5.9.1 **to** Clause" or "5.9.2 **from** Clause" in OpenMP 5.2 (or TR11) and look under "Modifiers" you find the modifier named _"iterator"._ — Thus, it still exists. _(But otherwise wouldn't be the first thing that got lost in the syntax reshuffling.)_

tests/5.1/target_update/test_target_update_iterator.c
```diff
@@ -0,0 +1,76 @@
...
+#include <math.h>
+
+#define N 1024
+
+typedef struct test_struct{
+    size_t len;
+    int *data;
+} test_struct_t;
+
+void init(struct test_struct *s ){
```
nolanbaker31, line 27: set s-> len to N

tests/5.1/target_update/test_target_update_iterator.c
```diff
@@ -0,0 +1,76 @@
...
+    for(size_t i = 0; i < s->len; i++){
+        s->data[i] = i;
+    }
+}
+
+void init_again(struct test_struct *s ){
+    for(size_t i = 0; i < s->len; i++){
+        s->data[i] = i+1;
```
nolanbaker31, line 35: +=

spophale (approved):

tests/5.1/target_update/test_target_update_iterator.c
```diff
@@ -0,0 +1,78 @@
...
+    #pragma omp target map(to: new_struct) map(tofrom: A[:N])
+    {
+        #pragma omp parallel for
+        for(int i = 0; i < N; i++){
+            A[i] = new_struct.data[i];
+        }
+    }
+    init_again(&new_struct);
+    // update with new values, (i*2)+1
+    #pragma omp target update to(iterator(it = 0:N): new_struct.data[it])
```
seyonglee, line 60: `target update` construct is called outside of any `target`/`target data` region, and thus no assignment will occur to or from the original list item.

tests/5.1/target_update/test_target_update_iterator.c
```diff
@@ -0,0 +1,78 @@
...
+    int errors = 0;
+    int A[N]; // array for testing
+
+    for(int i = 0; i < N; i++){
+        A[i] = 0;
+    }
+    test_struct_t new_struct;
+
+    init(&new_struct);
+    #pragma omp target map(to: new_struct) map(tofrom: A[:N])
```
seyonglee, line 51: When mapping `new_struct`, its member (`new_struct.data`) should also be explicitly mapped since it is a pointer (compiler cannot know the size of the data referenced by the pointer.)

tests/5.1/target_update/test_target_update_iterator.c
```diff
@@ -0,0 +1,78 @@
...
+#define N 1024
+
+typedef struct test_struct{
+    size_t len;
+    int *data;
+} test_struct_t;
+
+void init(struct test_struct *s ){
+    s->len = N;
+    for(size_t i = 0; i < s->len; i++){
```
spophale, line 29: Needs allocation as data is a ptr.

Add:
s->data =  malloc(sizeof(int)*N);

tests/5.1/target_update/test_target_update_iterator.c
spophale: Agree, it needs to be a) mapped and b)attached.

The maps should be:
```
#pragma omp target enter data map(to: new_struct.data[:N])

    #pragma omp target map(to: new_struct, new_struct.data[:N]) map(tofrom: A[:N]) //pointer attachment
    {
        #pragma omp parallel for
```

tests/5.1/target_update/test_target_update_iterator.c
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+    init_again(&new_struct);
+    // update with new values, (i*2)+1
+    #pragma omp target update to(iterator(it = 0:N): new_struct.data[it])
+    #pragma omp target map(tofrom: A[:N])
+    for(int i = 0; i < N; i++){
+            A[i] += new_struct.data[i];
+    }
+    for(int i = 0; i < N; i++){
```
spophale, line 65: Add:

` #pragma omp target exit data map(delete:new_struct.data[:N])`

before final check on host.

nolanbaker31:
Added recommended changes. Test passes on latest GCC when I remove iterator clause (not implemented)

spophale (approved):

krishols (approved):

PR #637, opened 2022-10-31
Target_Device context selector

nolanbaker31:
Not sure what I would use for AMD arch.
Fails on GCC 12.2.1 w/ message:

`internal compiler error: Segmentation fault
   34 |          default( for)
      |                   ^~~
0x109cace3 crash_signal
	/autofs/nccs-svm1_proj/stf010/elwasif/gccbuild/src/gcc/gcc/toplev.cc:322
0x10805c0b omp_dynamic_cond
	/autofs/nccs-svm1_proj/stf010/elwasif/gccbuild/src/gcc/gcc/omp-general.cc:2093
0x1080b4a3 omp_dynamic_selector_p`

Fails on LLVM w/ message: 
`'target_device' is not a valid context set in a `declare variant``

PR #654, opened 2023-01-09
target_device with "kind" clause (in metadirective) test

nolanbaker31:
References #637, this is for the "kind" clause specifically.

I am struggling to find any clear list of "isa" or "arch", as these are implementation-defined.

### DIFF
--- a/tests/5.1/metadirective/test_metadirective_target_device_kind.c
+++ b/tests/5.1/metadirective/test_metadirective_target_device_kind.c
@@ -1,0 +1,63 @@
+//===--------------------- test_metadirective_target_device_kind.c ---------------------===//
+//
+// OpenMP API Version 5.1 Nov 2020
+// 
+// Tests that target_device is recognized & will work properly when matching an nvidia
+// GPU. More specifically, uses the kind clause to see if gpu or nohost is recognized
+// as the target device's "kind" implementation. 
+//
+////===---------------------------------------------------------------------===//
+
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+#define N 1024
+
+int test_metadirective_target_device() {
+   int errors = 0;
+   int A[N];
+
+   for (int i = 0; i < N; i++) {
+      A[i] = 1;
+   }
+
+   #pragma omp target map(tofrom: A)
+   {
+      #pragma omp parallel num_threads(4)
+      {
+      // Expect that one of these are true, so the array should be set to masked thread num (0)
+      #pragma omp metadirective \
+         when( target_device={kind(gpu)}: masked ) \
+         when( target_device={kind(nohost)}: masked ) \
+         default( for)
+            for (int i = 0; i < N; i++) {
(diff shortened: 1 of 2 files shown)


tob2:
> I am struggling to find any clear list of "isa" or "arch", as these are implementation-defined.

I think there is no way out to either restrict oneself to a small set of values, defined either in the OpenMP spec itself like `kind(host)` or `kind(nohost)` — or those in "OpenMP Additional Definitions" document, both available at https://openmp.org/specifications/

Or to use compiler-specific values, which are hopefully somewhat common on the platforms we are interested it (Nvidia, AMD, Intel?).

_For GCC, those are listed at https://gcc.gnu.org/onlinedocs/libgomp/OpenMP-Context-Selectors.html – the crossrefs go to https://gcc.gnu.org/onlinedocs/gcc/Submodel-Options.html for the respective hardware. I think at least nvidia/gcn values should mostly match LLVM, except that I think LLVM only has 'amdgcn' while GCC had 'gcn' and then added 'amdgcn' as well._

tob2 (approved):
I have not deeply thought about the testcases and whether testing could be improved (and ignored odd corner cases like default device = host but nonhost devices exist).

But I think the tests should work in sensible environments → LGTM.

spophale:
@mjcarr458 will provide second review.

spophale (approved):

jrreap (approved):
Looks good to go!

nolanbaker31:
> Or to use compiler-specific values, which are hopefully somewhat common on the platforms we are interested it (Nvidia, AMD, Intel?).

LLVM has the following list of available "arch" tags: 
`'arm' 'armeb' 'aarch64' 'aarch64_be' 'aarch64_32' 'ppc' 'ppcle' 'ppc64' 'ppc64le' 'x86' 'x86_64' 'amdgcn' 'nvptx' 'nvptx64'`

While x86, x86_64, nvptx and amdgcn are supported, others such as gcn, i386, etc. are not supported in LLVM. I am unsure if writing a test for the shared tags would make sense, as this would not be portable.

While it could be possible to check the compiler on runtime and write the test around this, that could cause its own issues.

tests/5.1/metadirective/test_metadirective_target_device_kind.c
```diff
@@ -0,0 +1,63 @@
...
+   int errors = 0;
+   int A[N];
+
+   for (int i = 0; i < N; i++) {
+      A[i] = 1;
+   }
+
+   #pragma omp target map(tofrom: A)
+   {
+      #pragma omp parallel num_threads(4)
```
nolanbaker31, line 28: Remove num_threads or try num_threads(128) (similar problem to PR #656), test with GCC & NVC

spophale:
Check actual number of threads executing the target region.

nolanbaker31:
Changed to use a similar method to #664 . Seg faults on GCC, but this is because kind is not implemented for target_device yet. (Works with `device={kind(...)}` and `target_device={device_num(...)}`

spophale (approved):

PR #656, opened 2023-01-12
Addressing issue 638. 

### LINKED ISSUE #638: unreliable expectation in test_target_uses_allocators_pteam
The recent commit 2c8ca2d6cb28f89a3f3916b10bf416c23ca151c8 on the master branch introduced the usage of `omp_get_thread_num()` inside `#pragma omp target`. Although the clause `num_threads(8)` was specified, it is not guaranteed that eight threads are indeed available and provided from my point of view.

In section "2.6.1 Determining the Number of Threads for a parallel Region" of the OpenMP Application Programming Interface (V. 5.0) the algorithm states:
```
...
else if (dyn-var = true) and (ThreadsRequested > ThreadsAvailable)
then 1 ≤ number of threads ≤ ThreadsAvailable;
...
else if (dyn-var = false) and (ThreadsRequested > ThreadsAvailable)
then behavior is implementation defined;
```
That means, even if we try to avoid a dynamic adjustment of the number of threads (e.g. with `omp_set_dynamic`), the specified number of threads are not reliably provided.

Subsequently, `x = omp_get_thread_num();` delivers not necessarily a value between 0 and 7 but any number between 0 and the number of available threads in general. In the case that the number of available threads is less then the number of requested threads (i.e. 8), the test fails due to `result[i] != device_result[i]`.

spophale:
Removing checking based on thread number.

Note: 
With LLVM trunk if no num_threads clause is used with target, the number of threads defaults to 128. If any number less than 128 is specified, the actual number of threads executing the region is 1.
It is reasonable to expect the same behavior from other LLVM based compilers.

closes #638

### DIFF
--- a/tests/5.0/target/test_target_uses_allocators_pteam.c
+++ b/tests/5.0/target/test_target_uses_allocators_pteam.c
@@ -14,22 +14,22 @@
 #include <stdlib.h>
 #include "ompvv.h"
 
-#define N 8
+#define N 256
 
 int test_uses_allocators_pteam() {
   int errors = 0;
   int x = 0;
+  int pteam_result[N] = {0};
   int device_result[N] = {0};
   int result[N] = {0};
 
   for (int i = 0; i < N; i++) {
     result[i] = 2 * i ;
   }
-
-#pragma omp target parallel for num_threads(8) uses_allocators(omp_pteam_mem_alloc) allocate(omp_pteam_mem_alloc: x) private(x) map(from: device_result)
+#pragma omp target parallel for map(from: device_result) uses_allocators(omp_pteam_mem_alloc) allocate(omp_pteam_mem_alloc: pteam_result) private(pteam_result)
   for (int i = 0; i < N; i++) {
-    x = omp_get_thread_num();    
-    device_result[i] = i + x;
+    pteam_result[i] = 2 * i ;
+    device_result[i] = pteam_result[i];
   }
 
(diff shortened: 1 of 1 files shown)


tests/5.0/target/test_target_uses_allocators_pteam.c
```diff
@@ -14,22 +14,22 @@
 #include <stdlib.h>
 #include "ompvv.h"
 
-#define N 8
+#define N 256
 
 int test_uses_allocators_pteam() {
   int errors = 0;
   int x = 0;
```
nolanbaker31, line 21: unused variable

nolanbaker31 (approved):
lgtm

spophale:
@seyonglee Pls can you provide a review ?

seyonglee:
It looks good to me too.
However, one possible issue with change (removing the exiplict `num_threads` clause) is that this test may depend on the value of the environment variable, `OMP_NUM_THREADS`, if set in the tested system. The tester should check whether `OMP_NUM_THREADS` is correctly set if existing.

seyonglee:
> It looks good to me too.
> However, one possible issue with change (removing the exiplict `num_threads` clause) is that this test may depend on the value of the environment variable, `OMP_NUM_THREADS`, if set in the tested system. The tester should check whether `OMP_NUM_THREADS` is correctly set if existing.
This may not matter anyway, since the new version does not depend on the number of threads.

PR #664, opened 2023-01-23
Target Device "device_num" test

nolanbaker31:
Similar to PR #654

Uses a device num of 0, which is the default device.

### DIFF
--- a/tests/5.1/metadirective/test_metadirective_target_device_num.c
+++ b/tests/5.1/metadirective/test_metadirective_target_device_num.c
@@ -28,7 +28,7 @@ int test_metadirective_target_device() {
    // Expect that device_num is 0, so the array should be mapped back.
    #pragma omp metadirective \
       when( target_device={device_num(dev)}: target defaultmap(none) map(tofrom: A)) \
-      default( target defaultmap(none) map(to: A)
+      default( target defaultmap(none) map(to: A))
       for(int i = 0; i < N; i++){
          A[i] = i;
       }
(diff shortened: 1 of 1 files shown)


tests/5.1/metadirective/test_metadirective_target_device_num.c
```diff
@@ -0,0 +1,62 @@
...
+   int errors = 0;
+   int A[N];
+
+   for (int i = 0; i < N; i++) {
+      A[i] = 1;
+   }
+
+   #pragma omp target map(tofrom: A)
+   {
+      #pragma omp parallel num_threads(4)
```
spophale, line 27: remove num_threads -- > then we will get 128 threads for NVIDIA GPUs and 256 for AMD. Need to check if this holds for GNU.

spophale: Also, what happens when this runs on host ?

tests/5.1/metadirective/test_metadirective_target_device_num.c
```diff
@@ -0,0 +1,62 @@
...
+      {
+      // Expect that device_num is 0, so the array should be set to masked thread num (0)
+      #pragma omp metadirective \
+         when( target_device={device_num(0)}: masked ) \
```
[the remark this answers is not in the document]
nolanbaker31: I like this idea, except that this would not be reported as an error. Trying another method.

spophale (approved):

jrreap (approved):
Looks good to go!

Issue #672, opened 2023-02-06
test_target_update_to_present.c: Invalid test case


tob2:
The first testcase assumes that neither **target update** nor anything else has mapped the data – such that **map(from:** is active, which just reserves the memory on the remote system – such that the value is random and does not match the previously assigned memory on the host:

https://github.com/SOLLVE/sollve_vv/blob/c5b28e96165a890ba79d91f3dc396d1e2fd78a7e/tests/5.1/target_update/test_target_update_to_present.c#L45-L53

The next code then places the memory on the device – and tries the same with present:
https://github.com/SOLLVE/sollve_vv/blob/c5b28e96165a890ba79d91f3dc396d1e2fd78a7e/tests/5.1/target_update/test_target_update_to_present.c#L56-L65

In principle, either testcase is by itself fine — the problem is only that the **map(from:** of the first testcase causes that the host value is overridden by undefined device memory.

**Solutions:**
=========

One way would be to use **map(alloc:** instead of **map(from:** for the first target region. This would still use undefined behavior. (The value of the variables is not specified – it is unlikely to be the host value, unless (unified-)shared memory is used, but still.)

```diff
--- a/tests/5.1/target_update/test_target_update_to_present.c
+++ b/tests/5.1/target_update/test_target_update_to_present.c
@@ -45,3 +45,3 @@ int test_motion_present() {
    #pragma omp target update to(scalar_var, A, new_struct) 
-   #pragma omp target map(tofrom: errors) defaultmap(none) map(from: scalar_var, A, new_struct)
+   #pragma omp target map(tofrom: errors) defaultmap(none) map(alloc: scalar_var, A, new_struct)
    {     
```
(Alternatively, the values could be reset inside the **target** region or after that region.)

Or to avoid the unspecified behavior and replace the target region by an API call:

```diff
--- a/tests/5.1/target_update/test_target_update_to_present.c
+++ b/tests/5.1/target_update/test_target_update_to_present.c
@@ -46,8 +46,6 @@ int test_motion_present() {
-   #pragma omp target map(tofrom: errors) defaultmap(none) map(from: scalar_var, A, new_struct)
-   {     
-        if(scalar_var == 1){errors++;}
-        if(A[0] == 0 || A[50] == 50){errors++;}
-        if(A[50] == 50 || A[51] == 51){errors++;}
-        if(new_struct.s == 10){errors++;}
-        if(new_struct.S[0] == 10){errors++;}
-   }
[...]

PR #673, opened 2023-02-07
test_target_update_to_present.c: Fix undefined behaviour

tob2:
Issue #672
The tests/5.1/target_update/test_target_update_to_present.c has the following issues:

* The first target region uses 'from:' but accesses the uninitialized memory on the device. (Used to test that the variable is not yet mapped.)
* On exit of the target region, the 'from:' causes that this memory is copied back to the host - but the checks in the second target region check that the original value is present, which will fail.

Solution:
* The second issue could be mended by using 'alloc:' instead of 'from:', but it still has the undefined behavior as the variables are then uninitialized. Hence, the first target region has been removed. Instead:
* A omp_target_is_present check (an OpenMP 5.0 API routine) is now used to check whether the variable has been mapped - both instead of the first target region (nothing should be mapped) and before the second one (here, the variables should be mapped; added for completeness).
* Finally, the memory is unmapped - just for cleanness.

@seyonglee, @nolanbaker31, @jrreap, @spophale – please review.

### DIFF
--- a/tests/5.1/target_update/test_target_update_to_present.c
+++ b/tests/5.1/target_update/test_target_update_to_present.c
@@ -43,18 +43,22 @@ int test_motion_present() {
   
    // Tests OpenMP 5.1 Specification pp. 207 lines 2-4
    #pragma omp target update to(scalar_var, A, new_struct) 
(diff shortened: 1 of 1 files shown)


nolanbaker31 (approved):
Nice changes, looks good to me.

spophale (approved):

Issue #671, opened 2023-02-06
test_target_defaultmap_present.c: Should always fail with a complient implementation.


tob2:
The problem is that `ptr` is not mapped.

The only OpenMP directives in the code are the following two – and `ptr` is accessed at the bottom of the target region:

https://github.com/SOLLVE/sollve_vv/blob/c5b28e96165a890ba79d91f3dc396d1e2fd78a7e/tests/5.1/target/test_target_defaultmap_present.c#L43-L58

The problem is `map(present: ptr)` — which will fail at run time as `ptr` has not been mapped and, thus, it is not present on the device.

Possible solution:
```diff
--- a/tests/5.1/target/test_target_defaultmap_present.c
+++ b/tests/5.1/target/test_target_defaultmap_present.c
@@ -45 +45 @@ int test_defaultmap_present() {
-   #pragma omp target map(tofrom: errors) defaultmap(present)
+   #pragma omp target map(tofrom: errors) defaultmap(present) private(ptr)
```

Another possible solution:
```diff
--- a/tests/5.1/target/test_target_defaultmap_present.c
+++ b/tests/5.1/target/test_target_defaultmap_present.c
@@ -43 +43 @@ int test_defaultmap_present() {
-   #pragma omp target enter data map(to: scalar_var, A, new_struct)
+   #pragma omp target enter data map(to: scalar_var, A, new_struct, ptr)
```

Cross ref: Originally added via Pull Request #390.

PR #674, opened 2023-02-07
test_target_defaultmap_present.c: Map 'ptr'

tob2:
Issue #671
* tests/5.1/target/test_target_defaultmap_present.c: Map 'ptr' in target enter data. (As 'ptr' is used in a target region with 'defaultmap(present)', it needs to be mapped before to avoid runtime error termination.)

@seyonglee, @nolanbaker31, @jrreap, @spophale – please review.


spophale (approved):
LGTM.

fel-cab (approved):

PR #675, opened 2023-02-08
Organized make.def to make it easier to understand and to change if n…

fel-cab:
I have tested the make.def on Crusher, Perlmutter and Summit. On Summit it is necessary to change the 80 for 70 for the architecture specific flags

### DIFF
--- a/sys/make/make.def
+++ b/sys/make/make.def
@@ -67,147 +67,93 @@ CC?=none
 CPP?=none
 FC?=none
 
+ifeq ($(OMP_VERSION), 5.0)
+   OMPV = -fopenmp-version=50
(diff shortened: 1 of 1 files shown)


fel-cab:
> I cannot remember what the issue was with the math.h + clang, but the work around has been eliminated in this version. Have you tested the file with all the tests in the "application_kernels" as well ?

I have tested this in Perlmutter, Summit, and Crusher, and it works for compilers nvc, cray, amd, llvm and gcc.

AnonNick (approved):
Tested on Crusher

nolanbaker31 (approved):
Passes on Summit

PR #678, opened 2023-02-08
Add new_test/test_requires_atomic_default_mem_order_relaxed.F90

seyonglee:
- NVHPC 22.11:
            C test failed: line 19: error: invalid text in pragma
            Fortran test failed: NVFORTRAN-S-0034-Syntax error at or near (.../sollve/tests/5.0/requires/test_requires_atomic_default_mem_order_relaxed.F90: 22)
        - LLVM 15.0.0: C test passed.
        - LLVM 16.0.0: C test passed.
        - LLVM 17.0.0: C test passed.
        - GCC 12.2.1: Both C and Fortran tests passed.
        - XL 16.1.1-10: 
            - C test passed.
            - Fortran test failed: line 22.9: 1515-019 (S) Syntax is incorrect.

### DIFF
--- a/tests/5.0/requires/test_requires_atomic_default_mem_order_relaxed.F90
+++ b/tests/5.0/requires/test_requires_atomic_default_mem_order_relaxed.F90
@@ -37,7 +37,10 @@ CONTAINS
 
     OMPVV_INFOMSG("test_requires_atomic_relaxed")
 
-    !$omp parallel num_threads(2) private(thrd, tmp)
+    call omp_set_dynamic(.false.)
+    call omp_set_num_threads(2)
+    
+    !$omp parallel private(thrd, tmp)
        thrd = omp_get_thread_num()
        IF (thrd .EQ. 0) THEN
           x = 10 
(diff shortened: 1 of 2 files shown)


tests/5.0/requires/test_requires_atomic_default_mem_order_relaxed.F90
```diff
@@ -0,0 +1,61 @@
...
+  INTEGER FUNCTION test_atomic_relaxed()
+    INTEGER:: x, y, thrd, tmp, errors
+
+    x = 0
+    y = 0
+    errors = 0
+
+    OMPVV_INFOMSG("test_requires_atomic_relaxed")
+
+    !$omp parallel num_threads(2) private(thrd, tmp)
```
spophale, line 40: We need to add a num_threads check here, in case we get only 1 thread. OR omp_set_dynamic(.false.) before parallel.

spophale (approved):

krishols (approved):

PR #680, opened 2023-02-10
New test for the omp_calloc routine (device version)

jrreap:
Test for the omp_calloc routine if allocation works on a target device properly.


tests/5.1/allocate/test_target_calloc.c
```diff
@@ -0,0 +1,80 @@
...
+  {
+
+    int *x;
+
+    omp_memspace_handle_t  memspace = omp_default_mem_space;
+    omp_alloctrait_t       traits[0] = {};
+    omp_allocator_handle_t alloc = omp_init_allocator(memspace,0,traits);
+
+    x = (int *)omp_calloc(64, N*sizeof(int), alloc);
+
```
fel-cab, line 34: I think you should add something like:
`if (x == NULL) { OMPVV_ERROR("omp_calloc returned null"); return (1); } `

tests/5.1/allocate/test_target_calloc.c
```diff
@@ -0,0 +1,80 @@
...
+#include <stdint.h>
+#include "ompvv.h"
+
+#define N 1024
+
+int test_omp_target_calloc() {
+
+  int errors = 0;
+
+  #pragma omp target map(tofrom: errors) map(to: N) uses_allocators(omp_default_mem_alloc)
```
spophale, line 24: This is not the way to create an allocator in the target region.
See: https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.0/target/test_target_uses_allocators_default.c

jrreap: Will polish this up and correctly allocate this

tests/5.1/allocate/test_target_calloc.c
```diff
@@ -0,0 +1,80 @@
...
+  #pragma omp target map(tofrom: errors) uses_allocators(omp_default_mem_alloc)
+  {
+
+    int *x;
+
+    x = (int *)omp_calloc(64, N*sizeof(int), omp_default_mem_alloc);
+
+    if (x == NULL) { 
+      OMPVV_ERROR("omp_calloc returned null"); 
+      return (1);
```
tob2, line 33: GCC complains here:
```
test_target_calloc.c:33:14: error: invalid branch to/from OpenMP structured block
```
(However, GCC does accept `exit(1);` or `abort();`.)

Likewise with Clang:
```
test_target_calloc.c:35:7: error: cannot return from OpenMP region
```

From the OpenMP spec:
* _structured block_ — For C/C++, an executable statement, possibly compound, with a single entry at the9
top and a single exit at the bottom, or an OpenMP construct. [...]

spophale (approved):

fel-cab (approved):

tests/5.1/allocate/test_target_calloc.c
```diff
@@ -0,0 +1,82 @@
+//===------------------------ test_calloc_host.c --------------------------===//
```
tob2, line 1:
```suggestion
//===------------------------ test_target_calloc.c ------------------------===//
```

PR #681, opened 2023-02-15
target_teams_distribute_thread_limit fix

### LINKED ISSUE #655: Test 4.5 test_target_teams_distribute_thread_limit.c seems off
**Describe the bug**
The test conflates `omp_get_thread_limit` and `omp_get_num_threads` and uses the former instead of `omp_get_teams_thread_limit` inside a teams region. It also mixes up the concept of threads in a parallel region vs threads for a parallel region.
The description states: "If the region is run with less threads than ..." but everything is actually just run with *1* thread per team and what is queried is the potential number of threads to run a new parallel region.

**Test that it applies to**
https://github.com/SOLLVE/sollve_vv/blob/master/tests/4.5/target_teams_distribute/test_target_teams_distribute_thread_limit.c

**Expected behavior**
Use a parallel region and `omp_get_num_threads`, use `omp_get_teams_thread_limit` before.

**Compiler**
Any.

**Accelerator hardware**
Any

nolanbaker31:
Uses omp_get_teams_thread_limit() instead of omp_get_thread_limit, as well as uses a parallel region in each team w/ omp_get_num_threads().

Fails on LLVM & GCC as omp_get_teams_thread_limit is returning 0 (or isn't recognized on LLVM), works with omp_get_thread_limit.

Closes #655

### DIFF
--- a/tests/4.5/target_teams_distribute/test_target_teams_distribute_thread_limit.c
+++ b/tests/4.5/target_teams_distribute/test_target_teams_distribute_thread_limit.c
@@ -1,15 +1,15 @@
-//===--- test_target_teams_distribute_num_threads.c--------------------------===//
+//===--- test_target_teams_distribute_thread_limit.c--------------------------===//
 //
 // OpenMP API Version 4.5 Nov 2015
 //
-// This test uses the num_threads clause on a target teams distribute directive to
+// This test uses the thread_limit clause on a target teams distribute directive to
 // indicate a requested number of threads to execute the teams distribute region.
 // The specifications indicate that the number of threads that are given can be any
 // number that is equal to or less than the indicated value. We first run a
 // target teams distribute region without the clause to see what the default
-// number of threads is, and then we use a value that is less than that in the
-// test of the num_threads clause. If the region is run with more threads than
-// indicated, the test errors. If the region is run with less threads than
+// number of maximum threads is, and then we use a value that is less than that in the
+// test of the thread_limit clause. If the region is run with more threads than
+// indicated, the test fails. If the region is run with less threads than
 // indicated, the test issues a warning since it is known that the device can
 // run with more threads than was actually given.
 //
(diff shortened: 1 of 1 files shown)


tests/4.5/target_teams_distribute/test_target_teams_distribute_thread_limit.c
```diff
@@ -22,34 +22,50 @@
...
-  int num_threads;
+int test_target_teams_distribute_thread_limit(){
+  int default_thread_limit;
+  int default_threads[N];
+  int num_thread_limit;
+  int num_threads[N];
   int errors = 0;
 
-#pragma omp target teams distribute map(from: default_threads)
+#pragma omp target teams distribute map(from: default_thread_limit, default_threads)
```
nolanbaker31, line 32: add parallel for

nolanbaker31: Does not work, using simple parallel instead

nolanbaker31:
Passes on GCC 12.1 when using omp_get_thread_limit instead of omp_get_teams_thread_limit. 

Using omp_get_teams_thread_limit has crashed every time on LLVM & GCC, so either not implemented or I am not using it correctly.

spophale:
We need to rethink the test "omp_get_teams_thread_limit() " was introduced in 5.1 and this is a 4.5 test.

spophale:
Simplifying. @nolanbaker31 see if you agree.

```
//===--- test_target_teams_distribute_thread_limit.c--------------------------===//
//
// OpenMP API Version 4.5 Nov 2015
//
// This test uses the thread_limit clause on a target teams distribute directive to
// indicate a requested number of threads to execute the teams distribute region.
// The specifications indicate that the number of threads that are given can be any
// number that is equal to or less than the indicated value. 
//
////===----------------------------------------------------------------------===//

#include <omp.h>
#include <stdio.h>
#include <stdlib.h>
#include "ompvv.h"

#define N 1024

int test_target_teams_distribute_thread_limit(){
  int num_threads = -9;
  int errors = 0;
  
  
  #pragma omp target teams distribute thread_limit(24) map(from: num_threads)
    for (int x = 0; x < N; ++x) {
      if (omp_get_team_num() == 0) {
        #pragma omp parallela
        {
        if (omp_get_thread_num() == 0)
          num_threads = omp_get_num_threads();
        }
      }
    }
    OMPVV_TEST_AND_SET_VERBOSE(errors, num_threads > 24);

  return errors;

}

int main(){
  int errors = 0;
  OMPVV_TEST_OFFLOADING;

  OMPVV_TEST_AND_SET_VERBOSE(errors, test_target_teams_distribute_thread_limit() != 0);
}

```

nolanbaker31:
> Simplifying. @nolanbaker31 see if you agree.

I agree this test case is effective. The only change I would say is to maybe use a lower thread_limit, such as 4. I believe this would give slightly more coverage. (i.e. in scenarios where the number of threads was already less then `thread_limit()`.)

spophale (approved):

seyonglee (approved):
It looks good to me

PR #682, opened 2023-02-16
Update test_target_update_mapper_from_discontiguous.c

### LINKED ISSUE #666: Test 5.0/target_update/test_target_update_mapper_from_discontiguous.c declare mapper needs to map contiguous array section
**Describe the bug**
Reported by: Whitney, Brian <[brian.whitney@intel.com](mailto:brian.whitney@intel.com)>

`#pragma omp declare mapper(newvec_t v) map(v, v.data[0:v.len:2])`

Because it is in a map, it has to be a contiguous array section.

The 5.0 spec states that : If a mapper is specified for the type being mapped, or explicitly specified with the mapper map-type-modifier, then all map clauses that appear on the declare mapper directive are treated as though they appeared on the construct with the map clause.

Since map does not allow discontiguous maps, it should be:
`#pragma omp declare mapper(newvec_t v) map(v, v.data[0:v.len])`

**Test that it applies to**

5.0/target_update/test_target_update_mapper_from_discontiguous.c

**Compiler**
Intel (applicable to all)

spophale:
Closes #666

### DIFF
--- a/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
+++ b/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
@@ -44,11 +44,11 @@ int target_update_from_mapper() {
   #pragma omp target data map(mapper(S), to:s) 
   { 
     #pragma omp target defaultmap(none) map(to:s.len)
-    for (int i = 0; i < s.len; i+=2) {
+    for (int i = 0; i < s.len; i++) {       
       s.data[i] += i ;
     }
   
-  #pragma omp target update from(s.data[0:s.len:2])
+  #pragma omp target update from(s.data[0:s.len:2]) //only update even array elements
   } 
   
 
--- a/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
+++ b/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
@@ -25,7 +25,7 @@ typedef struct newvec {
 size_t i;
 int errors = 0;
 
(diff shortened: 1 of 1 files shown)


fel-cab:
The test name has discontiguous word, and in the comments it says that maps memory array with a stride. But the stride 2 was removed and changed to contiguous (no stride). Should the name be changed or the test. I don't see a big difference with tests/5.0/declare_mapper/test_declare_mapper_target_struct.c

fel-cab (approved):

spophale:
Ping @nolanbaker31

nolanbaker31 (approved):

tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
```diff
@@ -17,40 +16,58 @@
...
       s.data[i] = i;
     }
   }//end target
-  #pragma omp target update from(s)
-  for (i =0; i < s.len; i += 2) {
-    OMPVV_TEST_AND_SET(errors, s.data[i] != i);
+  
+  #pragma omp target data map(mapper(S), to:s) 
+  { 
+    #pragma omp target defaultmap(none) map(to:s.len)
```
seyonglee, line 46: Do we need `map(to:s.len)` here? It seems to be redundant.

spophale: est_update_mapper_dis.c:37:25: error: variable 's' must have explicitly specified data sharing attributes, data mapping attributes, or in an is_device_ptr clause
    for (int i = 0; i < s.len; i++) {       
                        ^

PR #683, opened 2023-02-22
Add new_test/test_atomic_acquire_release.F90

seyonglee:
- NVHPC 22.11:
            - C test failed: line 30: error: invalid text in pragma
            - Fortran test: NVFORTRAN-S-0034-Syntax error at or near identifier release
        - LLVM 15.0.0: C test passed.
        - LLVM 17.0.0: C test passed.
        - GCC 12.2.1:
        - XL 16.1.1-10:
            - C test passed.
            - Fortran test failed: line 42.16: 1515-019 (S) Syntax is incorrect.

### DIFF
--- a/tests/5.0/atomic/test_atomic_acquire_release.F90
+++ b/tests/5.0/atomic/test_atomic_acquire_release.F90
@@ -1,0 +1,58 @@
+!//===------ test_atomic_acquire_release.F90 --------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+! Adapted from OpenMP examples acquire_release.2.c
+! When the atomic read operation on thread 1 reads a non-zero value from y,
+! this results in a release/acquire synchronization that in turn implies that 
+! the assignment to x on thread 0 happens before the read of x on thread 1. 
+!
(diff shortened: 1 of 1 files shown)


tests/5.0/atomic/test_atomic_acquire_release.F90
```diff
@@ -0,0 +1,56 @@
...
+  INTEGER FUNCTION test_atomic_acquire_release()
+    INTEGER:: x, y, thrd, tmp, errors
+
+    x = 0
+    y = 0
+    errors = 0
+
+    OMPVV_INFOMSG("test_atomic_acquire_release")
+
+    !$omp parallel num_threads(2) private(thrd, tmp)
```
spophale, line 36: Should we be checking if we got 2 threads ? What happens of the implementation gives us only 1 thread ?

seyonglee: If there is only 1 thread, the test becomes a trivial case, where atomic constructs are not needed at all.

spophale (approved):

fel-cab (approved):

PR #685, opened 2023-03-01
Add new_test/test_atomic_num_hint.F90

seyonglee:
new_test/test_atomic_num_hint.F90

        - NVHPC 22.11:
            - C test failed: line 28: error: invalid text in pragma
            - Fortran test failed: NVFORTRAN-S-0034-Syntax error at or near identifier hint 
        - LLVM 15.0.0: C test passed.
        - LLVM 17.0.0: C test passed.
        - GCC 12.2.1:
        - XL 16.1.1-10:
            - C test passed, but test ran with one thread, so the results are not conclusive
            - Fortran test failed:  line 41.13: 1515-019 (S) Syntax is incorrect.

### DIFF
--- a/tests/5.0/atomic/test_atomic_num_hint.F90
+++ b/tests/5.0/atomic/test_atomic_num_hint.F90
@@ -43,7 +43,7 @@ CONTAINS
        !$omp end atomic 
     !$omp end parallel
 
-    OMPVV_ERROR_IF(num_threads .LT. 0, "Test ran with invalid number of teams (less than zero)")
+    OMPVV_ERROR_IF(num_threads .LT. 0, "Test ran with invalid number of threads (less than zero)")
     OMPVV_WARNING_IF(num_threads .EQ. 1, "Test ran with one thread, so the results are not conclusive")
 
     OMPVV_TEST_AND_SET_VERBOSE(errors, x .NE. num_threads)
@@ -69,7 +69,7 @@ CONTAINS
        !$omp end atomic 
     !$omp end parallel
 
-    OMPVV_ERROR_IF(num_threads .LT. 0, "Test ran with invalid number of teams (less than zero)")
+    OMPVV_ERROR_IF(num_threads .LT. 0, "Test ran with invalid number of threads (less than zero)")
     OMPVV_WARNING_IF(num_threads .EQ. 1, "Test ran with one thread, so the results are not conclusive")
 
     OMPVV_TEST_AND_SET_VERBOSE(errors, x .NE. num_threads)
--- a/tests/5.0/atomic/test_atomic_num_hint.F90
+++ b/tests/5.0/atomic/test_atomic_num_hint.F90
@@ -1,0 +1,80 @@
+!//===------ test_atomic_num_hint.F90 -------------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+!  This test checks if atomic hints passed by enum value
+!  are accepted by the compiler. If the sync hint is not
+!  yet defined in the specification, it defaults to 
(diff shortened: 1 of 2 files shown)


spophale (approved):

tests/5.0/atomic/test_atomic_num_hint.F90
```diff
@@ -0,0 +1,80 @@
...
+    x = 0
+    num_threads = -1
+
+    OMPVV_INFOMSG("test_atomic_with_used_enum_value")
+
+    !$omp parallel num_threads(2) default(shared)
+       IF (omp_get_thread_num() .EQ. 0) THEN
+          num_threads = omp_get_num_threads()
+       END IF
+       !$omp atomic hint(4) ! corrosponds to omp_sync_hint_nonspeculative
```
tob2, line 41:
```suggestion
       !$omp atomic hint(4) ! corresponds to omp_sync_hint_nonspeculative
```
Question: Why does the testcase use '4' instead of omp_sync_hint_nonspeculative ? That values should be in 'omp.h' (for C/C++) and in 'omp_lib.h' (Fortran include) and in the 'omp_lib' module (Fortran).  [Same for the C testcase.]

As the value is normative, using '4' should also work – but it feels much more natural to use the name, doesn't it? And if 4 works and omp_sync_hint_nonspeculative fails, that would be IMHO an implementation bug.

tests/5.0/atomic/test_atomic_num_hint.F90
```diff
@@ -0,0 +1,80 @@
...
+    x = 0
+    num_threads = -1
+
+    OMPVV_INFOMSG("test_atomic_with_unused_enum_value")
+
+    !$omp parallel num_threads(OMPVV_NUM_THREADS_HOST) default(shared)
+       IF (omp_get_thread_num() .EQ. 0) THEN
+          num_threads = omp_get_num_threads()
+       END IF
+       !$omp atomic hint(4132) ! As of OMP Spec 5.0 only values till 0x8 have been taken
```
tob2, line 67: (BTW, the highest possible value is not 8 but 8 + 4 + 2 + 1 =15.)

I wonder what's the purpose to use a non-taken number here; what's the expected result?

While I see the wording _"The valid constants must include the following, which can be extended with implementation-defined values:"_ which permits additional values, it is unclear what an implementation should/might do with a wrong value: ignore? compile-time error? runtime-error?

I assume that many implementations fully ignore the _hint_ (after parsing) – at least GCC currently does so.

seyonglee: As I understand, spec does not clarify what should happen when non-taken number is passed as a hint, but spec says that "The presence of a hint clause does not affect the semantics of the atomic construct, and all hints may be ignored.", and thus the purpose of this test may be to check whether the implementation CAN ignore the unexpected hint argument.

seyonglee: By the way, "As of OMP Spec 5.0 only values till 0x8 have been taken" means that OpenMP V5.0 has defined up to 0x8 bit (and the highest possible value is not 15 because there are restrictions on which hint values can be combined).

tests/5.0/atomic/test_atomic_num_hint.F90
seyonglee: We have a separate test `test_atomic_hint.c` and `test_atomic_hint.F90` where hint constants defined by spec are used. The purpose of this test is to see whether implementations work if hint clause argument is an integer number.

jrreap (approved):
Looks good, assuming that the comments have been resolved at this point.

PR #686, opened 2023-03-01
Add new_test/test_atomic_num_hint_device.F90

seyonglee:
- NVHPC 22.11:
            - C test failed: line 29: error: invalid text in pragma
            - Fortran test failed: NVFORTRAN-S-0034-Syntax error at or near identifier hint
        - LLVM 15.0.0:
            - C test passed, but test ran with one thread, so the results are not conclusive
        - LLVM 17.0.0:
            - C test passed, but test ran with one thread, so the results are not conclusive
        - GCC 12.2.1:
        - XL 16.1.1-10:
            - C test passed, but test ran on host with one thread, so the results are not conclusive
            - Fortran test failed:   line 43.13: 1515-019 (S) Syntax is incorrect.

### DIFF
--- a/tests/5.0/atomic/test_atomic_num_hint_device.F90
+++ b/tests/5.0/atomic/test_atomic_num_hint_device.F90
@@ -1,0 +1,85 @@
+!//===------ test_atomic_num_hint_device.F90 ------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
+!  This test checks if atomic hints passed by enum value
+!  are processed correctly for device execution. If the sync hint is not
+!  yet defined in the specification, it defaults to 
+!  omp_sync_hint_none (0x0). 
+!
+!//===---------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_atomic_num_hint_device
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+
+  OMPVV_TEST_OFFLOADING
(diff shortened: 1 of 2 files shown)


spophale (approved):

krishols (approved):

tests/5.0/atomic/test_atomic_num_hint_device.F90
```diff
@@ -0,0 +1,85 @@
...
+!  This test checks if atomic hints passed by enum value
+!  are processed correctly for device execution. If the sync hint is not
+!  yet defined in the specification, it defaults to 
+!  omp_sync_hint_none (0x0). 
+!
+!//===---------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
```
krishols, line 14: I think this is unused

tests/5.0/atomic/test_atomic_num_hint_device.c
```diff
@@ -17,7 +17,7 @@
 #define N 1024
```
krishols, line 17: Same here - I think this is unused

tests/5.0/atomic/test_atomic_num_hint_device.F90
seyonglee: Nice catch!

tests/5.0/atomic/test_atomic_num_hint_device.c
seyonglee: Thanks!

Issue #649, opened 2023-01-02
[5.1] Add test for 'map' clause with 'present' modifier


tob2:
For OpenMP 5.1, please add a test for:
* `map(present, alloc: var)` — that is the `present` modifier.
  * Note that the _map-type-modifier_ needs to be present as well (and last in the modifier list), even if it is effectively ignored.
  * While the `alloc` _map-type-modifier_ makes most sense, also all others, i.e. from/to/tofrom/release/delete, are also permitted.
  * Please add test for both `target` and `target (enter/exit) data`

Notes:
* There are already a test for `defaultmap(present` and for `target update to(present:`, but not for `map`.
  Additionally, currently only C is tested for – **Fortran tests cases are missing** (as is `target update from(present:`).
  Available tests:
  * https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.1/target/test_target_defaultmap_present.c
  * https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.1/target/test_target_defaultmap_present_scalar.c
  * https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.1/target_update/test_target_update_to_present.c
* In OpenMP TR11 (technical report and first OpenMP 6.0 preview), the _map-type-modifier_ can be left out (and otherwise freely placed instead of being last), i.e. `map(present: var)` and `map(tofrom , present: var)` are valid since TR11 _(the former makes sense, the latter is valid but odd)._

@seyonglee @fel-cab @krishols @jrreap — in case you have time and feel like adding it.

tob2:
While the C testcase was merged via Pull Request #687, the Fortran pull request is #721

PR #687, opened 2023-03-01
Add 5.1 test for present map type modifier

krishols:
Adapted from 5.0 test for close map type modifier

### DIFF
--- a/tests/5.1/target/test_target_map_with_present_modifier.c
+++ b/tests/5.1/target/test_target_map_with_present_modifier.c
@@ -34,8 +34,9 @@ int test_present_modifier() {
     a[i] = i;
     member.b[i] = i;
   }
-
-#pragma omp target map (present, tofrom: scalar, a, member) 
+# pragma omp target data map(tofrom:scalar, a, member) 
+  {
+#pragma omp target map(present, tofrom: scalar, a, member) 
   {
     scalar += 1;
     member.var += 2;
@@ -44,6 +45,7 @@ int test_present_modifier() {
       a[i] += i;
       member.b[i] += i;
     }
+  }
   }
 
   for (i = 0; i < N; i++) {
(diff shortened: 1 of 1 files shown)


tests/5.1/target/test_target_map_with_present_modifier.c
```diff
@@ -0,0 +1,67 @@
...
+  } member; 
+
+  member.var = 1;
+  
+  for (i = 0; i < N; i++) { 
+    a[i] = i;
+    member.b[i] = i;
+  }
+
+#pragma omp target map (present, tofrom: scalar, a, member)
```
tob2, line 38: This testcase has a bare **target** with **present** modifier. Thus, the following applies:

> If a **map** clause appears on a **target**, **target data**, **target enter data** or **target exit data** construct with a **present** _map-type-modifier_ then on entry to the region if the corresponding list item does not appear in the device data environment then an error occurs and the program terminates.

As in sollve_vv, there is no real infrastructure to test for should-fail tests, I assume the testcase should run. But in that case, the **target** construct must either be inside a **target data** region that maps the variables or the data must be mapped before with **target enter data**. 

(Or a combination of both: some variables one way, some the other way.)

* For a **target exit data**, the **present** modifier could also be used as that should work, given the variables have been mapped before.
* For the **target** construct, an **always** modifier would be required in addition, if the map-to value has been modified between the 'target (enter) data' and the 'target' — and when the result should be checked before the 'target exit data' or insie the 'target data' region, respectively.

nolanbaker31: Might be useful to refer to 5.1/target_update/test_target_update_to_present.c

nolanbaker31:
Closes #649

krishols:
Passes with LLVM

spophale (approved):

mjcarr458 (approved):
looks good to me.

PR #688, opened 2023-03-01
Target Update tests

spophale:
Adding tests to check target update to and from arrays when stride is specified.


fel-cab:
I checked them with LLVM and GCC, they compile with LLVM but not with GCC. However, LLVM updates the whole array, ignoring the stride.

seyonglee (approved):
The test codes look OK to me.

fel-cab (approved):

PR #689, opened 2023-03-06
indirect declare target test

mjcarr458:
Start of working on indirect declare target test. Here I copied down the basic structure provided by Swaroop in slack, but I still have a couple questions regarding it. Most importantly it uses the enter clause which was noted as a 5.2 feature. Since this is a 5.1 test is the a different work around I should use. Any suggestions or help here would be great appreciated.

### DIFF
--- a/tests/5.1/target/test_target_declare_indirect.c
+++ b/tests/5.1/target/test_target_declare_indirect.c
@@ -1,3 +1,12 @@
+//----------------test_target_declare_indirect.c-----------------
+//
+// OpenMP API Version 5.1 Mar 2023
+//
+// This test intends to test the functionality of an indirect
+// declare target region. A random int is chosen that correlates
+// to a function pointer. That function is then chosen and the 
+// test ensures it runs as expected.
+// --------------------------------------------------------------
 #include <omp.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


spophale:
@mjcarr458  - See my edits.

tests/5.1/target/test_target_declare_indirect.c
```diff
@@ -0,0 +1,61 @@
...
+}
+
+int fun2(){
+	return 10;
+}
+
+int fun3(){
+	return 15;
+}
+#pragma omp begin declare enter(fun1, fun2, fun3) indirect
```
tob2, line 20:
```suggestion
#pragma omp declare target enter(fun1, fun2, fun3) indirect
```
(or 'begin/end declare target' around the functions but then without 'enter' clause)

spophale (approved):

fel-cab:
Is it necessary to have 3 functions and a rand to test the indirect?. I think the rand creates a not necessary complexity. I would change it to use the three functions and increment the values, instead of choosing one of them with a rand.

mjcarr458:
Change enter() to to()

fel-cab (approved):

PR #690, opened 2023-03-07
Fix test_omp_target_aligned_alloc_device.c

tob2:
* Add missing 'write' clause to 'omp atomic'
* Fix typo in omp_destroy_allocator
* Fix warning with -DVERBOSE_MODE as '%' (modulo) of the condition is processed by printf.

 @nolanbaker31 @spophale @jrreap @krishols @mjcarr458 @seyonglee – please review

### DIFF
--- a/tests/5.1/allocate/test_omp_target_aligned_alloc_device.c
+++ b/tests/5.1/allocate/test_omp_target_aligned_alloc_device.c
@@ -35,7 +35,10 @@ int test_omp_aligned_alloc_on_device() {
       OMPVV_ERROR("omp_aligned_alloc returned null"); 
       errors++;
     } else {
-      OMPVV_TEST_AND_SET_VERBOSE(errors, ((intptr_t)(x))%64 != 0);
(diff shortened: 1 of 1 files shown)


spophale (approved):

seyonglee:
1) OpenMP Specification V5.1 section **2.14.5 target Construct** (page 197, line 29) says "allocator-traits-array is an identifier of const omp_alloctrait_t * type", and thus the type of below `traits` array should be `const omp_alloctrait_t`.

`omp_alloctrait_t       traits[1] = {{omp_atk_alignment, 64}};`

2) `uses_allocators` clause has the following syntax:

`uses_allocators(allocator[(allocator-traits-array)] [,allocator[(allocator-traits-array)] …])`
 
Therefore, `#pragma omp target map(tofrom: errors) uses_allocators(alloc[traits])` should be changed to `#pragma omp target map(tofrom: errors) uses_allocators(alloc(traits)) `

jrreap (approved):
Good catch with the printf, also updated the target pragma to apply the trait array correctly as well. Good to go.

PR #692, opened 2023-03-08
Add new_test/5.0/test_taskloop_reduction.F90

seyonglee:
- NVHPC 22.11:
            - C test failed: line 33: error: this OpenMP construct is not supported in NVIDIA subset: The 'taskloop' construct is not supported.
            - Fortran test failed: NVFORTRAN-F-1235-The OpenMP 'taskloop' construct is not supported.
        - LLVM 15.0.0: C test passed.
        - LLVM 17.0.0: C test passed.
        - GCC 12.2.1:
        - XL 16.1.1-10:
            - C test passed but the test ran with one thread, so parallelism of taskloop can't be guaranteed.
            - Fortran test failed: line 40.23: 1515-022 (S) Syntax Error: Extra token " reduction " was found. The token is ignored.

### DIFF
--- a/tests/5.0/taskloop/test_taskloop_reduction.F90
+++ b/tests/5.0/taskloop/test_taskloop_reduction.F90
@@ -1,0 +1,75 @@
+!//===------ test_taskloop_reduction.F90 ----------------------------------===//
+!
+! OpenMP API Version 5.0 Nov 2018
+!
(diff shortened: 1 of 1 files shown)


spophale (approved):

mjcarr458 (approved):
Looks good to me

PR #696, opened 2023-03-15
Aligned calloc routine (host and device)

jrreap:
Adds test for the aligned variant of the calloc routine. Tests both on device and host.

### DIFF
--- a/tests/5.1/allocate/test_aligned_calloc.c
+++ b/tests/5.1/allocate/test_aligned_calloc.c
@@ -20,14 +20,10 @@
 int test_aligned_calloc() {
   int errors = 0;
   
-  omp_memspace_handle_t  memspace = omp_default_mem_space;
-  omp_alloctrait_t       traits[1] = {{omp_atk_alignment, 64}};
-  omp_allocator_handle_t alloc = omp_init_allocator(memspace,1,traits);
-
   int *x;
   int not_correct_array_values = 0;
 
-  x = (int *)omp_aligned_calloc(64, N, N*sizeof(int), alloc);
+  x = (int *)omp_aligned_calloc(64, N, N*sizeof(int), omp_default_mem_alloc);
 
   if (x == NULL) { 
     OMPVV_ERROR("omp_aligned_calloc returned null"); 
@@ -54,10 +50,9 @@ int test_aligned_calloc() {
       errors++;
     }
 
-    omp_free(x, alloc);
+    omp_free(x, omp_default_mem_alloc);
   }
 
-  omp_destroy_allocator(alloc);
 
   return errors;
 }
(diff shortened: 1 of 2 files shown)


jrreap:
`/autofs/nccs-svm1_home1/jreap/sollve_vv/tests/5.1/allocate/test_target_aligned_calloc.c:27:63: error: called object type 'omp_allocator_handle_t' (aka 'enum omp_allocator_handle_t') is not a function or function pointer 

#pragma omp target map(tofrom: errors) uses_allocators(alloc(traits))`

Throws this error for device test, seems confused by the custom allocator traits

tests/5.1/allocate/test_aligned_calloc.c
```diff
@@ -0,0 +1,73 @@
...
+    omp_free(x, alloc);
+  }
+
+  omp_destroy_allocator(alloc);
+
+  return errors;
+}
+
+int main() {
+  OMPVV_TEST_OFFLOADING;
```
fel-cab, line 66: This is a host test, so we should remove the `OMPVV_TEST_OFFLOADING`

tests/5.1/allocate/test_aligned_calloc.c
```diff
@@ -0,0 +1,73 @@
...
+#include "ompvv.h"
+
+#define N 1024
+
+int test_aligned_calloc() {
+  int errors = 0;
+  
+  omp_memspace_handle_t  memspace = omp_default_mem_space;
+  omp_alloctrait_t       traits[1] = {{omp_atk_alignment, 64}};
+  omp_allocator_handle_t alloc = omp_init_allocator(memspace,1,traits);
```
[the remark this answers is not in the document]
spophale: Why ?

spophale:
We need to check if %64 is the correct way to go to check alignment requirement. Spec says:
```
Memory allocated by omp_aligned_calloc will be byte-aligned to at least the maximum of
the alignment required by malloc, the alignment trait of the allocator and the alignment argument value.
```

tests/5.1/allocate/test_aligned_calloc.c
jrreap: Good catch! I'll patch it up.

jrreap:
The %64 actually comes from their `omp_alloc` example in the 5.1 examples document. They are passing special alignment parameters to the allocator and check if it's properly aligned in a similar way. It's in section 10.2 "Memory Allocators" (page 426 in the online PDF)

jrreap:
Get nvlink errors with cuda 11.5.2 and llvm 17.0.0-latest on summit. Seems offloading may just not be supported yet?

`nvlink error   : Undefined reference to 'omp_aligned_calloc' in '/tmp/test_target_aligned_calloc-cadfa3-nvptx64-nvidia-cuda-sm_70-d4219b.cubin'`

jrreap:
Both tests run as expected and pass on gcc 12.2.1 with cuda 11.5.1. Looks like clang just does not support the aligned on device quite yet. Should be good to go now.

spophale (approved):

fel-cab (approved):

PR #698, opened 2023-03-20
Changing test so the device memory update is discontigous

fel-cab:


tests/5.0/target_update/test_target_update_mapper_to_discontiguous.c
```diff
@@ -32,26 +32,35 @@ int target_update_to_mapper() {
...
+    }
+    // update even array position values from host
+    // This should set them to i  
+    #pragma omp target update to(s.data[0:s.len:2])
 
-#pragma omp target update to(s)
+    //update array on the device
+    #pragma omp target //map(alloc: result[0:N])
+    for (i = 0; i < s.len; i++) {
+      s.data[i] = i;
```
spophale, line 52: This needs to be `s.data[i] += i` and not an assignment.

fel-cab: Fixed, thanks

fel-cab:
Related LLVM issue: https://github.com/llvm/llvm-project/issues/61452

seyonglee (approved):
It looks OK to me, even though this test fails with all compilers on Summit:
        - NVHPC 22.11: C test failed (compile-time error).
        - LLVM 15.0.0: C test failed (compile-time error).
        - GCC 12.2.1: C test failed (compile-time error).
        - XL 16.1.1-10: C test failed at runtime (The value of errors is 512)

PR #701, opened 2023-04-04
Ordered Doacross - Defined Source

nolanbaker31:


tests/5.2/ordered/test_ordered_doacross_defined_source.c
```diff
@@ -0,0 +1,62 @@
...
+  int c[N];
+  a[0] = 0;
+  b[0] = 0;
+  c[0] = 0;
+
+  #pragma omp parallel for ordered
+  for(int i = 1; i < N; i++){
+    a[i] = i;
+    #pragma omp ordered doacross(sink: i-2)
+    b[i-1] = a[i-2];
```
fel-cab, line 43: for i=1, a[i-2] is outside the array

andrewkallai:
Because (source: ) is logically equivalent to (source: omp_cur_iteration) according to the semantics of source on the doacross clause, and according to the restrictions on the doacross clause: "If vector is specified with source as the dependence-type then it must be omp_cur_iteration," is this test necessary? tests/5.2/ordered/test_ordered_doacross.c uses (source: ) for line 44 where source is the dependence-type on the doacross clause. So in that test it is undefined, but equivalent to defining it as (source: omp_cur_iteration).

andrewkallai:
frontier
clang 18.0.0
42:25: error: 'ordered' directive with 'doacross' clause cannot be closely nested inside ordered region without specified parameter
   42 |     #pragma omp ordered doacross(sink: i-2)
/autofs/nccs-svm1_home1/andrewka/sollve_vv/tests/5.2/ordered/test_ordered_doacross_defined_source.c:44:41: error: expected ')'
   44 |     #pragma omp ordered doacross(source:i-1)
/autofs/nccs-svm1_home1/andrewka/sollve_vv/tests/5.2/ordered/test_ordered_doacross_defined_source.c:44:33: note: to match this '('
   44 |     #pragma omp ordered doacross(source:i-1)
/autofs/nccs-svm1_home1/andrewka/sollve_vv/tests/5.2/ordered/test_ordered_doacross_defined_source.c:44:25: error: 'ordered' directive with 'doacross' clause cannot be closely nested inside ordered region without specified parameter
   44 |     #pragma omp ordered doacross(source:i-1)
3 errors generated.

tests/5.2/ordered/test_ordered_doacross_defined_source.c
```diff
@@ -0,0 +1,62 @@
...
+  a[0] = 0;
+  b[0] = 0;
+  c[0] = 0;
+
+  #pragma omp parallel for ordered
+  for(int i = 1; i < N; i++){
+    a[i] = i;
+    #pragma omp ordered doacross(sink: i-1)
+    b[i] = a[i-1];
+    #pragma omp ordered doacross(source:i)
```
andrewkallai, line 44:
```suggestion
    #pragma omp ordered doacross(source:omp_cur_iteration)
```

andrewkallai (changes requested):
Since omp_cur_iteration is not specified for source, the compiler will error. However, if it is specified, then GCC will accept the clause. Because an undefined source is equivalent to source:omp_cur_iteration, this test is equivalent to the undefined source test. So unless this explicit change is desired to be tested, this test should be dropped.

spophale:
We already have tests with omp_cur_iteration.

PR #703, opened 2023-04-05
Declare Mapper w/ Present 

nolanbaker31:
Fails on GCC (not implemented)

Fails on LLVM 17 w/ message:
`Libomptarget message: device mapping required by 'present' map type modifier does not exist for host address 0x000000011df198e0 (8000 bytes) `

`Libomptarget error: Call to getTargetPointer returned null pointer ('present' map type modifier). `

`Libomptarget error: Call to targetDataBegin via targetDataMapper for custom mapper failed.`

### DIFF
--- a/tests/5.2/declare_mapper/test_declare_mapper_present.c
+++ b/tests/5.2/declare_mapper/test_declare_mapper_present.c
@@ -23,7 +23,7 @@ typedef struct myvec{
 } myvec_t;
 
 
-#pragma omp declare mapper(myvec_t v) map(present, tofrom: v, v.data[0:v.len]) 
+#pragma omp declare mapper(default:myvec_t v) map(present, tofrom: v, v.data[0:v.len]) 
 
 void init( myvec_t *s )
 { 
--- a/tests/5.2/declare_mapper/test_declare_mapper_present.c
+++ b/tests/5.2/declare_mapper/test_declare_mapper_present.c
@@ -22,6 +22,20 @@ typedef struct myvec{
     int *data;
 } myvec_t;
 
+// In addition to the present modifier, the always is requiered
+// Given that the present and default modifiers are to be tested
+// the struct has to be in the GPU prior to the default mapper
+// is tested, therefore the "omp target enter data" with "to"
+// is used. Furthermore, given that it is the "from" that is to be
+// checked, the "omp target exit data" is set to "delete"
+// in this scenario if the always is not on the declare mapper
+// the "omp target" is inside the "omp target enter data"
+// with the "map to", which would prevent the effect of the
+// declared mapped to have effects outside. Therefore, by adding
+// the always, the mapper force the update of the declared mapper
+// to be seen outside the "target data region"
+
+#pragma omp declare mapper(default:myvec_t v) map(always, present, from: v, v.len, v.data[0:v.len]) 
 
 
 void init( myvec_t *s )
@@ -39,16 +53,15 @@ int test_declare_mapper_present() {
   myvec_t s;
   s.data = (int *)calloc(N,sizeof(int));
   s.len  = N;
-  #pragma omp target enter data map(to: s, s.data[0:s.len])
 
-  #pragma omp declare mapper(default:myvec_t v) map(present, from: v, v.data[0:v.len]) 
-
+  #pragma omp target enter data map(to: s.len, s.data[0:s.len])
 
   #pragma omp target
   {
     init(&s);
   }
-  #pragma omp target exit data map(from: s, s.data[0:s.len])
+
+  #pragma omp target exit data map(delete: s.len, s.data[0:s.len])
 
   for (int i = 0; i < N; ++i) {
     OMPVV_TEST_AND_SET(errors, s.data[i] != i);
--- a/tests/5.2/declare_mapper/test_declare_mapper_present.c
+++ b/tests/5.2/declare_mapper/test_declare_mapper_present.c
@@ -19,15 +19,14 @@
 
 typedef struct myvec{
     size_t len;
-    double *data;
+    int *data;
 } myvec_t;
 
 
-#pragma omp declare mapper(default:myvec_t v) map(present, tofrom: v, v.data[0:v.len]) 
 
 void init( myvec_t *s )
 { 
-  for(size_t i = 0; i < s->len; i++)
+  for(int i = 0; i < s->len; i++)
     s->data[i] = i; 
 }
 
@@ -37,16 +36,17 @@ int test_declare_mapper_present() {
   OMPVV_INFOMSG("test_declare_mapper_present");
   int errors = 0;
 
-   myvec_t s;
+  myvec_t s;
 
-   s.data = (double *)calloc(N,sizeof(double));
-   s.len  = N;
-  #pragma omp target data map(tofrom: s, s.data[0:s.len])
+  s.data = (int *)calloc(N,sizeof(int));
+  s.len  = N;
+  #pragma omp target enter data map(to: s, s.data[0:s.len])
+
+  #pragma omp declare mapper(default:myvec_t v) map(present, from: v, v.data[0:v.len]) 
+
+  #pragma omp target map(s)
   {
-    #pragma omp target
-    {
-        init(&s);
-    }
+    init(&s);
   }
 
   for (int i = 0; i < N; ++i) {
--- a/tests/5.2/declare_mapper/test_declare_mapper_present.c
+++ b/tests/5.2/declare_mapper/test_declare_mapper_present.c
@@ -37,17 +37,18 @@ int test_declare_mapper_present() {
   int errors = 0;
 
   myvec_t s;
-
   s.data = (int *)calloc(N,sizeof(int));
   s.len  = N;
   #pragma omp target enter data map(to: s, s.data[0:s.len])
 
   #pragma omp declare mapper(default:myvec_t v) map(present, from: v, v.data[0:v.len]) 
 
-  #pragma omp target map(s)
+
+  #pragma omp target
   {
     init(&s);
   }
+  #pragma omp target exit data map(from: s, s.data[0:s.len])
 
   for (int i = 0; i < N; ++i) {
     OMPVV_TEST_AND_SET(errors, s.data[i] != i);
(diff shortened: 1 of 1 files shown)


nolanbaker31:
Test still fails on LLVM even after adding mapper(default:...)

spophale:
@nolanbaker31 which LLVM compiler are you using ?

nolanbaker31:
> @nolanbaker31 which LLVM compiler are you using ?

Clang nightly build, 17.0.0-20230417

tests/5.2/declare_mapper/test_declare_mapper_present.c
```diff
@@ -0,0 +1,67 @@
...
+int test_declare_mapper_present() { 
+
+  OMPVV_INFOMSG("test_declare_mapper_present");
+  int errors = 0;
+
+   myvec_t s;
+
+   s.data = (double *)calloc(N,sizeof(double));
+   s.len  = N;
+  #pragma omp target data map(tofrom: s, s.data[0:s.len])
```
spophale, line 44: remove "from"

spophale: you are right!

andrewkallai:
frontier
clang 18.0.0
FAIL. exit code: 134
test_declare_mapper_present.c.o:
Libomptarget message: device mapping required by 'present' map type modifier does not exist for host address 0x000055f3717bf870 (8000 bytes)
Libomptarget error: Call to getTargetPointer returned null pointer ('present' map type modifier).
Libomptarget error: Call to targetDataBegin via targetDataMapper for custom mapper failed.
Libomptarget error: Consult https://openmp.llvm.org/design/Runtimes.html for debugging options.
Libomptarget error: Source location information not present. Compile with -g or -gline-tables-only.
Libomptarget fatal error 1: failure of target construct while offloading is mandatory
timeout: the monitored command dumped core

tests/5.2/declare_mapper/test_declare_mapper_present.c
```diff
@@ -0,0 +1,67 @@
...
+  {
+    #pragma omp target
+    {
+        init(&s);
+    }
+  }
```
fel-cab, line 50:
```suggestion
  #pragma omp target map(s)
  {
      init(&s);
  }

```
As the test is written, I consider that it is not testing that the `declare mapper` is working. It is only testing if the compiler accepts the directive is recognized by the compiler. This is because the mapper is not used. With this modification the mapper is used.

seyonglee: I guess that the reason for having the nesting `target data` construct is to test the `present` modifier in line 26.
If we remove the nesting `target data` construct, the compiled program will fail at runtime due to the `present` requirement.

tests/5.2/declare_mapper/test_declare_mapper_present.c
```diff
@@ -0,0 +1,67 @@
...
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+#define N 1000
+
+typedef struct myvec{
+    size_t len;
+    double *data;
```
andrewkallai, line 22: Change all double types to ints where double is used as this can potentially be an issue for value comparison.

tests/5.2/declare_mapper/test_declare_mapper_present.c
```diff
@@ -0,0 +1,67 @@
...
+
+int test_declare_mapper_present() { 
+
+  OMPVV_INFOMSG("test_declare_mapper_present");
+  int errors = 0;
+
+   myvec_t s;
+
+   s.data = (double *)calloc(N,sizeof(double));
+   s.len  = N;
```
andrewkallai, line 43: after this line I would add #pragma omp target enter data map(to: s, s.data[0:s.len]) to satisfy the requirement of the struct type being present on the device prior to executing the mapper.

tests/5.2/declare_mapper/test_declare_mapper_present.c
```diff
@@ -0,0 +1,67 @@
...
+
+#define N 1000
+
+typedef struct myvec{
+    size_t len;
+    double *data;
+} myvec_t;
+
+
+#pragma omp declare mapper(default:myvec_t v) map(present, tofrom: v, v.data[0:v.len])
```
andrewkallai, line 26: I would put this to replace line 44 under my target enter data suggesstion. Thus, the mapper is declared after the struct has already been mapped to the device, otherwise the map would fail as it would inherently require itself to be present. Change the tofrom to from as the the struct has already been mapped to the device.

andrewkallai (changes requested):

tests/5.2/declare_mapper/test_declare_mapper_present.c
andrewkallai: I would agree with the suggested change as it relies on the declare mapper characteristics and does not influence how s is mapped.

tests/5.2/declare_mapper/test_declare_mapper_present.c
```diff
@@ -0,0 +1,67 @@
...
+
+  myvec_t s;
+
+  s.data = (int *)calloc(N,sizeof(int));
+  s.len  = N;
+  #pragma omp target enter data map(to: s, s.data[0:s.len])
+
+  #pragma omp declare mapper(default:myvec_t v) map(present, from: v, v.data[0:v.len]) 
+
+  #pragma omp target map(s)
```
spophale, line 47: Do we need map(s) here ?

spophale: map(s) does the same, without a explicit modifier it is "tofrom".

tob2: But a bare `map(tofrom: s)` won't help here, either – I mean because it is already present?!?

Thus, before testing, IMHO
* either an "**always**" modifier is needed in the mapper or the usage of the mapper. _(I think it is permitted, but I vaguely remember some more recent changes; either in 5.2 or later, not checked.)_ 
* Or after the target region, the ref count has to be reduced to force the copy out (like: **target exit data**)
* or the data has to be explicitly be copied (i.e. using the **target update** directive).

seyonglee: OpenMP V5.1 Section 2.21.7.1 map Clause says:

> If the mapper map-type-modifier is not present, the behavior is as if the mapper(default) modifier was specified. 

This means that the `mapper(default)` will be applied to the `map` clause by default.

seyonglee: @tob2 I agree that this test will fail because it is already present.
(Test in line 53 will fail as commented below.)

tests/5.2/declare_mapper/test_declare_mapper_present.c
```diff
@@ -0,0 +1,67 @@
...
+  {
+    init(&s);
+  }
+
+  for (int i = 0; i < N; ++i) {
+    OMPVV_TEST_AND_SET(errors, s.data[i] != i);
```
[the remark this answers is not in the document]
andrewkallai: I disagree that the present modifier would prohibit the from map-type from taking effect, as the struct is present on the device and the declared mapper can then accordingly substitute the given map clause with the mapper's contents. However, the mapper is doing a map-type of from for the struct, and an array section. Because the default mapping for a struct of that type has been rewritten, the map-type decay for tofrom -> from = from. However the array section has its own default mapping behavior, which I do not understand from pg. 156 of 5.2:
23 If an array appears as a list item in a map clause, multiple parts of the array have corresponding
24 storage in the device data environment prior to a task encountering the construct associated with
25 the map clause, and the corresponding storage for those parts was created by maps from more
26 than one earlier construct, the behavior is unspecified.
Is this the case for the array section in question?

fel-cab:
@andrewkallai and I tested the example and the problem (at least on LLVM) seems to be that the declare mapper at the end of the kernel cannot handle the `v.data[0:v.len]` correctly in combination with the `target data enter`. We made the following test and it does work.
`
#include <omp.h>
#include <stdio.h>
#include <stdlib.h>
#include "ompvv.h"
typedef struct myvec{
    int I01;
    int I02;
    int I03;
    int I04;
} myvec_t;
void init( myvec_t *s )
{ 
  s->I01 = 1;
  s->I02 = 2;
  s->I03 = 3;
  s->I04 = 4;
}
int test_declare_mapper_present() { 
  OMPVV_INFOMSG("test_declare_mapper_present");
  int errors = 0;
  myvec_t s;
  s.I01 = 0;
  s.I02 = 0;
  s.I03 = 0;
  s.I04 = 0;
  #pragma omp target enter data map(to: s) 
  #pragma omp declare mapper(default:myvec_t v) map(present, from: v) 
  #pragma omp target 
  {
    init(&s);
  }
  OMPVV_TEST_AND_SET(errors, (s.I01 != 1) && (s.I02 != 2)&& (s.I03 != 3)&& (s.I04 != 4) );
  // the following is required to delete the data on the device
  #pragma omp target exit data map(delete: s)
  return errors;
}

int main () {
  OMPVV_TEST_OFFLOADING;
  int errors = 0;
  OMPVV_TEST_AND_SET_VERBOSE(errors, test_declare_mapper_present());
}  
`

spophale:
Test failing due to the positioning of the delclare mapper with LLVM is confusing. More investigation needed.

andrewkallai:
Code seems to accomplish from mapping as desired:
```
#include <stdio.h>
#include <stdlib.h>

#define N 1000

typedef struct myvec {
  // size_t len;
  int *data;
} myvec_t;

void init(myvec_t *s) {
  for (int i = 0; i < N; i++)
    s->data[i] = i;
}

int main() {
  myvec_t s;

  s.data = (int *)calloc(N, sizeof(int));
// s.len  = N;
#pragma omp target enter data map(to : s, s.data[0 : N])

#pragma omp declare mapper(default : myvec_t v) \
    map(present, from : v, v.data[0 : N])

#pragma omp target //map(s)
  {
    init(&s);
  }
#pragma omp target map(s)
  printf("%i\n", s.data[1]);
}
```

fel-cab:
I have modified the test with the `always` modifier on the declare mapper. And now it works. With LLVM trunk

spophale (approved):

PR #710, opened 2023-04-14
test for affinity_format env variable added, with change in Makefile …

fel-cab:
…and adjusment to make.def

The test runs with the modified Makefile with clang 17.0.0


spophale:
Outdated.

PR #711, opened 2023-04-19
Add new_test/5.1/test_target_defaultmap_present_pointer.F90

seyonglee:
- GCC 12.2.1:
            - C test passed.
            - Fortran test failed: libgomp: nvptx_alloc error: out of memory
                - If `ptr => A` is added before target enter directive, the following error occurs:
                    [OMPVV_INFOMSG ompvv.F90:249] The value of errors is 1025 
        - XL 16.1.1-10:
            - Fortran test failed: line 42.38: 1515-022 (S) Syntax Error: Extra token " defaultmap " was found. The token is ignored.
        - NVHPC 22.11:
            - C test failed: NVC++-F-1196-OpenMP - 'defaultmap(present)' behaviour is not supported yet.
            - Fortran test failed: NVFORTRAN-S-1197-OpenMP GPU - presents cannot be used with defaultmap clause.
        - LLVM 17.0.0: C test passed.

### DIFF
--- a/tests/5.1/target/test_target_defaultmap_present_pointer.F90
+++ b/tests/5.1/target/test_target_defaultmap_present_pointer.F90
@@ -1,0 +1,61 @@
+!===--- test_target_defaultmap_present_pointer.F90 ---------------------- -===//
+!
+! OpenMP API Version 5.1 Nov 2021
+!
+! This test checks behavior of the defaultmap clause when the specified 
+! implicit-behavior is present. The variable-categories available for defaultmap
+! are scalar, aggregate, and pointer. If implicit-behavior is present, each 
+! variable referenced in the construct in the category specified by 
+! variable-category is treated as if it had been listed in a map clause with the
+! map-type of alloc and map-type-modifier of present.
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_target_defaultmap_present_pointer
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+  OMPVV_TEST_OFFLOADING
+
+  OMPVV_TEST_VERBOSE(defaultmap_present_pointer() .NE. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+
+CONTAINS
+  INTEGER FUNCTION defaultmap_present_pointer()
+    INTEGER :: errors, i
+    INTEGER, TARGET, DIMENSION(N) :: A !aggregate
+    INTEGER, POINTER :: ptr(:) !pointer
+
+    errors = 0
+
+    DO i = 1, N
+        A(i) = i
+    END DO
+
+    !$omp target enter data map(to: ptr)
+    !$omp target map(tofrom: errors) defaultmap(present:pointer)
+    ptr => A
+    DO i = 1, N
+        IF (ptr(i) .NE. i) then
+            errors = errors + 1
+        END IF 
+        ptr(i) = i + 2
+    END DO
+    !$omp end target
+    !$omp target exit data map(delete: ptr)
+
+    OMPVV_ERROR_IF(errors .GT. 0, "Values were not mapped to the device properly")
+
+    DO i = 1, N
+        OMPVV_TEST_AND_SET(errors, A(i) .NE. 2+i)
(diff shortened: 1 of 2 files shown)


seyonglee:
GCC Fortran compiler fails with the following error message: `libgomp: nvptx_alloc error: out of memory`, but it is unclear whether this error is caused by bugs in the test program or a compiler bug. The out-of-memory error occurs when an unassociated Fortran pointer is passed to the target region.

OpenMP V5.1 Section 2.21.7.1 map Clasue, page 351 says "If the reference count of the corresponding list item is one or the always map-type-modifier is present, and if the map-type is to or tofrom, then: ... 
For each part of the list item that is not an attached pointer, the value of that part of the original list item is assigned to that part of the corresponding list item."

spophale:
@tob2 Do you have any insights as to the errors we see here ?

seyonglee:
[Update on July 11, 2023]
GCC 13.1.1: Fortran test fails: libgomp: nvptx_alloc error: out of memory
    libgomp: device memory allocation fail

spophale:
Tested on Summit (Core module gcc/13.2.1-20231129). I get the following error:
```
libgomp: nvptx_alloc error: out of memory
libgomp: device memory allocation fail
```

spophale (approved):

fel-cab (approved):

tob2:
In December:
> @tob2 , will it be possible to check with latest GCC ?

Sorry for missing this. :-/

Testing on my laptop with nvptx offloading:

* Using today's **GCC mainline** (**GCC 14**; _to be released in about 2 months_) → it **works** but only ...

... after the following change, required as otherwise the ptr target data is not copied back to the host (required unless  it  _self_ mapping / (unified) shared memory / host fallback):
```diff
--- a/tests/5.1/target/test_target_defaultmap_present_pointer.F90
+++ b/tests/5.1/target/test_target_defaultmap_present_pointer.F90
@@ -51 +51 @@ CONTAINS
-    !$omp target exit data map(delete: ptr)
+    !$omp target exit data map(from: ptr)
```

* Using plain **GCC 13** (`gcc (SUSE Linux) 13.2.1 20240206`) **fails** as `defaultmap(present` is not supported.

* Using **OG13** (`GNU Fortran (GCC) 13.2.1 20231026 [OG13]`) **works** here: 
`[OMPVV_RESULT test_target_defaultmap_present_pointer.F90] Test passed on the device.` 

_(OG13 = vendor branch, based on the GCC 13 branch with some local changes only in mainline or not yet in mainline. The shown date is the one of the last merge from the GCC 13 branch. Last commit in my case if from Dec 20, 2023.)_

Yesterday:
> Tested on Summit (Core module gcc/13.2.1-20231129). I get the following error:
> ```
> libgomp: nvptx_alloc error: out of memory
> libgomp: device memory allocation fail
> ```

Hmm, I wonder why that result is different to mine — Summit surely has surely more memory than my laptop card and the testcase looks unsuspicious.

And contrary to GCC mainline, OG13 did not see any relevant changes between "20231129" (= probably the build date at OLCF) and today / my OG13 build date [mainly a couple of 'omp allocate/allocators' and AMD GPU changes – but Summit is Nvidia]. (Nor did GCC 13 change much, but as 'defaultmap(present:' is not on plain GCC 13, I assume that Summit's GCC 13 is actually from the OG13 branch.)

Thus:
* Test case works for me (after the obvious fix) — with both plain GCC 14 and with OG13.
*  I have no spontaneous idea what goes wrong on Summit.

tschwinge:
I, too, don't have any immediate idea what's going wrong.

Please tell us your specific GCC revision: `gcc --version | head -n 1`.

Given the error message, the relevant code must be: <https://gcc.gnu.org/cgit/gcc/tree/libgomp/target.c?h=devel/omp/gcc-13&id=a6af434d6b26666ceddc56cc4459019929372c31#n1533>, that is the main memory mapping routine, `gomp_map_vars_internal`.  If you're able to, does putting a GDB breakpoint on the "Memory allocation failed" case in the libgomp nvptx plugin: <https://gcc.gnu.org/cgit/gcc/tree/libgomp/plugin/plugin-nvptx.c?h=devel/omp/gcc-13&id=a6af434d6b26666ceddc56cc4459019929372c31#n1724> give any clue, what `size` allocation fails etc.?

PR #712, opened 2023-04-19
New test/5.1/test target defaultmap present scalar.f90

seyonglee:
- GCC 12.2.1: Both C and Fortran tests passed.
        - XL 16.1.1-10:
            - C test failed: The value of errors is 4.
            - Fortran test failed: line 41.38: 1515-022 (S) Syntax Error: Extra token " defaultmap " was found. The token is ignored.
        - NVHPC 22.11:
            - C test failed: NVC++-F-1196-OpenMP - 'defaultmap(present)' behaviour is not supported yet. 
            - Fortran test failed: NVFORTRAN-S-1197-OpenMP GPU - presents cannot be used with defaultmap clause.
        - LLVM 17.0.0:
            - C test failed: The value of errors is 4.

### DIFF
--- a/tests/5.1/target/test_target_defaultmap_present_scalar.F90
+++ b/tests/5.1/target/test_target_defaultmap_present_scalar.F90
@@ -31,11 +31,16 @@ CONTAINS
     INTEGER :: errors, scalar_var
     REAL :: float_var
     DOUBLE PRECISION :: double_var
+    LOGICAL :: is_shared_env
 
     errors = 0
+    is_shared_env = .false.
     scalar_var = 1
     float_var = 10.7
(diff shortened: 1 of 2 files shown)


spophale:
On a shared memory system this test (both C and Fortran) could get in trouble by modifying the variables that are only mapped to the device.

seyonglee:
> On a shared memory system this test (both C and Fortran) could get in trouble by modifying the variables that are only mapped to the device.

Added codes to check whether the test is running on a shared data environment and add a warning message and skip some error-checking codes if running on the shared data environment.

spophale (approved):

fel-cab (approved):

PR #717, opened 2023-04-19
Add new_test/5.1/test_target_has_device_addr.F90

seyonglee:
[Work in progress]
        - GCC 12.2.1:
            - C test passed.
            - Fortran test failed: libgomp: nvptx_alloc error: out of memory

### DIFF
--- a/tests/5.1/target/test_target_has_device_addr.F90
+++ b/tests/5.1/target/test_target_has_device_addr.F90
@@ -1,0 +1,67 @@
+!===--- test_target_has_device_addr.F90 ------------------------------------===//
+!
+! OpenMP API Version 5.1 Nov 2020
+!
+! This test verifies the 'has_device_addr' feature added to the target construct.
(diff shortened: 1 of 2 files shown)


seyonglee:
This PR has similar error as  PR #711 : `libgomp: nvptx_alloc error: out of memory`, but it is unclear whether this error is caused by bugs in the test program or a compiler bug. The out-of-memory error occurs when an unassociated Fortran pointer is passed to the target region.

OpenMP V5.1 Section 2.21.7.1 map Clasue, page 351 says "If the reference count of the corresponding list item is one or the always map-type-modifier is present, and if the map-type is to or tofrom, then: ...
For each part of the list item that is not an attached pointer, the value of that part of the original list item is assigned to that part of the corresponding list item."

fel-cab (approved):
I ran both test with gcc 12.1.1 and both passed without errors

spophale (approved):

PR #723, opened 2023-04-26
Add new_test/5.1/test_target_memcpy_async_no_obj.F90

seyonglee:
- GCC 12.2.1:
            - C test passed.
            - Fortran test failed: Error: 'omp_target_memcpy_async' at (1) is not a variable
        - XL 16.1.1-10:
            - C test failed: undefined reference to `omp_target_memcpy_async'
            - Fortran test failed: line 50.19: 1516-036 (S) Entity omp_target_alloc has undefined type.
        - NVHPC 22.11:
            - C test failed: FAIL. exit code: 134
            - Fortran test failed: NVFORTRAN-S-0034-Syntax error at or near end of line 59
        - LLVM 17.0.0:
            - C test failed: Condition test_target_memcpy_async_no_obj() != 0 failed

### DIFF
--- a/tests/5.1/target/test_target_memcpy_async_no_obj.F90
+++ b/tests/5.1/target/test_target_memcpy_async_no_obj.F90
@@ -58,6 +58,7 @@ CONTAINS
     ! copy to target
     omp_target_memcpy_async(mem_dev_cpy, mem, csize, dst_offset, src_offset, t, h, depobj_count)
 
+    !$omp taskwait
     !$omp target is_device_ptr(mem_dev_cpy) device(t)
     !$omp teams distribute parallel do
     DO i=1, N
--- a/tests/5.1/target/test_target_memcpy_async_no_obj.F90
+++ b/tests/5.1/target/test_target_memcpy_async_no_obj.F90
@@ -56,7 +56,7 @@ CONTAINS
     END DO
 
     ! copy to target
-    omp_target_memcpy_async(mem_dev_cpy, mem, csize, dst_offset, src_offset, t, h, depobj_count)
+    errors = omp_target_memcpy_async(mem_dev_cpy, mem, csize, dst_offset, src_offset, t, h, depobj_count)
 
     !$omp taskwait
     !$omp target is_device_ptr(mem_dev_cpy) device(t)
@@ -69,7 +69,7 @@ CONTAINS
     !$omp end target
 
     ! copy to host memory
-    omp_target_memcpy_async(mem, mem_dev_cpy, csize, dst_offset, src_offset, h, t, depobj_count)
+    errors = omp_target_memcpy_async(mem, mem_dev_cpy, csize, dst_offset, src_offset, h, t, depobj_count)
 
     !$omp taskwait
     DO i=1, N
@@ -78,7 +78,7 @@ CONTAINS
 
     ! free resources
     DEALLOCATE(arr)
-    omp_target_free(mem_dev_cpy, t)
+    CALL omp_target_free(mem_dev_cpy, t)
 
     test_memcpy_async_no_obj = errors
   END FUNCTION test_memcpy_async_no_obj
--- a/tests/5.1/target/test_target_memcpy_async_no_obj.F90
+++ b/tests/5.1/target/test_target_memcpy_async_no_obj.F90
@@ -5,8 +5,8 @@
 !  Inspired from OpenMP 5.1 Examples Doc, 5.16.4 & 8.9
 !  This test utilizes the omp_target_memcpy_async construct to
 !  allocate memory on the device asynchronously. The construct
-!  uses 0 for 'depobj_count', so that the clause is not dependent
-!  and memory is therefore copied synchronously.
+!  uses 0 for 'depobj_count'; therefore, the generated target task
+!  is not a dependent task, but the memory copy is still asynchronous.
 !
 !//===---------------------------------------------------------------------===//
 
--- a/tests/5.1/target/test_target_memcpy_async_no_obj.F90
+++ b/tests/5.1/target/test_target_memcpy_async_no_obj.F90
@@ -49,7 +49,17 @@ CONTAINS
     csize = c_sizeof(arr(1)) * N
     mem_dev_cpy = omp_target_alloc(csize, t)
 
+    OMPVV_TEST_AND_SET_VERBOSE(errors, .NOT. c_associated(mem))
+    IF(.NOT. c_associated(mem)) THEN
+      test_memcpy_async_depobj = errors
+      RETURN
+    END IF
+
     OMPVV_TEST_AND_SET_VERBOSE(errors, .NOT. c_associated(mem_dev_cpy))
+    IF(.NOT. c_associated(mem_dev_cpy)) THEN
+      test_memcpy_async_depobj = errors
+      RETURN  
+    END IF
 
     DO i=1, N
       arr(i) = i
@@ -57,6 +67,14 @@ CONTAINS
 
     ! copy to target
     errors = omp_target_memcpy_async(mem_dev_cpy, mem, csize, dst_offset, src_offset, t, h, depobj_count)
+
+    IF(errors /= 0) THEN
+      OMPVV_ERROR("omp_target_memcpy_async returns not 0");
+      DEALLOCATE(arr)
+      CALL omp_target_free(mem_dev_cpy, t)
+      test_memcpy_async_depobj = errors
+      RETURN
+    END IF
 
     !$omp taskwait
     !$omp target is_device_ptr(mem_dev_cpy) device(t)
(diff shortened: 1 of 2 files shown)


seyonglee:
[Update on July 11, 2023]
- GCC 13.1.1:
            - Fortran test passed on the device.

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,86 @@
...
+    !$omp teams distribute parallel do
+    DO i=1, N
+      CALL c_f_pointer(mem_dev_cpy, fptr, [N])
+      fptr(i) = fptr(i) * 2 ! initialize data on device
+    END DO
+    !$omp end teams distribute parallel do
+    !$omp end target
+
+    ! copy to host memory
+    errors = omp_target_memcpy_async(mem, mem_dev_cpy, csize, dst_offset, src_offset, h, t, depobj_count)
```
fel-cab, line 72:
```suggestion
    errors = errors + omp_target_memcpy_async(mem, mem_dev_cpy, csize, dst_offset, src_offset, h, t, depobj_count)
```
In case the first omp_target_memcpy_async returns non-zero, and this second one returns zero

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,86 @@
...
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, .NOT. c_associated(mem_dev_cpy))
+
+    DO i=1, N
+      arr(i) = i
+    END DO
+
+    ! copy to target
+    errors = omp_target_memcpy_async(mem_dev_cpy, mem, csize, dst_offset, src_offset, t, h, depobj_count)
+
```
fel-cab, line 60: Shouldn't the function returns here if the error is non-zero, or if one of the  memory allocation functions return a null pointer. This to avoid segmentation faults.

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,86 @@
...
+    INTEGER (C_INT) :: h, t, depobj_count
+
+    errors = 0
+    h = omp_get_initial_device()
+    t = omp_get_default_device()
+    dst_offset = 0
+    src_offset = 0
+    depobj_count = 0
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() .LT. 1 .OR. t .LT. 0)
```
spophale, line 45: Give info msg and exit here if error > 0

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,86 @@
...
+
+    DO i=1, N
+      arr(i) = i
+    END DO
+
+    ! copy to target
+    errors = omp_target_memcpy_async(mem_dev_cpy, mem, csize, dst_offset, src_offset, t, h, depobj_count)
+
+    !$omp taskwait
+    !$omp target is_device_ptr(mem_dev_cpy) device(t)
```
spophale, line 62: Check value of error before proceeding. If it is non-zero it is an error and should exit with appropriate error msg.

tests/5.1/target/test_target_memcpy_async_no_obj.F90
spophale: Not required. The error should be 0.

spophale:
@seyonglee error checks will have to be added to C test case as well.

tests/5.1/target/test_target_memcpy_async_no_obj.F90
fel-cab: done

tests/5.1/target/test_target_memcpy_async_no_obj.F90
fel-cab: done

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,104 @@
...
+    OMPVV_TEST_AND_SET_VERBOSE(errors, omp_get_num_devices() .LT. 1 .OR. t .LT. 0)
+
+    ALLOCATE(arr(N))
+    mem = c_loc(arr(1))
+    csize = c_sizeof(arr(1)) * N
+    mem_dev_cpy = omp_target_alloc(csize, t)
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, .NOT. c_associated(mem))
+    IF(.NOT. c_associated(mem)) THEN
+      test_memcpy_async_depobj = errors
```
tob2, line 54:
```suggestion
      test_memcpy_async_no_obj = errors
```

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,104 @@
...
+      RETURN
+    END IF
+
+    OMPVV_TEST_AND_SET_VERBOSE(errors, .NOT. c_associated(mem_dev_cpy))
+    IF(.NOT. c_associated(mem_dev_cpy)) THEN
+      test_memcpy_async_depobj = errors
```
tob2, line 60:
```suggestion
      test_memcpy_async_no_obj = errors
```

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,104 @@
...
+    END DO
+
+    ! copy to target
+    errors = omp_target_memcpy_async(mem_dev_cpy, mem, csize, dst_offset, src_offset, t, h, depobj_count)
+
+    IF(errors /= 0) THEN
+      OMPVV_ERROR("omp_target_memcpy_async returns not 0");
+      DEALLOCATE(arr)
+      CALL omp_target_free(mem_dev_cpy, t)
+      test_memcpy_async_depobj = errors
```
tob2, line 75:
```suggestion
      test_memcpy_async_no_obj = errors
```

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,104 @@
...
+      RETURN
+    END IF
+
+    !$omp taskwait
+    !$omp target is_device_ptr(mem_dev_cpy) device(t)
+    !$omp teams distribute parallel do
+    DO i=1, N
+      CALL c_f_pointer(mem_dev_cpy, fptr, [N])
+      fptr(i) = fptr(i) * 2 ! initialize data on device
```
tob2, line 84: In principle we need to nullify the pointer here as it is otherwise undefined; as the pointer is implicitly `map(tofrom: fptr)` and OpenMP's Fortran rules state that it not only maps the bare pointer but also its pointer target.

_BTW: This is the reason for the out-of-memory with GCC as the compiler first tries to allocate on the device sufficient memory and if the array bounds have some random values, that can be large. In my test run, it was 72340172838076799 bytes which surely exceeds the memory available on any device_

BUT: We also have to reset the pointer at the end of the target region to the value it had before the target region (in this case: NULL), but that's invalid as inside TARGET with a closely nested TEAMS no other closely nested statement is permitted; therefore, a 'fptr => null() cannot be added.

 Thus, for simplicity, we just declare the ptr variable locally, which also avoids potential races:

Namely:
```suggestion
      BLOCK
      DOUBLE PRECISION, POINTER :: fptr(:)
      CALL c_f_pointer(mem_dev_cpy, fptr, [N])
      fptr(i) = fptr(i) * 2 ! initialize data on device
      END BLOCK
```

tests/5.1/target/test_target_memcpy_async_no_obj.F90
```diff
@@ -0,0 +1,104 @@
...
+
+  OMPVV_TEST_VERBOSE(test_memcpy_async_no_obj() .NE. 0)
+
+  OMPVV_REPORT_AND_RETURN()
+
+CONTAINS
+  INTEGER FUNCTION test_memcpy_async_no_obj()
+    INTEGER :: errors, i
+    DOUBLE PRECISION, TARGET, ALLOCATABLE :: arr(:)
+    DOUBLE PRECISION, POINTER :: fptr(:)
```
tob2, line 33: If the suggestion below is applied, this becomes unused.
```suggestion
```

tob2 (commented):
Comments to fix the Fortran testcase; works then with GCC.

tests/5.1/target/test_target_memcpy_async_no_obj.F90
fel-cab: @tob2 
I would like to run the tests with a up to date GCC and GFORTRAN with Nvidia and AMD GPUs, but I don't have access to any system that have them. Do you have a script to build it that you can share with me?

spophale:
@fel-cab to test with Cray fortran.

tests/5.1/target/test_target_memcpy_async_no_obj.F90
fel-cab: I don't have any way to test this change. No fortran compiler that I have access to can handle this. In AMD and Nvidia GPUs systems with compilers (Nvidia, Cray and AMD). The test fails at compilation in all the cases.

spophale:
@spophale will check with gcc on Summit.

spophale (approved):

fel-cab (approved):

PR #725, opened 2023-05-03
Add new_test/5.1/test_target_memcpy_rect_async_no_obj.F90

seyonglee:
Add new_test/5.1/test_target_memcpy_rect_async_no_obj.F90
Add missing task wait directives to the C version.

        - GCC 12.2.1:
            - C test passed.
            - Fortran test failed: Error: 'omp_target_memcpy_async' at (1) is not a variable
        - XL 16.1.1-10:
            - C test failed: undefined reference to `omp_target_memcpy_rect_async'
            - Fortran test failed: line 53.15: 1516-036 (S) Entity omp_target_alloc has undefined type.
        - NVHPC 22.11:
            - C test failed: FAIL. exit code: 134
            - Fortran test failed: NVFORTRAN-S-0034-Syntax error at or near end of line 65
        - LLVM 17.0.0:
            - C test passed.

### DIFF
--- a/tests/5.1/target/test_target_memcpy_rect_async_no_obj.F90
+++ b/tests/5.1/target/test_target_memcpy_rect_async_no_obj.F90
@@ -1,0 +1,93 @@
+!===--- test_target_memcpy_rect_async_no_obj.F90 -------------------------------===//
+!
+! OpenMP API Version 5.1 Nov 2020
+!
+! Inspired from OpenMP 5.1 Examples Doc, 5.16.4 & 8.9
+! This test utilizes the omp_target_memcpy_rect_async construct to
+! allocate 2D memory on the device asynchronously. The construct
+! uses '0' for 'depobj_count', so that the clause is not dependent
+! and memory is therefore copied synchronously.
+!
+!//===---------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 5
+#define M 10
+
+PROGRAM test_target_memcpy_rect_async_no_obj
+  USE iso_fortran_env
+  USE, INTRINSIC :: iso_c_binding
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+  OMPVV_TEST_OFFLOADING
+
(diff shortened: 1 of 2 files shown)


tests/5.1/target/test_target_memcpy_rect_async_no_obj.c
```diff
@@ -68,8 +71,9 @@ int test_target_memcpy_async_depobj() {
                                 h,          t,
                                 0,          NULL);
 
+    #pragma omp taskwait
```
spophale, line 74: Why do we need this taskwait ?

spophale (approved):
Fortran looks good!

tests/5.1/target/test_target_memcpy_rect_async_no_obj.c
andrewkallai: Per 5.1 spec pg. 425 starting from line 25: "The omp_target_memcpy_rect_async routine executes as if part of a target task that is generated by the call to the routine and for which execution may be deferred" which from my guess means that the arrays hostRect and devRect could perhaps be accessed before the   omp_target_memcpy_rect_async() call specifically allocates memory for them. So, the taskwait ensures that any tasks before the pragma are executed.

tests/5.1/target/test_target_memcpy_rect_async_no_obj.F90
```diff
@@ -0,0 +1,93 @@
...
+    TYPE (C_PTR) :: mem, devRect
+    INTEGER (C_SIZE_T) :: csize
+    INTEGER (C_SIZE_T) :: offsets(2), volume(2), dimensions(2)
+    INTEGER (C_INT) :: h, t, depobj_count, num_dims
+    DOUBLE PRECISION, TARGET :: hostRect(M,N)
+
+    errors = 0
+    h = omp_get_initial_device()
+    t = omp_get_default_device()
+    volume = (/ 10, 5 /)
```
fel-cab, line 43:
```suggestion
    volume = (/ M, N /)
```

fel-cab (approved):

PR #728, opened 2023-05-03
New test for omp_set_num_teams()

jrreap:
Adds a test for the new `omp_set_num_teams()` routine which sets the default number of teams assigned to all subsequent teams regions. Also tests that an explicitly defined `num_teams()` clause overrides this set default as defined in the spec.

Compiles and runs as expected on Perlmutter with llvm 17

### DIFF
--- a/tests/5.1/teams/test_target_set_num_teams.c
+++ b/tests/5.1/teams/test_target_set_num_teams.c
@@ -29,7 +29,7 @@ int main() {
 	
 	#pragma omp target teams map(tofrom: num_teams)
 	{
-		if (omp_get_team_num() == 0 && omp_get_thread_num() == 0) {
+		if (omp_get_team_num() == 0 ) {
 			num_teams = omp_get_num_teams();
 		}
 	}               
@@ -39,12 +39,13 @@ int main() {
 
 	#pragma omp target teams map(tofrom: num_teams) num_teams(OMPVV_NUM_TEAMS_DEVICE)
 	{
-		if (omp_get_team_num() == 0 && omp_get_thread_num() == 0) {
+		if (omp_get_team_num() == 0 ) {
 			num_teams = omp_get_num_teams();
 		}
 	}
 
 	OMPVV_ERROR_IF(num_teams != OMPVV_NUM_TEAMS_DEVICE, "Number of teams was not properly overriden by the num_teams clause");
+	
 	OMPVV_TEST_AND_SET(errors, num_teams != OMPVV_NUM_TEAMS_DEVICE);
 
 	OMPVV_REPORT_AND_RETURN(errors);
--- a/tests/5.1/teams/test_target_set_num_teams.c
+++ b/tests/5.1/teams/test_target_set_num_teams.c
@@ -15,8 +15,6 @@
 #include <stdlib.h>
 #include "ompvv.h"
 
-#define N 1024
-
 int main() {
 
 	OMPVV_TEST_OFFLOADING;
@@ -24,24 +22,24 @@ int main() {
 	int errors = 0; 
 	int num_teams = 0;
 
-#pragma omp target
-{
-	omp_set_num_teams(1);
-}
+        #pragma omp target
+        {
+	  omp_set_num_teams(4);
+        }
 	
 	#pragma omp target teams map(tofrom: num_teams)
 	{
-		if (omp_get_team_num() == 0) {
+		if (omp_get_team_num() == 0 && omp_get_thread_num == 0) {
 			num_teams = omp_get_num_teams();
 		}
 	}               
 	
-	OMPVV_ERROR_IF(num_teams != 1, "Number of teams detected was not the amount set by omp_set_num_teams()");
-	OMPVV_TEST_AND_SET(errors, num_teams != 1);
+	OMPVV_ERROR_IF(num_teams != 4, "Number of teams detected was not the number set by omp_set_num_teams()");
+	OMPVV_TEST_AND_SET(errors, num_teams != 4);
 
 	#pragma omp target teams map(tofrom: num_teams) num_teams(OMPVV_NUM_TEAMS_DEVICE)
 	{
-		if (omp_get_team_num() == 0) {
+		if (omp_get_team_num() == 0 && omp_get_thread_num == 0) {
 			num_teams = omp_get_num_teams();
 		}
 	}
--- a/tests/5.1/teams/test_target_set_num_teams.c
+++ b/tests/5.1/teams/test_target_set_num_teams.c
@@ -30,7 +30,7 @@ int main() {
 	}               
 	
 	OMPVV_ERROR_IF(num_teams <= 4, "Number of teams detected was not the number set by omp_set_num_teams()");
-	OMPVV_TEST_AND_SET(errors, num_teams != 4);
+	OMPVV_TEST_AND_SET(errors, num_teams <= 4);
 
 	#pragma omp teams num_teams(OMPVV_NUM_TEAMS_HOST)
 	{
--- a/tests/5.1/teams/test_target_set_num_teams.c
+++ b/tests/5.1/teams/test_target_set_num_teams.c
@@ -39,7 +39,7 @@ int main() {
 		}
 	}
 
-	OMPVV_ERROR_IF(num_teams != OMPVV_NUM_TEAMS_HOST, "Number of teams was not properly overriden by the num_teams clause");	
+	OMPVV_ERROR_IF(num_teams <= OMPVV_NUM_TEAMS_HOST, "Number of teams was not properly overriden by the num_teams clause");	
 	OMPVV_TEST_AND_SET(errors, num_teams != OMPVV_NUM_TEAMS_HOST);
 	OMPVV_REPORT_AND_RETURN(errors);
 }
--- a/tests/5.1/teams/test_target_set_num_teams.c
+++ b/tests/5.1/teams/test_target_set_num_teams.c
@@ -25,21 +25,21 @@ int main() {
 	#pragma omp teams 
 	{
 		if (omp_get_team_num() == 0 ) {
-			num_teams = omp_get_num_teams();
+			num_teams = omp_get_max_teams();
 		}
 	}               
 	
-	OMPVV_ERROR_IF(num_teams <= 4, "Number of teams detected was not the number set by omp_set_num_teams()");
-	OMPVV_TEST_AND_SET(errors, num_teams <= 4);
+	OMPVV_ERROR_IF(num_teams != 4, "Upper bound of the number of teams is not the number set by omp_set_num_teams()");
+	OMPVV_TEST_AND_SET(errors, num_teams != 4);
 
 	#pragma omp teams num_teams(OMPVV_NUM_TEAMS_HOST)
 	{
 		if (omp_get_team_num() == 0 ) {
-			num_teams = omp_get_num_teams();
+			num_teams = omp_get_max_teams();
 		}
 	}
 
-	OMPVV_ERROR_IF(num_teams <= OMPVV_NUM_TEAMS_HOST, "Number of teams was not properly overriden by the num_teams clause");	
-	OMPVV_TEST_AND_SET(errors, num_teams <= OMPVV_NUM_TEAMS_HOST);
+	OMPVV_ERROR_IF(num_teams != OMPVV_NUM_TEAMS_HOST, "Number of teams was not properly overriden by the num_teams clause");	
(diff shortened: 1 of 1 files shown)


tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,50 @@
...
+#define N 1024
+
+int main() {
+
+	OMPVV_TEST_OFFLOADING;
+
+	int errors = 0; 
+	int num_teams = 0;
+
+	omp_set_num_teams(1);
```
spophale, line 27: Query max and set it to 1 less than the max value and check from within the target region.

tob2: I might miss something here, but doesn't the initial _nteams-var_ ICV, which exists per device, have the value zero, and 'omp_set_num_teams' sets the value for the current device (= host)?

Thus, I expect that inside 'omp target teams', the _nteams-var_ ICV is _zero_, unless the default device is the initial device. And if it is zero, regarding the number of teams for a **teams** construct, "the effect is as if _upper-bound_ was specified as an implementation defined value greater than or equal to one."

Or have I missed something? If not:
* One way out would be to test it only on the host (i.e. 'omp teams' not 'omp target teams').
* When using it with 'target', it cannot be in the _same_ target region as 'teams' due to the restrictions on 'teams' inside 'target'.
  However, I wonder whether it would work to set it in one 'omp target' and then use it in another 'omp target teams', relying on the preservation of the ICVs between subsequent target regions.

_(Note: Only glanced at TR11, but I don't expect relevant differences between 5.1, 5.2, and TR11 in this area, except that the handling of env vars such as OMP_NUM_TEAMS was clarified wrt non-initial devices and OMP_NUM_TEAMS_ALL has been added.)_

tob2: _First, the inconsistency with 'omp_get_max_teams' is now tracked in OpenMP Specification Issue 3619; the sentiment is to change the _default value_ to an implementation defined positive value (currently it is 0), which avoids the issue about what value 'omp_get_max_teams' will report. — But that's as an aside._

Regarding `omp_set_num_teams(1);`, as mentioned, this only sets the ICV of the host. You might want to do:
```suggestion
#pragma omp target
{
	omp_set_num_teams(1);
}
```

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,50 @@
...
+	
+	#pragma omp target teams map(tofrom: num_teams)
+	{
+		if (omp_get_team_num() == 0) {
+			num_teams = omp_get_num_teams();
+		}
+	}               
+	
+	OMPVV_ERROR_IF(num_teams == 1, "Number of teams detected was not the amount set by omp_set_num_teams()");
+	OMPVV_TEST_AND_SET(errors, num_teams == 1);
```
tob2, line 37: I don't quite understand this. Assuming you want to set the ICV via `omp_set_num_teams(1);`, shouldn't the expected value not be 1? But then I had expected the following.

_(Actually I had expect this `num_teams > 0 && num_teams <= nteams_var_value` — but as `omp_set_num_teams` was called with value one, this boils down to the following check.)_
```suggestion
	OMPVV_ERROR_IF(num_teams != 1, "Number of teams detected was not the amount set by omp_set_num_teams()");
	OMPVV_TEST_AND_SET(errors, num_teams != 1);
```

spophale (approved):

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,51 @@
...
+
+	OMPVV_TEST_OFFLOADING;
+
+	int errors = 0; 
+	int num_teams = 0;
+
+        #pragma omp target
+        {
+	  omp_set_num_teams(4);
+        }
```
fel-cab, line 28: With clang on Frontier
```
ld.lld: error: undefined symbol: omp_set_num_teams
>>> referenced by /tmp/test_target_set_num_teams.c.o.amdgcn-amd-amdhsa.gfx90a-0321f1.o:(__omp_offloading_7c_1476ac9d_main_l25)
>>> referenced by /tmp/test_target_set_num_teams.c.o.amdgcn-amd-amdhsa.gfx90a-0321f1.o:(__omp_offloading_7c_1476ac9d_main_l25)
clang: error: ld.lld command failed with exit code 1 (use -v to see invocation)

```

From the specification 5.1, page 399:
3.4.3 omp_set_num_teams
Restrictions
Restrictions to the omp_set_num_teams routine are as follows:
• The routine may not be called from within a parallel region that is not the implicit parallel region
that surrounds the whole OpenMP program.

spophale: It is not being called from a parallel region. 
It is possible that the clang llvm does not support it yet.

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,51 @@
...
+	
+	#pragma omp target teams map(tofrom: num_teams)
+	{
+		if (omp_get_team_num() == 0 && omp_get_thread_num() == 0) {
+			num_teams = omp_get_num_teams();
```
tob2, line 33: (**GCC** supports **omp_set_num_teams** since GCC 12, but:)

This **fails with GCC** (tried: mainline/GCC 14) with:
```C
test_target_set_num_teams.c: In function ‘main’:
test_target_set_num_teams.c:32:48: error: OpenMP runtime API call ‘omp_get_thread_num’ strictly nested in a ‘teams’ region
   32 |                 if (omp_get_team_num() == 0 && omp_get_thread_num() == 0) {
      |                                                ^~~~~~~~~~~~~~~~~~~~
test_target_set_num_teams.c:42:48: error: OpenMP runtime API call ‘omp_get_thread_num’ strictly nested in a ‘teams’ region
   42 |                 if (omp_get_team_num() == 0 && omp_get_thread_num() == 0) {
      |                                                ^~~~~~~~~~~~~~~~~~~~
```

spophale: @tob2 , thanks for testing it. This is an easy fix.

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,52 @@
...
+	{
+		if (omp_get_team_num() == 0 ) {
+			num_teams = omp_get_num_teams();
+		}
+	}               
+	
+	OMPVV_ERROR_IF(num_teams != 4, "Number of teams detected was not the number set by omp_set_num_teams()");
+	OMPVV_TEST_AND_SET(errors, num_teams != 4);
+
+	#pragma omp target teams map(tofrom: num_teams) num_teams(OMPVV_NUM_TEAMS_DEVICE)
```
andrewkallai, line 40: This test is not concerned with testing the num_teams clause as the num_teams clause does not set the nteams ICV which omp_set_num_teams() relates to. As such, lines 40-49 can possibly be removed.

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,52 @@
...
+
+        #pragma omp target
+        {
+	  omp_set_num_teams(4);
+        }
+	
+	#pragma omp target teams map(tofrom: num_teams)
+	{
+		if (omp_get_team_num() == 0 ) {
+			num_teams = omp_get_num_teams();
```
andrewkallai, line 33: omp_get_max_teams() should be used instead to query the nteams ICV which is what omp_set_num_teams() is used to change.

andrewkallai (changes requested):

spophale:
Following a discussion at the OpenMP languages committee, it seems that the set_num_teams is only meant for the host side at this time. We will hold off on testing it on target till the language is cleaned up.

tests/5.1/teams/test_target_set_num_teams.c
fel-cab: Given that the specification says that: 
"The omp_set_num_teams routine affects the number of threads to be used for subsequent
4 teams regions that do not specify a num_teams clause ..."
Either way could be used, I think.

fel-cab (approved):

tests/5.1/teams/test_target_set_num_teams.c
andrewkallai: This test does not use the method to retrieve the nteams ICV, omp_get_max_teams(). So the actual value of the variable is not being queried. The number of teams created in the teams region (with no clause) is being queried, however it is being expected that the number queried is equal to the number set, when it is legal to have the value be less than or equal to the number set.

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,45 @@
...
+	omp_set_num_teams(4);
+        	
+	#pragma omp teams 
+	{
+		if (omp_get_team_num() == 0 ) {
+			num_teams = omp_get_num_teams();
+		}
+	}               
+	
+	OMPVV_ERROR_IF(num_teams != 4, "Number of teams detected was not the number set by omp_set_num_teams()");
```
andrewkallai, line 32:
```suggestion
	OMPVV_ERROR_IF(num_teams <= 4, "Number of teams detected was not the number set by omp_set_num_teams()");
```
pg. 101 5.1 spec
21 If the num_teams clause is not specified and the value of the nteams-var ICV is greater than zero,
22 the number of teams created is less or equal to the value of the nteams-var ICV. Otherwise, the
23 number of teams created is implementation defined, but it will be greater than or equal to 1.
The nteams-var is greater than zero, as such it is allowed for the number of teams created to be less than or equal to the ICV value.

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,45 @@
...
+	OMPVV_TEST_AND_SET(errors, num_teams != 4);
+
+	#pragma omp teams num_teams(OMPVV_NUM_TEAMS_HOST)
+	{
+		if (omp_get_team_num() == 0 ) {
+			num_teams = omp_get_num_teams();
+		}
+	}
+
+	OMPVV_ERROR_IF(num_teams != OMPVV_NUM_TEAMS_HOST, "Number of teams was not properly overriden by the num_teams clause");
```
andrewkallai, line 42:
```suggestion
	OMPVV_ERROR_IF(num_teams <= OMPVV_NUM_TEAMS_HOST, "Number of teams was not properly overriden by the num_teams clause");	
```
pg. 101 5.1 spec
16 If the num_teams clause is present, lower-bound is the specified lower bound and upper-bound is
17 the specified upper bound on the number of teams requested. If a lower bound is not specified, the
18 lower bound is set to the specified upper bound. The number of teams created is implementation
19 defined, but it will be greater than or equal to the lower bound and less than or equal to the upper
20 bound.
Because only an upperbound is present, the number of teams created is allowable to be less than or equal to the upperbound.

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,45 @@
...
+
+	#pragma omp teams num_teams(OMPVV_NUM_TEAMS_HOST)
+	{
+		if (omp_get_team_num() == 0 ) {
+			num_teams = omp_get_num_teams();
+		}
+	}
+
+	OMPVV_ERROR_IF(num_teams != OMPVV_NUM_TEAMS_HOST, "Number of teams was not properly overriden by the num_teams clause");	
+	OMPVV_TEST_AND_SET(errors, num_teams != OMPVV_NUM_TEAMS_HOST);
```
andrewkallai, line 43:
```suggestion
	OMPVV_TEST_AND_SET(errors, num_teams <= OMPVV_NUM_TEAMS_HOST);
```

tests/5.1/teams/test_target_set_num_teams.c
```diff
@@ -0,0 +1,45 @@
...
+        	
+	#pragma omp teams 
+	{
+		if (omp_get_team_num() == 0 ) {
+			num_teams = omp_get_num_teams();
+		}
+	}               
+	
+	OMPVV_ERROR_IF(num_teams != 4, "Number of teams detected was not the number set by omp_set_num_teams()");
+	OMPVV_TEST_AND_SET(errors, num_teams != 4);
```
andrewkallai, line 33:
```suggestion
	OMPVV_TEST_AND_SET(errors, num_teams <= 4);
```

andrewkallai (changes requested):

andrewkallai (approved):

Issue #733, opened 2023-05-12
Undefined behavior in tests/5.0/requires/test_requires_unified_address.c


tob2:
This testcase uses, **requires unified_address**, i.e. which just means the following (full quote below, important bits quoted here; bold is by me):

> Non-host devices may still have discrete memories and **dereferencing** [...] **a host pointer on a non-host device remains unspecified behavior.**

But then it does the following, which clearly assumes that the memory is shared — the mapping actually does a "map(to:" instead of "firstprivate" of the pointer but it still points to the memory allocated on the host, which only works with (unified) shared memory:

https://github.com/SOLLVE/sollve_vv/blob/e8e1e9fcee03b491a0c7143e36e006edda50e9c2/tests/5.0/requires/test_requires_unified_address.c#L27-L30

Additionally, the code expects that the assignment works, which is checked at:

https://github.com/SOLLVE/sollve_vv/blob/e8e1e9fcee03b491a0c7143e36e006edda50e9c2/tests/5.0/requires/test_requires_unified_address.c#L34-L36

### Hence,
* Either change **requires unified_address** to **requires unified_shared_address**
* or change the code such that it only uses code permitted according to the rules above, e.g.
  * not using **is_device_ptr**
  * Doing pointer offsets on the host for device pointer
  * Use **omp_target_alloc** to get the memory; or obtain the device address (or a mapped pointer) via **use_device_ptr** or inside a target region - and use that address without is_device_ptr and possibly after adding an offet.
* * *
OpenMP 5.0 wording for **requires unified_address** is:
> the implementation guarantees that all devices accessible through OpenMP API routines and directives use a unified address space. In this address space, a pointer will always refer to the same location in memory from all devices accessible through OpenMP. The pointers returned by omp_target_alloc and accessed through use_device_ptr are guaranteed to be pointer values that can support pointer arithmetic while still being native device pointers. The is_device_ptr clause is not necessary for device pointers to be translated in target regions, and pointers found not present are not set to null but keep their original value. [...] **Target devices may still have discrete memories and dereferencing a device pointer on the host device or host pointer on a target device remains unspecified behavior.**

TR11 wording:
> requires an implementation to guarantee that all devices accessible through OpenMP API routines and directives use a unified
[...]

PR #734, opened 2023-05-12
Issue 733: Fix test_requires_unified_address.c

tob2:
Issue #733 — _The problem is that the code assumes unified shared memory but only required unified-shared address_

Before this commit, the code assumed that the memory between host and device is shared, which is not guaranteed. Change the code to only check what the OpenMP spec guarantees for unified_address.

Namely, that is_device_ptr is not needed and that that host-side pointer arithmetic is permitted for device pointers.

Notes:
* This test fixes the fail for GCC mainline (=being developed GCC 14), but only with **is_device_pointer** clause; without it segfaults.
* The issue quotes from the 5.0 spec and for TR11, which clarified/changed some minor bits, which should not be relevant here.

@nolanbaker31 @spophale @jrreap @krishols @mjcarr458 @seyonglee – please review.

### DIFF
--- a/tests/5.0/requires/test_requires_unified_address.c
+++ b/tests/5.0/requires/test_requires_unified_address.c
@@ -20,23 +20,42 @@ int unified_address() {
 
    int errors = 0;
    int i;
-   int * mem_ptr = (int *)malloc(N * sizeof(int));
+   int *mem_ptr = (int *)omp_target_alloc(N * sizeof(int), omp_get_default_device());
+   int *mem_ptr2 = NULL;
 
    OMPVV_ERROR_IF(mem_ptr == NULL, "Memory was not properly allocated");
    
-   #pragma omp target map(to: mem_ptr)
+   #pragma omp target map(from:mem_ptr2) /* is_device_ptr(mem_ptr) - which is optional.  */
    {
       for (i = 0; i < N; i++) {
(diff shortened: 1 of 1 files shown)


tests/5.0/requires/test_requires_unified_address.c
```diff
@@ -20,23 +20,42 @@ int unified_address() {
 
    int errors = 0;
    int i;
-   int * mem_ptr = (int *)malloc(N * sizeof(int));
+   int *mem_ptr = (int *)omp_target_alloc(N * sizeof(int), omp_get_default_device());
+   int *mem_ptr2 = NULL;
 
    OMPVV_ERROR_IF(mem_ptr == NULL, "Memory was not properly allocated");
```
fel-cab, line 26: I suggest to return from the function after `OMPVV_ERROR_IF`: `if(mem_ptr == NULL) return 1;`

fel-cab (approved):

tob2:
@spophale — ping?!

tob2:
Already approved by @fel-cab (Thanks!)

Still needs another review — @spophale ?
Or maybe: @nolanbaker31  @jrreap @krishols @mjcarr458 @seyonglee – please review.

spophale (approved):

spophale:
Thanks @tob2 !

PR #735, opened 2023-05-26
Added initial file for testing mapping of pointer with not matching data item

andrewkallai:
On Perlmutter:
PrgEnv-gnu (cc): gcc (GCC) 12.2.0 20220819 (HPE): Fails to compile
35:8: error: 'p' not specified in enclosing 'target'
   35 |       p[index] = i;
30:11: note: enclosing 'target'
   30 |   #pragma omp target defaultmap(none:pointer)
PrgEnv-nvidia (cc): nvc 22.7: Fails to offload, but compiles with warnings:
/usr/bin/ld: warning: /tmp/pgcudafatJXckP7rI5YfH.o: missing .note.GNU-stack section implies executable stack
/usr/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
PrgEnv-nvhpc (cc):  nvc 22.7: Fails to offload, but compiles with warnings:
/usr/bin/ld: warning: /tmp/pgcudafatb74qdfPwlpRu.o: missing .note.GNU-stack section implies executable stack
/usr/bin/ld: NOTE: This behaviour is deprecated and will be removed in a future version of the linker
Runtime output for each:
[OMPVV_ERROR: test_declare_target_map_device_pointer.c:44] ERROR COUNT (DIFFERING VALUES): 1000
[OMPVV_ERROR: test_declare_target_map_device_pointer.c:51]  Condition test_pointer() != 0 failed
[OMPVV_INFO: test_declare_target_map_device_pointer.c:50] Test is running on host.
[OMPVV_INFO: test_declare_target_map_device_pointer.c:52] The value of errors is 1.
[OMPVV_RESULT: test_declare_target_map_device_pointer.c] Test failed on the host.
PrgEnv-cray (cc): Cray clang version 15.0.1: Fails to compile
error: variable 'p' must have explicitly specified data sharing attributes, data mapping attributes, or in an is_device_ptr clause
      p[index] = i;
note: explicit data sharing attribute, data mapping attribute, or is_device_ptr clause requested here
  #pragma omp target defaultmap(none:pointer)
1 error generated.


tob2:
Admittedly, I do not complete understand what the current version tries to do. However, I do agree with the GCC and the HPE/Cray compiler — namely, that due to 'defaultmap(none:pointer)' using 'p' inside of the first target region is invalid as it does not appear in any data-sharing clause.

* * *

I understand the purpose of this pull request/testcase correctly, the idea is to test **OpenMP 5.2**'s
"If a matching mapped list item is not found, the pointer retains its original value as per the firstprivate semantics described in Section 5.4.4." (last paragraph in 5.2's "5.8.6 Pointer Initialization for Device Data Environments")

In **OpenMP 5.1**, the same section (there: 2.21.7.2) had:
"If a matching mapped list item is not found, the assigned initial value of the pointer is _NULL_ unless otherwise specified (see Section 2.5.1)."

* * *
How about something like the following? With OpenMP 5.1 (or e.g. current GCC with non-shared-memory offloading), p1 and p2 are NULL in the target region, while with OpenMP 5.2 the are supposed to keep their value (as tested below).

```C
#include <stdlib.h>
#include <stdint.h>
#include <omp.h>

int main () {
  int A[3] = {1,2,3};
  int *p1, *p2, *p3;
  intptr_t ip;
  p1 = (int*) 0x12345;
  p2 = (int*) omp_target_alloc (sizeof(int) * 5, omp_get_default_device());
  p3 = &A[0];
  ip = (intptr_t) p2;
  #pragma omp target enter data map(to:A)
  #pragma omp target
  {
    /* The pointees aren't mapped. */
    /* OpenMP 5.1 -> NULL */
    /* OpenMP 5.2 -> same value as before the target region. */
    if ((intptr_t) p1 != 0x12345) abort ();
    if ((intptr_t) p2 != ip) abort ();
    /* Mapped => must point to the corresponding device storage of 'A' */
    if (p3[0] != 1 || p3[1] != 2 || p3[2] != 3) abort ();
  }
  omp_target_free (p2, omp_get_default_device());
  return 0;
}
```

PR #743, opened 2023-06-25
Added initial file for testing mapping of pointer with not matching data item

andrewkallai:
frontier
clang 18.0.0
[OMPVV_WARNING: test_target_map_device_pointer.c:58] _ompvv_isOffloadingOn: 1, _ompvv_isSharedEnv: 1, test is invalid
summit
gcc 13.2.1
[OMPVV_WARNING: test_target_map_device_pointer.c:57] _ompvv_isOffloadingOn: 1, _ompvv_isSharedEnv: 1, test is invalid
nvc 22.11
line 67: warning: statement is unreachable [code_is_unreachable]
    return errors;
[OMPVV_WARNING: test_target_map_device_pointer.c:58] _ompvv_isOffloadingOn: 1, _ompvv_isSharedEnv: 1, test is invalid

### DIFF
--- a/tests/5.2/target/test_target_map_device_pointer.c
+++ b/tests/5.2/target/test_target_map_device_pointer.c
@@ -1,53 +1,63 @@
-//===--- test_declare_target_map_device_pointer.c ------------------------------------------===//
-// 
-// OpenMP API Version 5.2 Nov 2021 
-//
-//This test checks if, after offloading, the initalized pointer variable retains its 
-//original value as per the semantics of the firstprivate clause. 
-//The pointer variable, nor the array it refers to, are not mapped to the device. This 
-//ensures that a matching mapped data item is not found for the pointer, so it retains
-//its original value.
-//===---------------------------------------------------------------------------------------===//
-
-#include <stdio.h>
-#include <stdlib.h>
-#include <omp.h>
+//--------------- test_target_map_device_pointer.c ---------------------------//
+// OpenMP API Version 5.2 Nov 2021
+// Pg. 627, line 24
+// This test ensures that--paraphrase of Section 5.8.6--a pointer that is
+// implicitly mapped or explicitely mapped to the device, predetermined to be
+// firstprivate in the device data environment, will retain its original value
+// (attained prior to the target region) and follow the semantics of the
+// firstprivate clause. This is true if a matching mapped list item, as defined
+// in Section 5.8.6, is not found for the pointer. In the test, the pointer p1
+// is initialized with the base array address of arr2. When p1 is implicitly
+// mapped to the device, it should retain the same base array address, as the
+// array itself is not mapped (a matching mapped list item is not found). This
+// check is made in the target region with a mapped integer variable holding the
+// base address of arr2. The pointer p2 is initialized to NULL, and is checked
+// to remain NULL after being initialized on a device. Following the checks, p1
+// is changed to NULL on the device and checked outside of the target region if
+// it was not changed from its initial value of the base address of arr2 (per
+// firstprivate semnatics). p2 is given an arbitrary value for the purpose of
+// ensuring that it too retains its original value after the target region. A
+// defaultmap clause is present on the target construct to explicitely ensure
+// that the default semantics of mapping apply to all mapped (implicit/explicit)
+// pointer types. The zero length array, arr1, is mapped to the device to ensure
+// that a runtime check occurs in accordance with the semantics described in the
+// target construct on page 285 of the specifications.
+//----------------------------------------------------------------------------//
 #include "ompvv.h"
+#include <omp.h>
+#include <stdint.h> //includes intptr_t
 
-#define N 1000
+#define N 16
 
-int test_pointer() {
-  int compute_array[N];
-  int *p;
+int test_target_pointer() {
   int errors = 0;
-  int i;
-  
-  // Array initialization
-  for (i = 0; i < N; i++)
-    compute_array[i] = i;
-  p = &compute_array[0];
+  int arr1[0], arr2[N];
+  int *p1 = &arr2[0];
+  int *p2 = NULL;
+  intptr_t hold_arr2 = (intptr_t)&arr2[0];
 
-  #pragma omp target defaultmap(none:pointer)
+  #pragma omp target map(tofrom:errors) map(to:hold_arr2,arr1[:0]) defaultmap(default:pointer)
   {
-    // Array modified through the pointer
-    int index=0;
-    for (i = N-1; i >= 0; i--){
-      p[index] = i;
-      index++;
-    }
-  } // end target
+    OMPVV_TEST_AND_SET(errors, ((intptr_t) p1 != hold_arr2) || (p2 != NULL));
+    
+    p1 = NULL;
+    p2 = (int *) 0x12345;
+  }
 
-  // Pointer values comparison
-  for (i = 0; i < N; i++)
-    if (p[i] != i) errors++;
-
-  OMPVV_ERROR_IF(errors!=0, "ERROR COUNT (DIFFERING VALUES): %d",errors);
-  
+  OMPVV_ERROR_IF(errors != 0,
+                 "At least one pointer did not retain its original value");
+  OMPVV_ERROR_IF(p1 != &arr2[0],
+                 "The value of p1 was updated, p1 was not firstprivate");
+  OMPVV_ERROR_IF(p2 != NULL,
+                 "The value of p2 was updated, p2 was not firstprivate");
+  OMPVV_TEST_AND_SET(errors, (p1 != &arr2[0]) || (p2 != NULL))
   return errors;
 }
-int main(){
-  int errors=0;
+
+int main() {
+  int errors = 0;
   OMPVV_TEST_OFFLOADING;
-  OMPVV_TEST_AND_SET_VERBOSE(errors, test_pointer() != 0);
+  OMPVV_TEST_AND_SET(errors, test_target_pointer() != 0)
   OMPVV_REPORT_AND_RETURN(errors);
+  return errors;
 }
--- a/tests/5.2/target/test_target_map_device_pointer.c
+++ b/tests/5.2/target/test_target_map_device_pointer.c
@@ -56,8 +56,9 @@ int test_target_pointer() {
 
 int main() {
   int errors = 0;
-  OMPVV_TEST_OFFLOADING;
-  OMPVV_TEST_AND_SET(errors, test_target_pointer() != 0)
+  int on_device = 0;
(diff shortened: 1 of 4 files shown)


andrewkallai:
Per @tob2
Admittedly, I do not complete understand what the current version tries to do. However, I do agree with the GCC and the HPE/Cray compiler — namely, that due to 'defaultmap(none:pointer)' using 'p' inside of the first target region is invalid as it does not appear in any data-sharing clause.

* * *

I understand the purpose of this pull request/testcase correctly, the idea is to test **OpenMP 5.2**'s
"If a matching mapped list item is not found, the pointer retains its original value as per the firstprivate semantics described in Section 5.4.4." (last paragraph in 5.2's "5.8.6 Pointer Initialization for Device Data Environments")

In **OpenMP 5.1**, the same section (there: 2.21.7.2) had:
"If a matching mapped list item is not found, the assigned initial value of the pointer is _NULL_ unless otherwise specified (see Section 2.5.1)."

* * *
How about something like the following? With OpenMP 5.1 (or e.g. current GCC with non-shared-memory offloading), p1 and p2 are NULL in the target region, while with OpenMP 5.2 the are supposed to keep their value (as tested below).

```C
#include <stdlib.h>
#include <stdint.h>
#include <omp.h>

int main () {
  int A[3] = {1,2,3};
  int *p1, *p2, *p3;
  intptr_t ip;
  p1 = (int*) 0x12345;
  p2 = (int*) omp_target_alloc (sizeof(int) * 5, omp_get_default_device());
  p3 = &A[0];
  ip = (intptr_t) p2;
  #pragma omp target enter data map(to:A)
  #pragma omp target
  {
    /* The pointees aren't mapped. */
    /* OpenMP 5.1 -> NULL */
    /* OpenMP 5.2 -> same value as before the target region. */
    if ((intptr_t) p1 != 0x12345) abort ();
    if ((intptr_t) p2 != ip) abort ();
    /* Mapped => must point to the corresponding device storage of 'A' */
    if (p3[0] != 1 || p3[1] != 2 || p3[2] != 3) abort ();
  }
  omp_target_free (p2, omp_get_default_device());
  return 0;
}
```

[Edited – namely adding:]

Notes to the example above:
* Implicit default mapping (=`defaultmap(default:pointer)`) is required. That is similar but not the same as `firstprivate` (= `defaultmap(firstprivate:pointer)`). With `firstprivate(p1,p2)` the code above works but `firstprivate(p3)` will then fail (unless shared memory). → https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.0/requires/test_requires_unified_address.c for a testcase of the latter. _(This is an OpenMP 5.0 testcase.)_
* The code above only works if `sizeof(*void)` is the same on host and device (maybe also if it is larger on the host); it also requires that `omp_target_alloc` returns a normal/bare pointer (should be the case if pointer size is the same).
   * That's extremely common and likely other things break if that assumption does not hold, but the OpenMP spec does not guarantee this unless `omp required unified_address` is used. But unified_address goes beyond as it requires that for any address it is clear that host and device addresses are separate, which is not required for the testcase above

tests/5.2/target/test_target_map_device_pointer.c
```diff
@@ -0,0 +1,53 @@
...
+  int *p;
+  int errors = 0;
+  int i;
+  
+  // Array initialization
+  for (i = 0; i < N; i++)
+    compute_array[i] = i;
+  p = &compute_array[0];
+
+  #pragma omp target defaultmap(none:pointer)
```
spophale, line 30: comment 1: remove defaultmap(none:pointer)
comment 2: how will this interact with implicit map of compute_array ? Maybe use a scalar?

tests/5.2/target/test_target_map_device_pointer.c
```diff
@@ -0,0 +1,53 @@
...
+int test_pointer() {
+  int compute_array[N];
+  int *p;
+  int errors = 0;
+  int i;
+  
+  // Array initialization
+  for (i = 0; i < N; i++)
+    compute_array[i] = i;
+  p = &compute_array[0];
```
spophale, line 28: p = omp_target_alloc(...)
This will give a valid address on device which can be queried and checked.

tests/5.2/target/test_target_map_device_pointer.c
```diff
@@ -0,0 +1,64 @@
...
+  {
+    OMPVV_TEST_AND_SET(errors, ((intptr_t) p1 != hold_arr2) || (p2 != NULL));
+    
+    p1 = NULL;
+    p2 = (int *) 0x12345;
+  }
+
+  OMPVV_ERROR_IF(errors != 0,
+                 "At least one pointer did not retain its original value");
+  OMPVV_ERROR_IF(p1 != &arr2[0],
```
spophale, line 49: Check if running on host as well as shared memory check.

tests/5.2/target/test_target_map_device_pointer.c
```diff
@@ -0,0 +1,64 @@
...
+//----------------------------------------------------------------------------//
+#include "ompvv.h"
+#include <omp.h>
+#include <stdint.h> //includes intptr_t
+
+#define N 16
+
+int test_target_pointer() {
+  int errors = 0;
+  int arr1[0], arr2[N];
```
fel-cab, line 34: Why `int arr1[0]`?

andrewkallai: omp examples 5.2: 6.3 Pointer Mapping
21 This scenario can be further split into two cases: the list item is a
22 zero-length array section (e.g., p[:0]) or it is not.
23 If it is a zero-length array section, this will trigger a runtime check on entry to the target region
24 for a previously mapped list item where the value of the pointer falls within the range of its base
25 address and ending address. If such a match is found the private pointer is initialized to the device
26 address corresponding to the value of the original pointer, and otherwise it is initialized to NULL
27 (or retains its original value if the unified_address requirement is specified for that
28 compilation unit).
29 If the list item (again, call it the pointee) is not a zero-length array section, the private pointer will
30 be initialized such that references in the target region to the pointee list item that use the pointer
31 will access the corresponding pointee.
Note^: Pointer is not initialized to NULL as stated (openmp forgot to update this part)

seyonglee: It seems that the zero-length array is a GCC extension but not a part of the C standard (https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html), which also says the following:

> Declaring zero-length arrays in other contexts, including as interior members of structure objects or as non-member objects, is discouraged. Accessing elements of zero-length arrays declared in such contexts is undefined and may be diagnosed.

spophale (approved):

spophale:
@andrewkallai to create a PR for changing behavior of "_ompvv_isSharedEnv" macro before this test is merged.

andrewkallai:
List of tests using OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT macro
./4.5/target_enter_data/test_target_enter_data_allocate_array_alloc.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_data/test_target_enter_data_components_to.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_data/test_target_enter_data_devices.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_data/test_target_enter_data_if.F90:         OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_data/test_target_enter_data_if.c:  OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedMemory)
./4.5/target_enter_data/test_target_enter_data_module_array.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_data/test_target_enter_data_set_default_device.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_exit_data/test_target_enter_exit_data_allocate_array_alloc_delete.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_exit_data/test_target_enter_exit_data_devices.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_exit_data/test_target_enter_exit_data_if.F90:         OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_exit_data/test_target_enter_exit_data_module_array.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_enter_exit_data/test_target_enter_exit_data_set_default_device.F90:        OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./4.5/target_teams_distribute/test_target_teams_distribute_defaultmap.F90:  OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(isSharedEnv)
./5.0/requires/test_requires_reverse_offload.c:    OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env);
./5.0/requires/test_requires_reverse_offload.F90:  OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env)
./5.0/target/test_target_device.c:    OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env);
./5.0/target/test_target_device.F90:    OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env)
./5.1/target/test_target_defaultmap_present_scalar.c:   OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env);
./5.1/target/test_target_defaultmap_present_scalar.F90:    OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(is_shared_env)

tests/5.2/target/test_target_map_device_pointer.c
```diff
@@ -0,0 +1,68 @@
...
+    OMPVV_TEST_AND_SET(errors, ((intptr_t) p1 != hold_arr2) || (p2 != NULL));
+    
+    p1 = NULL;
+    p2 = (int *) 0x12345;
+  }
+
+  OMPVV_TEST_OFFLOADING_PROBE;
+  OMPVV_TEST_SHARED_ENVIRONMENT_PROBE;
+
+  if (_ompvv_isOffloadingOn && !_ompvv_isSharedEnv){
```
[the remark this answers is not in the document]
spophale: PR #788 should have fixed it.

fel-cab (approved):

PR #746, opened 2023-06-30
Add new_test/5.1/test_target_is_accessible.F90

seyonglee:
- GCC 13.1.1:
            - Both C and Fortran tests passed on the device: Omp_target_is_accessible returning false
- XL 16.1.1-10: 
            - C test failed: undefined reference to `omp_target_is_accessible'
            - Fortran test failed: line 51.18: 1516-036 (S) Entity omp_target_is_accessible has undefined type.
- NVHPC 22.11:
            - C test failed: test_target_is_accessible.c.o: FAIL. exit code: 134
            - Fortran test failed: test_target_is_accessible.F90.o: FAIL. exit code: 134
 - LLVM 17.0.0:
            - C test failed: undefined reference to `omp_target_is_accessible'

### DIFF
--- a/tests/5.1/target/test_target_is_accessible.F90
+++ b/tests/5.1/target/test_target_is_accessible.F90
@@ -1,0 +1,72 @@
+!===--- test_target_is_accessible.F90 -------------------------------===//
+!
+! OpenMP API Version 5.1 Nov 2020
+!
+! This test checks that the omp_target_is_accessible device routine.
+! In this test the output of the target_is_accessible call should return
+! true because the storage indicated by the first and second arguements
+! is accessible by the targeted device. This test is closely adapdted
+! from the 5.1 OpenMP example sheet.
+!
+!//===---------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+PROGRAM test_target_is_accessible
+  USE iso_fortran_env
+  USE, INTRINSIC :: iso_c_binding
+  USE ompvv_lib
+  USE omp_lib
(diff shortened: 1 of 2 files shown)


tests/5.1/target/test_target_is_accessible.F90
```diff
@@ -0,0 +1,72 @@
...
+
+    errors = 0
+    isSharedMemory = 0
+    check_test = 2
+    N = 100
+    dev = omp_get_default_device()
+
+    !Assumes that on shared-memory systems, no copy is done
+    !$omp target map(to: isSharedMemory)
+    isSharedMemory = 1
```
spophale, line 43: What is the intent here ?

seyonglee: If the target system supports unified shared memory, the host will see this change, and thus isSharedMemory = check_test (line 51) = 1. 
If the target system does not support unified shared memory, the host will see isSharedMemory = 0, and check_test = 0 too.
In the host fallback case, the host will see isSharedMemory = 1, and check_test = 1 too.

spophale (approved):

tests/5.1/target/test_target_is_accessible.F90
fel-cab: I don't think `isSharedMemory` is necessary. Since the program is to test `omp_target_is_accessible` not shared memory.

fel-cab (approved):

PR #749, opened 2023-07-06
Fixes for tests/5.0/target_update/*_discontiguous.c tests

jtb20:
This patch fixes two problems with the OpenMP 5.0 discontiguous 'target update' tests: firstly, the length in an array section specifier does not mean that "base+length" describes the exclusive upper limit for the elements to be transferred when the stride is not one, but rather it is the *number of elements* to be transferred.  (This is different in Fortran, where the array syntax is (lower bound:upper bound:stride) instead, and no elements beyond the upper bound will be accessed.)

That means that several tests contain buffer overruns, as they use e.g. [0:N:2] for an array section specifier for an array of size N -- such an array section describes elements at indices 0, 2, 4, [...], 2*(N-1), which was clearly not the intention.

The patch also fixes an uninitialised 'error' count variable in test_target_update_mapper_to_discontiguous.c.

(I've been working on discontiguous OpenMP "target update" support in GCC, and came across these problems during testing.)

### DIFF
--- a/tests/5.0/target_update/test_target_update_from_discontiguous.c
+++ b/tests/5.0/target_update/test_target_update_from_discontiguous.c
@@ -32,7 +32,7 @@ int target_update_from_discontiguous() {
     }//end target
 
 
-    #pragma omp target update from(result[0:N:2])
+    #pragma omp target update from(result[0:N/2:2])
 
   }
 
--- a/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
+++ b/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
@@ -48,7 +48,7 @@ int target_update_from_mapper() {
       s.data[i] += i ;
     }
   
-  #pragma omp target update from(s.data[0:s.len:2]) //only update even array elements
+  #pragma omp target update from(s.data[0:s.len/2:2]) //only update even array elements
   } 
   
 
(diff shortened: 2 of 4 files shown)


tob2 (approved):
@spophale @nolanbaker31 @seyonglee @fel-cab — Please review.

@jtb20's commit does:
* Adds a missing initialization for a stack variable
* Fixes an out-of-bound issues with array strides in `omp update`, cf. below.

This LGTM.

* * *

Note that the following change is because strides are handles quite differently in Fortran (base language) and in C/C++ (OpenMP defined):

```diff
-  #pragma omp target update from(result[0:N:2])
+  #pragma omp target update from(result[0:N/2:2])
```

* Fortran has: _lower-bound_ and _upper-bound_ — and the last array-section element is the one which just fits into the _upper-bound_ after keeping adding the stride to the index.
* For C/C++, OpenMP has _lower-bound_ and _length_ — such that always _length_ elements are in the array section, either every element (without stride) or every _stride_-th element.
 
From the OpenMP 5.2 spec ("3.2.5 Array Sections" — this is a C/C++ only section):
> Each of the _lower-bound, length,_ and _stride_ expressions if specified must be an integral type expression of the base language. When evaluated they represent a set of integer values as follows:
`{ lower-bound, lower-bound + stride, lower-bound + 2 * stride,... , lower-bound + ((length - 1) * stride) }`

seyonglee (approved):
Thank you for catching this error!

fel-cab (approved):

PR #750, opened 2023-07-07
New test for omp_get_max_teams()

andrewkallai:
PLEASE REFER TO CLOSED PR729 FOR RELEVANT DISCUSSION COMMENTS
frontier
clang 18.0.0
ld.lld: error: undefined symbol: omp_get_max_teams
|>>> referenced by /tmp/test_target_get_max_teams.c.o.amdgcn-amd-amdhsa.gfx90a-b5ea00.o:(__omp_offloading_72_118bc122_test_get_max_teams_l32)
|>>> referenced by /tmp/test_target_get_max_teams.c.o.amdgcn-amd-amdhsa.gfx90a-b5ea00.o:(__omp_offloading_72_118bc122_test_get_max_teams_l32)
clang: error: ld.lld command failed with exit code 1 (use -v to see invocation)
/sw/frontier/ums/ums012/llvm/18.0.0-20230828/bin/clang-linker-wrapper: error: 'clang' failed
clang: error: linker command failed with exit code 1 (use -v to see invocation)
summit
gcc 13.2.1
PASS
nvc 22.11
line 61: warning: statement is unreachable [code_is_unreachable]
    return errors;
Remark: individual warnings can be suppressed with "--diag_suppress <warning-name>"
NVC++-S-1073-Procedures called in a OpenMP target region must have 'omp declare target' information - omp_get_max_teams (/autofs/nccs-svm1_home1/andrewka/sollve_vv/tests/5.1/teams/test_target_get_max_teams.c: 34)
NVC++/power Linux 22.11-0: compilation completed with severe errors

### DIFF
--- a/tests/5.1/teams/test_target_get_max_teams.c
+++ b/tests/5.1/teams/test_target_get_max_teams.c
@@ -1,56 +1,59 @@
-//===--------------------- test_target_get_max_teams.c ----------------------===//
-//
+//--------------- test_target_get_max_teams.c --------------------------------//
 // OpenMP API Version 5.1 Nov 2020
-//
-// This test uses the omp_get_max_teams routine to check what the max teams 
-// capacity is for this device. It should return the total amount of teams
-// that could be allocated to a specific teams region
-//
-//===------------------------------------------------------------------------===//
-
-#include <omp.h>
-#include <stdio.h>
-#include <stdlib.h>
+// ***************************
+// ROUTINE: omp_get_max_teams
+// ***************************
+// This test uses the omp_get_max_teams routine to check what the max teams
+// capacity is for the current device. If the routine returns 0, the nteams-var
+// ICV is 0, thus it is not treated as an upper bound and the behavior of the
+// program is implementation defined. If the routine were to return a negative
+// integer, the same behavior applies. If the routine returns a positive
+// integer, the value of the ICV is treated as an upperbound. If the value is a
+// number greater than the maximum capacity which an implementation supports,
+// than the test will pass as the number of teams is less than that value, but
+// the value itself has no functional purpose. The ICV is initialized to 0 per
+// device. Thus, if omp_get_num_teams returns a positive number as
+// expected, the test will pass when the ICV value is 0 or less than 0. In the
+// case where the OMP_NUM_TEAMS environmental variable is set to a positive
+// integer, or the omp_set_num_teams routine is given a positive integer as an
+// argument, the test will pass if and only if the omp_get_num_teams routine
+// returns an integer that is less than or equal to the positive ICV value.
+//----------------------------------------------------------------------------//
 #include "ompvv.h"
+#include <omp.h>
 
-#define N 1024
+int test_get_max_teams(int offload) {
+  int errors = 0;
+  int max_teams = omp_get_max_teams(); // on host
+  int num_teams = 0; // a value that is not possible for omp_get_num_teams()
+
+  // clang-format off
+  #pragma omp target map(tofrom : num_teams) if(offload)
+  {
+    #pragma omp teams 
+    {
+      num_teams = omp_get_num_teams();
+    }
+  }
+  // clang-format on
+
+  OMPVV_ERROR_IF(
+      max_teams <= 0 && num_teams <= 0,
+      "Number of teams reported is non-positive, should not be possible");
+  OMPVV_ERROR_IF(
+      max_teams > 0 && num_teams > max_teams,
+      "Number of teams reported exceeded max number of teams (max no. > 0)");
+  OMPVV_TEST_AND_SET(errors, max_teams <= 0 && num_teams <= 0);
+  OMPVV_TEST_AND_SET(errors, max_teams > 0 && num_teams > max_teams);
+
+  return errors;
+}
 
 int main() {
-
-        OMPVV_TEST_OFFLOADING;
-
-        int errors = 0;
-        int MAX_TEAMS = omp_get_max_teams(); //on host
-        int num_teams = MAX_TEAMS + 1; //a value that is not possible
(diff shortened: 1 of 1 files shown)


tests/5.1/teams/test_target_get_max_teams.c
```diff
@@ -0,0 +1,59 @@
+//--------------- test_target_get_max_teams.c --------------------------------//
+// OpenMP API Version 5.1 Nov 2020
+// ***************************
+// ROUTINE: omp_get_max_teams
+// ***************************
+// This test uses the omp_get_max_teams routine to check what the max teams
+// capacity is for the current device. If the routine returns 0, the nteams-var
```
fel-cab, line 7:
```suggestion
// capacity is for host and the current device. If the routine returns 0, the nteams-var
```

tests/5.1/teams/test_target_get_max_teams.c
```diff
@@ -0,0 +1,59 @@
...
+  {
+    #pragma omp teams 
+    {
+      num_teams = omp_get_num_teams();
+    }
+  }
+  // clang-format on
+
+  OMPVV_ERROR_IF(
+      max_teams <= 0 && num_teams <= 0,
```
seyonglee, line 41: The number of teams should be greater than or equal to 1 in any case, and thus we don't need `max_teams <= 0` condition here; delete `max_teams <= 0`.

tests/5.1/teams/test_target_get_max_teams.c
```diff
@@ -0,0 +1,59 @@
...
+      "Number of teams reported is non-positive, should not be possible");
+  OMPVV_ERROR_IF(
+      max_teams > 0 && num_teams > max_teams,
+      "Number of teams reported exceeded max number of teams (max no. > 0)");
+  OMPVV_TEST_AND_SET(errors, max_teams <= 0 && num_teams <= 0);
```
seyonglee, line 46: The number of teams should be greater than or equal to 1 in any case, and thus we don't need `max_teams <= 0` condition here; delete `max_teams <= 0`.

spophale:
Changes are required to capture the different error conditions.

tob2:
* I think the test could be _extended_ by explicitly setting the ICV via `omp_set_num_teams` and then checking again whether that value is not exceeded.
   * For the device it would need to be set in a target region – but I am not 100% sure that that guaranteed to be persistent between two target regions (in GCC it is).
   * For the device side, you need to query the maximal value of the _nteams-var_ ICV on the device side as it could be different. — This requires a separate 'omp target' as calling  omp_get_max_teams stricly nested in 'target teams' is not permitted.
   BTW: the OMP_NUM_TEAMS_DEV / OMP_NUM_TEAMS_DEV_<nr> environment variable can be use to set a different ICV value from the host.

* In terms of an implementation, the maximal number of teams can be different between 'target teams' + host fallback (== device(omp_get_initial_device())') and/or 'if(0)' and/or device execution. _In case of GCC, '3' is the default on the host but with host fallback ('if(0)' or device(omp_initial_device)), it is 0._
  Thus, you could add another check with 'omp teams' besides 'omp target teams if(offload).

andrewkallai:
tests/5.1/env_var/test_omp_num_teams_env_2.c sets the environmental directly, tests/5.1/runtime_calls/test_teams_region_routines.c sets the ICV through omp_set_num_teams routine, and PR728 is a test specifically for the omp_set_num_teams routine. I don't see a reason to extend this test file as it is suppossed to check the ICV for host and device environments regardless, and is suppossed to pass if someone explicitely sets the ICV before execution.

seyonglee:
> tests/5.1/env_var/test_omp_num_teams_env_2.c sets the environmental directly, tests/5.1/runtime_calls/test_teams_region_routines.c sets the ICV through omp_set_num_teams routine, and PR728 is a test specifically for the omp_set_num_teams routine. I don't see a reason to extend this test file as it is suppossed to check the ICV for host and device environments regardless, and is suppossed to pass if someone explicitely sets the ICV before execution.

We may not need to add `omp_set_num_teams` or set the environment variable `OMP_NUM_TEAMS` in this test, but as @tob2 mentioned, one copy of the `nteams-var` ICV exists per device, and thus different devices can have different values. However, the current test uses the same `nteams-var` ICV value of the host for both the host and device tests. Therefore, this benchmark will need to be extended to use device-specific values of the `teams-var` ICV for each subtest.

tests/5.1/teams/test_target_get_max_teams.c
```diff
@@ -0,0 +1,56 @@
...
+// returns an integer that is less than or equal to the positive ICV value.
+//----------------------------------------------------------------------------//
+#include "ompvv.h"
+#include <omp.h>
+
+int test_get_max_teams(int offload) {
+  int errors = 0;
+  int max_teams = omp_get_max_teams(); // on host
+  int num_teams = 0; // a value that is not possible for omp_get_num_teams()
+
```
spophale, line 30: #pragma omp target if(offload)
  {
     max_teams = omp_get_max_teams();
}

tests/5.1/teams/test_target_get_max_teams.c
```diff
@@ -29,9 +29,12 @@ int test_get_max_teams(int offload) {
   int max_teams;
   int num_teams = 0; // a value that is not possible for omp_get_num_teams()
 
-  #pragma omp target map(tofrom : max_teams, num_teams) if(offload)
+  #pragma omp target map(tofrom: max_teams) if(offload)
   {
     max_teams = omp_get_max_teams();
+  }
+  #pragma omp target map(tofrom : max_teams, num_teams) if(offload)
```
fel-cab, line 36:
```suggestion
  #pragma omp target map(tofrom : num_teams) if(offload)
```

spophale (approved):

Issue #756, opened 2023-07-20
test_target_has_device_addr.F90: Wrong testcase violating restrictions


tob2:
→ https://github.com/SOLLVE/sollve_vv/blob/master/tests/5.1/target/test_target_has_device_addr.F90
→ Was added in Pull Request #717

The code does:
```F90
    INTEGER, POINTER :: first_scalar_device_addr
...
    !$omp target map(from: first_scalar_device_addr, first_arr_device_addr) map(to: x, arr)
      first_scalar_device_addr => x
...
    !$omp end target
```

There are multiple problems with this:
1. "The association status of a list item that is a pointer must not be undefined unless it is a structure component and it results from a predefined default mapper." _(TR11 [160:15–16], "5.8.3 map Clause" under "Restrictions to the **map** clause")_

While only added in TR11, it is a bug fix to the specification and not a new feature. The background for that change is _(TR11 wording but feature is old)_:

> * If a list item in a map clause is an associated pointer that is not attach-ineligible and the pointer is not the base pointer of another list item in a map clause on the same construct, then it is treated as if its pointer target is implicitly mapped in the same clause.

The problem is that a compiler cannot distinguish an undefined pointer from an associated pointer.

**Solution:** you need to **nullify()** / **null()** the pointer before mapping.

2. The following feature – here, the second bullet applies (at least after nullifying):

> * If the association status of a list item with the POINTER attribute that appears in a map clause on the construct is associated upon entry to the target region, the list item must be associated with the same pointer target upon exit from the region.
> * If the association status of a list item with the POINTER attribute that appears in a map clause on the construct is disassociated upon entry to the target region, the list item must be disassociated upon exit from the region.

_(OpenMP 5.2 [287:29-31, 288:1-3], "Restrictions to the **target** construct" in "13.8 target Construct")_

While only in OpenMP 5.2, the issue is generic — and also the main reason that the testcase has problems. Namely, the following applies:

* At least for arrays, all Fortran implementations use an array descriptor, which is a struct containing the array bounds and usually also the data size, some attributes, the type, — and a pointer to the bare data.
* For "map(tofrom: array)", implementations usually copy-in (or firstprivatize) the array descriptor and only the actual data is copied from/to. This implies:
[...]

PR #757, opened 2023-07-20
Issue 756: Fix test_target_has_device_addr.F90

tob2:
Issue #756
Do not change pointer association inside the target region. Use detour via c_ptr for many checks, passing it back after c_f_pointer conversion.

See issue for more details.

@seyonglee @fel-cab @spophale – please review.

### DIFF
--- a/tests/5.1/target/test_target_has_device_addr.F90
+++ b/tests/5.1/target/test_target_has_device_addr.F90
@@ -15,6 +15,7 @@
 
 PROGRAM test_target_has_device_addr
   USE iso_fortran_env
+  USE iso_c_binding
   USE ompvv_lib
   USE omp_lib
   implicit none
@@ -33,6 +34,15 @@ CONTAINS
     INTEGER, POINTER :: first_arr_device_addr(:)
     INTEGER, POINTER :: second_scalar_device_addr
     INTEGER, POINTER :: second_arr_device_addr(:)
+    INTEGER, POINTER :: third_scalar_device_addr
+    INTEGER, POINTER :: third_arr_device_addr(:)
+    type(c_ptr) :: cptr_scalar1, cptr_arr1
+    type(c_ptr) :: cptr_scalar2, cptr_arr2
+    type(c_ptr) :: cptr_scalar3, cptr_arr3
+
+    nullify (first_scalar_device_addr, first_arr_device_addr)
+    nullify (second_scalar_device_addr, second_arr_device_addr)
+    nullify (third_scalar_device_addr, third_arr_device_addr)
 
     errors = 0
     x = 10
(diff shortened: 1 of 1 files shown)


seyonglee:
@tob2 Is not-changing-pointer-association inside the target region required by the OpenMP/Fortran standard? Or is it recommended as a good programming practice?

The original program (`test_target_has_device_addr.F90`) passes the test when compiled by GCC 13.1.1 if the following nullification statements are added:
    `nullify(first_scalar_device_addr, first_arr_device_addr)`
    `nullify(second_scalar_device_addr, second_arr_device_addr)`

tob2:
> Is not-changing-pointer-association inside the target region required by the OpenMP/Fortran standard? Or is it recommended as a good programming practice?

* In **OpenMP 5.0,** it is listed in the "**map** Clause" section under restrictions: https://www.openmp.org/spec-html/5.0/openmpsu109.html#x142-6240002.19.7.1 – "If the association status of a list item with the POINTER attribute is disassociated upon entry to a target region, the list item must be disassociated upon exit from the region."
* **In/since OpenMP 5.1,** it is moved to the "**target** Construct" section under restrictions: https://www.openmp.org/spec-html/5.1/openmpsu68.html#Q1-96-392 — "If the association status of a list item with the POINTER attribute that appears in a **map** clause on the construct is disassociated upon entry to the target region, the list item must be disassociated upon exit from the region."

_(The quotes above are for pointers that are **disassociated** upon entry to the target reagion. For **associated** pointers, see preceding bullet point in the spec.)_

_BTW: Regarding the third pointer association state, **undefined** (dangling, initialized, freed, unknown state): That's also invalid (with one exception) for obvious technical reasons, but the the wording for it was only added in TR11 → quoted in Issue #756 or see directly in TR11 [160:15–16], "5.8.3 map Clause"._

tob2:
@seyonglee @fel-cab @spophale – **Ping,** please review

@seyonglee – all questions answered? Or do you have additional ones?

tob2:
@seyonglee @fel-cab @spophale – **Ping,** please review

@seyonglee – all questions answered? Or do you have additional ones?

seyonglee (approved):
It looks good to me.

seyonglee:
Compiling this test using GCC V13.1.1 on Summit fails with the following error message: `unresolved symbol _gfortran_associated`. Does this mean that GCC V13.1.1 does not support the  `associated` intrinsic function on the device?

tob2:
GCC up to version 13 — on the device side — dorsnot automatically link neither the gfortran runtime library nor the math library. Hence, use "-foffload-options=-lgfortran".
* See example at [GCC 13 manualnon offload options](https://gcc.gnu.org/onlinedocs/gcc-13.2.0/gcc/C-Dialect-Options.html#index-foffload-options)

In GCC mainline, this has changed: if one of those two libraries is linked on the host, it will automatically also be linked on the device.
* Backporting to GCC OG13 (devel/omp/gcc-13) is planned.

Note: For arrays, gfortran uses a runtime library as ASSOCIATE not only checks the pointer address but als whether other properties like the shape are the same.  (As demanded by the Fortran standard.)

spophale (approved):

PR #761, opened 2023-08-08
Adding more tests with teams loop construct

lthakur007:
Please review..

### DIFF
--- a/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
+++ b/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
@@ -1,3 +1,11 @@
+//===-------------- test_target_teams_loop_allocate.c---------------------===//
+//
+// OpenMP API Version 5.0 Nov 2018
+// The tests in this file tests 'teams loop' construct coupled with map()
+// clause in combination with different allocator types
+//
+//===----------------------------------------------------------------------===//
+
 #include <omp.h>
 #include <stdio.h>
 #include <stdlib.h>
--- a/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
+++ b/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
@@ -31,7 +31,6 @@ int testTargetTeamsAllocateCl() {
 #pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
         uses_allocators(omp_cgroup_mem_alloc) \
         allocate(omp_low_lat_mem_alloc: local) firstprivate(local)
-//  #pragma omp loop private(local)
   for (int i = 0; i < N; i++) {
     local = a[i] + b[i];
     c[i] = local;
--- a/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
+++ b/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
@@ -28,10 +28,10 @@ int testTargetTeamsAllocateCl() {
     c[i] = 0;
   }
   // Execute on target
-#pragma omp target teams map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
+#pragma omp target teams loops map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
         uses_allocators(omp_cgroup_mem_alloc) \
         allocate(omp_low_lat_mem_alloc: local) private(local)
-  #pragma omp loop private(local)
+//  #pragma omp loop private(local)
   for (int i = 0; i < N; i++) {
     local = a[i] + b[i];
     c[i] = local;
@@ -60,7 +60,7 @@ int testTargetTeamsAllocateArrayDefaultAlloc() {
   // Execute on target
 #pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
         uses_allocators(omp_default_mem_alloc)\
-        allocate(omp_default_mem_alloc: local) private(local)
+        allocate(omp_default_mem_alloc: local) firstprivate(local)
   for (int i = 0; i < N; i++) {
     local[i] = a[i] + b[i];
     c[i] = local[i];
@@ -89,7 +89,7 @@ int testTargetTeamsAllocateArrayLargeCapAlloc() {
   // Execute on target
 #pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
         uses_allocators(omp_large_cap_mem_alloc)\
-        allocate(omp_large_cap_mem_alloc: local) private(local)
+        allocate(omp_large_cap_mem_alloc: local) firstprivate(local)
   for (int i = 0; i < N; i++) {
     local[i] = a[i] + b[i];
     c[i] = local[i];
@@ -145,7 +145,7 @@ int testTargetTeamsAllocateArrayHighBWAlloc() {
   // Execute on target
 #pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
         uses_allocators(omp_high_bw_mem_alloc)\
-        allocate(omp_high_bw_mem_alloc: local) private(local)
+        allocate(omp_high_bw_mem_alloc: local) firstprivate(local)
   for (int i = 0; i < N; i++) {
     local[i] = a[i] + b[i];
     c[i] = local[i];
@@ -174,7 +174,7 @@ int testTargetTeamsAllocateArrayLowLatAlloc() {
   // Execute on target
 #pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
         uses_allocators(omp_low_lat_mem_alloc)\
-        allocate(omp_low_lat_mem_alloc: local) private(local)
+        allocate(omp_low_lat_mem_alloc: local) firstprivate(local)
   for (int i = 0; i < N; i++) {
     local[i] = a[i] + b[i];
     c[i] = local[i];
@@ -203,7 +203,7 @@ int testTargetTeamsAllocateArrayCGroupAlloc() {
   // Execute on target
 #pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
         uses_allocators(omp_cgroup_mem_alloc)\
-        allocate(omp_cgroup_mem_alloc: local) private(local)
+        allocate(omp_cgroup_mem_alloc: local) firstprivate(local)
   for (int i = 0; i < N; i++) {
     local[i] = a[i] + b[i];
     c[i] = local[i];
@@ -232,7 +232,7 @@ int testTargetTeamsAllocateArrayPTeamAlloc() {
   // Execute on target
 #pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
         uses_allocators(omp_pteam_mem_alloc)\
-        allocate(omp_pteam_mem_alloc: local) private(local)
+        allocate(omp_pteam_mem_alloc: local) firstprivate(local)
   for (int i = 0; i < N; i++) {
     local[i] = a[i] + b[i];
     c[i] = local[i];
--- a/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
+++ b/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
@@ -28,9 +28,10 @@ int testTargetTeamsAllocateCl() {
     c[i] = 0;
   }
   // Execute on target
-#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
-        uses_allocators(omp_default_mem_alloc)\
-        allocate(omp_default_mem_alloc: local) private(local)
+#pragma omp target teams map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
+        uses_allocators(omp_cgroup_mem_alloc) \
+        allocate(omp_low_lat_mem_alloc: local) private(local)
+  #pragma omp loop private(local)
   for (int i = 0; i < N; i++) {
     local = a[i] + b[i];
     c[i] = local;
--- a/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
+++ b/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
@@ -29,7 +29,7 @@ int testTargetTeamsAllocateCl() {
   }
   // Execute on target
 #pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
-        uses_allocators(omp_cgroup_mem_alloc) \
+        uses_allocators(omp_low_lat_mem_alloc) \
         allocate(omp_low_lat_mem_alloc: local) firstprivate(local)
   for (int i = 0; i < N; i++) {
     local = a[i] + b[i];
--- a/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
+++ b/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
@@ -28,9 +28,9 @@ int testTargetTeamsAllocateCl() {
     c[i] = 0;
   }
   // Execute on target
-#pragma omp target teams loops map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
         uses_allocators(omp_cgroup_mem_alloc) \
-        allocate(omp_low_lat_mem_alloc: local) private(local)
+        allocate(omp_low_lat_mem_alloc: local) firstprivate(local)
 //  #pragma omp loop private(local)
   for (int i = 0; i < N; i++) {
     local = a[i] + b[i];
--- a/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
+++ b/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
@@ -246,6 +246,9 @@ int testTargetTeamsAllocateArrayPTeamAlloc() {
 
 int main() {
   int errors = 0;
+  
+  OMPVV_TEST_OFFLOADING;
+  
   OMPVV_TEST_AND_SET_VERBOSE(errors, testTargetTeamsAllocateCl());
   OMPVV_TEST_AND_SET_VERBOSE(errors, testTargetTeamsAllocateArrayDefaultAlloc());
   OMPVV_TEST_AND_SET_VERBOSE(errors, testTargetTeamsAllocateArrayLargeCapAlloc());
--- a/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
+++ b/tests/5.0/teams_loop/test_target_teams_loop_allocate.c
@@ -1,0 +1,250 @@
+#include <omp.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include "ompvv.h"
+
+#define N 1024
+
+/**
+  This is a basic test to demonstrate how allocate can be used with
+  target teams loop construct. Allocate clause is used to allocate a local
+  variable of size int.
+*/
+int testTargetTeamsAllocateCl() {
+  int a[N], b[N], c[N], local = 0;
+  int errors = 0;
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    c[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_default_mem_alloc)\
+        allocate(omp_default_mem_alloc: local) private(local)
+  for (int i = 0; i < N; i++) {
+    local = a[i] + b[i];
+    c[i] = local;
+  }
+  // Validate
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, c[i] != a[i] + b[i]);
+  }
+  OMPVV_TEST_AND_SET_VERBOSE(errors, local != 0);
+  return errors;
+}
+
+/**
+  This is a basic test to demonstrate use of allocate clause with
+  omp_default_mem_alloc allocator to allocate an array.
+*/
+int testTargetTeamsAllocateArrayDefaultAlloc() {
+  int a[N], b[N], c[N], local[N];
+  int errors = 0;
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    local[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_default_mem_alloc)\
+        allocate(omp_default_mem_alloc: local) private(local)
+  for (int i = 0; i < N; i++) {
+    local[i] = a[i] + b[i];
+    c[i] = local[i];
+  }
+  // Validate
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, c[i] != a[i] + b[i]);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, local[i] != 0);
+  }
+  return errors;
+}
+
+/**
+  This is a basic test to demonstrate use of allocate clause with
+  omp_large_cap_mem_alloc allocator to allocate an array.
+*/
+int testTargetTeamsAllocateArrayLargeCapAlloc() {
+  int a[N], b[N], c[N], local[N];
+  int errors = 0;
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    local[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_large_cap_mem_alloc)\
+        allocate(omp_large_cap_mem_alloc: local) private(local)
+  for (int i = 0; i < N; i++) {
+    local[i] = a[i] + b[i];
+    c[i] = local[i];
+  }
+  // Validate
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, c[i] != a[i] + b[i]);
+    OMPVV_TEST_AND_SET_VERBOSE(errors, local[i] != 0);
+  }
+  return errors;
+}
+
+/**
+  This is a basic test to demonstrate use of allocate clause with
+  omp_const_mem_alloc allocator to allocate an variable.
+*/
+int testTargetTeamsAllocateConstAlloc() {
+  int a[N], b[N];
+  const int local = 3;
+  int errors = 0;
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 0;  // Odd
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N]) map(from: b[0:N])\
+        firstprivate(local) uses_allocators(omp_const_mem_alloc)\
+         allocate(omp_const_mem_alloc: local)
+  for (int i = 0; i < N; i++) {
+    b[i] = a[i] + local;
+  }
+  // Validate
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET_VERBOSE(errors, b[i] != a[i] + local);
+  }
+  return errors;
+}
+
+/**
+  This is a basic test to demonstrate use of allocate clause with
+  omp_high_bw_mem_alloc allocator to allocate an array.
+*/
+int testTargetTeamsAllocateArrayHighBWAlloc() {
+  int a[N], b[N], c[N], local[N];
+  int errors = 0;
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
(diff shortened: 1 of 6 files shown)


tests/5.0/teams_loop/test_target_teams_loop_depend.c
```diff
@@ -0,0 +1,49 @@
...
+  int a[N], b[N], c[N];
+  int errors = 0;
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = i;
+    b[i] = 0;
+    c[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(tofrom: a[0:N]) depend(out: a[0:N])
```
seyonglee, line 24: Add `nowait` clauses to this and other target directives, and add `taskwait` directive after the last target construct. 
(Without `nowait` clauses, the generated target task will be an included task; thus, this test cannot check whether the `depend` clause works as intended.)

seyonglee:
Add brief descriptions to the beginning of each test using the format used in other tests.

seyonglee:
GCC 13.1.1 on the Summit system fails on `test_target_teams_loop_allocate.c` with the following error messages:
(the other tests pass with GCC 13.1.1.)

> 25:17: error: 'local' specified in 'allocate' clause but not in an explicit privatization clause

However, this seems to be a compiler bug.

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,250 @@
+#include <omp.h>
```
spophale, line 1: Add test description.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,238 @@
+#include <omp.h>
```
spophale, line 1: Add test description here, see format of existing tests.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,238 @@
...
+int testDefaultMapToFrom() {
+  int errors = 0;
+  int device_data[N], host_data[N];
+  for (int i = 0; i < N; i++) {
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(tofrom)
+  for (int i = 0; i < N; i++) {
+    host_data[i] = device_data[i];
```
spophale, line 21: Should be  host_data[i] += device_data[I]; //to verify if host_data was copied correctly.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,238 @@
...
+int testDefaultMapTo() {
+  int errors = 0;
+  int device_data[N], host_data[N];
+  for (int i = 0; i < N; i++) {
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(to)
+  for (int i = 0; i < N; i++) {
+    host_data[i] = device_data[i];
```
spophale, line 67: Suggested change: host_data[i] += device_data[i];

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,238 @@
...
+  }
+  // checking results
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET(errors, host_data[i] != 0);
```
spophale, line 71: On a shared memory system or host fallback host_data will be updated (which is correct). This will lead to incorrect error reporting.

Better to do a map(from: host_data) on the target and check for updated host_data.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,238 @@
...
+int testDefaultMapDefault() {
+  int errors = 0;
+  int device_data[N], host_data[N];
+  for (int i = 0; i < N; i++) {
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(default)
+  for (int i = 0; i < N; i++) {
+    host_data[i] = device_data[i];
```
spophale, line 89: Suggested change: host_data[i] += device_data[i];

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,238 @@
...
+int testDefaultMapToFromAggr() {
+  int errors = 0;
+  int device_data[N], host_data[N];
+  for (int i = 0; i < N; i++) {
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(tofrom:aggregate)
+  for (int i = 0; i < N; i++) {
+    host_data[i] = device_data[i];
```
spophale, line 135: Suggested change: host_data[i] += device_data[i];

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,238 @@
...
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(alloc) map(to: device_data[0:N])
+  for (int i = 0; i < N; i++) {
+    host_data[i] = device_data[i];
+  }
+  // checking results to verify host_data is unchanged
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET(errors, host_data[i] != 0);
```
spophale, line 220: Same issue as mentioned before for host fallback.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,238 @@
...
+/**
+  This is a basic test to demonstrate defaultmap(tofrom:pointer)
+  clause with target teams loop construct.
+*/
+int testDefaultMapToFromPtr() {
+  int errors = 0;
+  int *ptr = 0;
+#pragma omp target teams loop defaultmap(tofrom:pointer)
+  for (int i = 0; i < N; i++) {
+    ptr = (int*)0xefefefef;
```
spophale, line 196: Multiple threads could be assigning to the same pointer variable. 
Either have an atomic write or an array of ptrs.

tob2:
> GCC 13.1.1 on the Summit system fails on `test_target_teams_loop_allocate.c` with the following error messages: (the other tests pass with GCC 13.1.1.)
> 
> > 25:17: error: 'local' specified in 'allocate' clause but not in an explicit privatization clause
> 
> However, this seems to be a compiler bug.

Actually, it is a code bug:

```C
#pragma omp target teams loop ... allocate(omp_default_mem_alloc: local) private(local)
```

Namely:
* the **private** clause is accepted by all of them: **loop**, **teams** and **target**, but "Clauses on Combined and Composite Constructs" section states: "The effect of the **private** clause is as if it is applied only to the innermost leaf construct that permits it." — Thus, **private** applies to only to **loop**.
* the **allocate** clause is accepted only by **teams** and **target** but not by **loop**. The "Clauses on Combined and Composite Constructs" section states: "The effect of the allocate clause is as if it is applied to all leaf constructs that permit the clause and to which a data-sharing attribute clause that may create a private copy of the same list item is applied." → In principle, **teams** and **target** would qualify, but neither of them has  a data-sharing attribute (as the 'private' is already consumed by 'loops')
* Additionally, the following is required — but not fulfilled **allocate**: "Restrictions to clauses on combined and composite constructs are as follows:
   * A clause that appears on a combined or composite construct must apply to at least one of the leaf constructs per the rules defined in this section."

tests/5.0/teams_loop/test_target_teams_loop_depend.c
lthakur007: added nowait as suggested...

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: The intent of using "defaultmap(to)" is to ensure that host side memory(process's stack memory) "host_data" is not updated. Hence the check "host_data[i] != 0" at line 71.
Because host_data is host memory allocated in stack, the expectation is that it doesn't come under "shared memory" aspects... Please let us know if your opinion differs...

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: Made changes as suggested...

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: Made changes as suggested...

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: Made changes as suggested...

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: Made changes as suggested...

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: Made changes as suggested...

lthakur007:
Hi @seyonglee and @seyonglee ,
Could you please share your opinion on the comment made by @tob2 .
Please advise if we should remove the test file or any modification would allow allocate() clause still applicable.

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,250 @@
...
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    c[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_default_mem_alloc)\
+        allocate(omp_default_mem_alloc: local) private(local)
```
tob2, line 25: > Please advise if we should remove the test file or any modification would allow allocate() clause still applicable.

A bit pointless, but you could add a 'firstprivate' clause (which then applies to target and teams, but not to 'loop') – and keep the 'private' for 'loop'.

But somewhat more sensible would be to split 'target teams loop' as follows – such that the private applies to 'teams' – That's still not very useful as the variable is twice privatized - once on 'teams' level and once on 'loop' level, but still.
```suggestion
#pragma omp target teams map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
        uses_allocators(omp_cgroup_mem_alloc) \
        allocate(omp_low_lat_mem_alloc: local) private(local)
  #pragma omp loop private(local)
```

How about something like the following? Keeping the private but adding another firstprivatized variable, which is now put into "const" memory to make it more interesting than "default":
```C++
int magic_value = 42;
#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
        uses_allocators(omp_default_mem_alloc) \
        allocate(omp_const_mem_alloc: magic_value) firstprivate(magic_value) \
        private(local)
  for (int i = 0; i < N; i++) {
    local = a[i] + b[i] + magic_value;
    c[i] = local;
  }
```

seyonglee: For me, the last option (use `first private(magic_value)`) seems to be the best.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
seyonglee: > Because host_data is host memory allocated in stack, the expectation is that it doesn't come under "shared memory" aspects... Please let us know if your opinion differs…

I don't understand the above comment; `host_data` on the host will be updated if on a shared memory system regardless of whether `host_data` is in the stack memory or global memory. 
I think that a better way to handle this issue is to print a warning message if this test runs on the shared memory system or host fallback occurs.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: added map(from: host_data) as suggested...

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: added map(from: host_data) as suggested...

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
lthakur007: added description

lthakur007:
addressed review comments.. please approve...

tests/5.0/teams_loop/test_target_teams_loop_depend.c
```diff
@@ -0,0 +1,49 @@
+#include <omp.h>
```
seyonglee, line 1: Add a test description here.

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    c[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
+        uses_allocators(omp_cgroup_mem_alloc) \
+        allocate(omp_low_lat_mem_alloc: local) private(local)
+  #pragma omp loop private(local)
```
seyonglee, line 34: The purpose of this test is to test the `teams loop` combined construct, but this sub-test uses separate `teams` construct and `loop` construct, defeating the purpose of this test.

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    local[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_large_cap_mem_alloc)\
+        allocate(omp_large_cap_mem_alloc: local) private(local)
```
seyonglee, line 92: The error @tob2 mentioned still exists here (`error: 'local' specified in 'allocate' clause but not in an explicit privatization clause`); one possible solution is to use `firstprivate(local)` instead of `private(local)`, but using the `firstprivate` clause will incur unnecessary initialization of the firstprivate variable, `local`.

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    local[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_high_bw_mem_alloc)\
+        allocate(omp_high_bw_mem_alloc: local) private(local)
```
seyonglee, line 148: The error @tob2 mentioned still exists here (`error: 'local' specified in 'allocate' clause but not in an explicit privatization clause`); one possible solution is to use `firstprivate(local)` instead of `private(local)`, but using the `firstprivate` clause will incur unnecessary initialization of the firstprivate variable, `local`.

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    local[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_low_lat_mem_alloc)\
+        allocate(omp_low_lat_mem_alloc: local) private(local)
```
seyonglee, line 177: The error @tob2 mentioned still exists here (`error: 'local' specified in 'allocate' clause but not in an explicit privatization clause`); one possible solution is to use `firstprivate(local)` instead of `private(local)`, but using the `firstprivate` clause will incur unnecessary initialization of the firstprivate variable, `local`.

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    local[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_pteam_mem_alloc)\
+        allocate(omp_pteam_mem_alloc: local) private(local)
```
seyonglee, line 235: The error @tob2 mentioned still exists here (`error: 'local' specified in 'allocate' clause but not in an explicit privatization clause`); one possible solution is to use `firstprivate(local)` instead of `private(local)`, but using the `firstprivate` clause will incur unnecessary initialization of the firstprivate variable, `local`.

lthakur007: Replaced "private(local)" with  "firstprivate(local)" as suggested...

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    local[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_cgroup_mem_alloc)\
+        allocate(omp_cgroup_mem_alloc: local) private(local)
```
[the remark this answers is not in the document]
lthakur007: Replaced "private(local)" with  "firstprivate(local)" as suggested...

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
lthakur007: Replaced "private(local)" with  "firstprivate(local)" as suggested...

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
lthakur007: Replaced "private(local)" with  "firstprivate(local)" as suggested...

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
lthakur007: Replaced "private(local)" with  "firstprivate(local)" as suggested...

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    local[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N])\
+        uses_allocators(omp_default_mem_alloc)\
+        allocate(omp_default_mem_alloc: local) private(local)
```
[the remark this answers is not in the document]
lthakur007: Replaced "private(local)" with  "firstprivate(local)" as suggested...

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
lthakur007: Replaced "teams" with  "teams loop" as suggested...

lthakur007:
Hi @seyonglee and @spophale ,
I have addressed the review comments. Please consider to scrutinize the fixes...
Fixes are working fine on AMD platforms....

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+  int a[N], b[N], c[N], local = 0;
+  int errors = 0;
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    c[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loops map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
```
seyonglee, line 31: Change `loops` to `loop`

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
seyonglee: Change `map(from: host_data)` to `map(tofrom: host_data)` since `host_data` is read first in the target region.
Change  `OMPVV_TEST_AND_SET(errors, host_data[i] != 0);` to `OMPVV_TEST_AND_SET(errors, host_data[i] != i);`

seyonglee: One remaining issue is that this test may not necessarily check the behavior of the `defaultmap(to)` clause.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,250 @@
...
+*/
+int testDefaultMapAlloc() {
+  int errors = 0;
+  int device_data[N], host_data[N];
+  for (int i = 0; i < N; i++) {
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(alloc) map(to: device_data[0:N])\
+         map(from: host_data)
```
seyonglee, line 226: Delete `map(from: host_data)`

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
spophale: @lthakur007, you can potentially add a constant based on omp_is_initial_device() to the host_data to differentiate between host_fallback vs. target execution.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,249 @@
...
+  clause with target teams loop construct.
+*/
+int testDefaultMapTo() {
+  int errors = 0;
+  int device_data[N], host_data[N];
+  for (int i = 0; i < N; i++) {
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(to) map(from: host_data)
```
spophale, line 74: needs to be:  map(tofrom: host_data)

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,249 @@
...
+  for (int i = 0; i < N; i++) {
+    host_data[i] += device_data[i];
+  }
+  // checking results
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET(errors, host_data[i] != 0);
```
spophale, line 80: Needs to check host_data[i] == i

spophale:
@lthakur007 , generally arrays are not the best data types to check if defaultmaps were correctly applied as they have an implicit data mapping of toform on target directive.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,249 @@
...
+  }
+  // checking results to verify host_data is unchanged
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET(errors, host_data[i] != 0);
+  }
+  return errors;
+}
+
+int main() {
+  int errors = 0;
```
spophale, line 237: Add:
int is_offloading;
OMPVV_TEST_AND_SET_OFFLOADING(is_offloading);


This will ensure that the correct message is printed. Like "Test passed on device .."

spophale:
Add for all tests in main:
int is_offloading;
OMPVV_TEST_AND_SET_OFFLOADING(is_offloading);

This will ensure that the correct message is printed. Like "Test passed on device .."

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,259 @@
...
+        uses_allocators(omp_cgroup_mem_alloc) \
+        allocate(omp_low_lat_mem_alloc: local) private(local)
```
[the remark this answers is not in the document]
lthakur007: Observed that its already changed...

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
lthakur007: Observed that its already changed...

lthakur007:
Hi @seyonglee and @spophale ,
I have addressed review comments... please consider the fixes....

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,249 @@
...
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(firstprivate)
+  for (int i = 0; i < N; i++) {
+    host_data[i] = device_data[i];
+  }
+  // checking results
+  for (int i = 0; i < N; i++) {
+    OMPVV_TEST_AND_SET(errors, host_data[i] != 0);
```
spophale, line 126: This test will fail when host fallback occurs OR machine has a unified address space.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,249 @@
...
+/**
+  This is a basic test to demonstrate defaultmap(tofrom:scalar)
+  clause with target teams loop construct.
+*/
+int testDefaultMapToFromScalar() {
+  int errors = 0;
+  int x = 0;
+#pragma omp target teams loop defaultmap(tofrom:scalar)
+  for (int i = 0; i < N; i++) {
+    x = N;
```
spophale, line 164: "x" is shared across the teams and there is a race to update it. It may not be apparent as all threads are setting it ti N, but this leads to a non-conformant OpenMP program.
 To check the 'tofrom' consider: 
#pragma omp target teams loop defaultmap(tofrom:scalar) map(from: array) 
for (int i = 0; i < N; i++) {
#pragma omp atomic
    x += i; //this checks if the 'to' for x is working correctly
array[i] = x + 2;
}
//Then check array and x values on host

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
```diff
@@ -0,0 +1,249 @@
...
+  clause with target teams loop construct.
+*/
+int testDefaultMapToFromAggr() {
+  int errors = 0;
+  int device_data[N], host_data[N];
+  for (int i = 0; i < N; i++) {
+    device_data[i] = i;
+    host_data[i] = 0;
+  }
+#pragma omp target teams loop defaultmap(tofrom:aggregate) map(from: host_data)
```
spophale, line 142: defaultmap will not map host_data "to" the device since there is a explicit time.

spophale:
@lthakur007 please review our comments at your convenience.

tests/5.0/teams_loop/test_target_teams_loop_defaultmap.c
spophale, line 142:
```suggestion
#pragma omp target teams loop defaultmap(tofrom:aggregate) map(tofrom: host_data)
```

spophale (approved):

tests/5.0/teams_loop/test_target_teams_loop_allocate.c
```diff
@@ -0,0 +1,261 @@
...
+  // Data Inititalize
+  for (int i = 0; i < N; i++) {
+    a[i] = 2*i;  // Even
+    b[i] = 2*i + 1;  // Odd
+    c[i] = 0;
+  }
+  // Execute on target
+#pragma omp target teams loop map(to: a[0:N], b[0:N]) map(from: c[0:N]) \
+        uses_allocators(omp_cgroup_mem_alloc) \
+        allocate(omp_low_lat_mem_alloc: local) firstprivate(local)
```
fel-cab, line 33:
```suggestion
        uses_allocators(omp_low_lat_mem_alloc) \
        allocate(omp_low_lat_mem_alloc: local) firstprivate(local)
```
Is this correct?
Otherwise the compilation error is geenerated!
```
error: allocator must be specified in the 'uses_allocators' clause
        allocate(omp_low_lat_mem_alloc: local) firstprivate(local)
                 ^~~~~~~~~~~~~~~~~~~~~
1 error generated.
```

lthakur007:
Hi @fel-cab ,
I have committed the changes you suggested. Please consider the PR for further processing...

PR #764, opened 2023-08-23
Add new_test/5.1/test_atomic_compare_device.F90

seyonglee:
While the original C test uses the `>` operator for the atomic compare operation, Fortran atomic compare construct allows only the `==` operator. Therefore, the Fortan test uses a slightly different algorithm to atomically perform the max operation using the `atomic compare capture` directive.

[Evaluation Results on Summit]
- GCC 13.1.1: both C and Fortran tests passed.
- XL 16.1.1-10:
    - Fortran test failed: line 63.13: 1514-050 (S) Specification statement is out of order. 
- NVHPC 22.11:
    - C test failed: line 36: error: invalid text in pragma
    - Fortran test failed: NVFORTRAN-S-0034-Syntax error at or near COMPARE 
- LLVM 17.0.0:
    - C test passed.

### DIFF
--- a/tests/5.1/atomic/test_atomic_compare_device.F90
+++ b/tests/5.1/atomic/test_atomic_compare_device.F90
@@ -1,0 +1,81 @@
+!//===------ test_atomic_compare_device.F90 --------------------------===//
+!
+! OpenMP API Version 5.1 Nov 2020
+!
+! Adapted from OpenMP example video https://www.youtube.com/watch?v=iS6IG7nzCSo
+! Creates an array with random numbers, and uses atomic compare to find the max,
+! testing against non-parallel maximum.
+!
+!//===----------------------------------------------------------------------===//
+
+#include "ompvv.F90"
+
+#define N 100
+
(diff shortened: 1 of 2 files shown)


tob2:
> While the original C test uses the `>` operator for the atomic compare operation, Fortran atomic compare construct allows only the `==` operator.

Fortran permits the MIN/MAX intrinsic, matching semantically "<". (In TR12, "<" has been added for symmetry.)

seyonglee:
> > While the original C test uses the `>` operator for the atomic compare operation, Fortran atomic compare construct allows only the `==` operator.
> 
> Fortran permits the MIN/MAX intrinsic, matching semantically "<". (In TR12, "<" has been added for symmetry.)

You are correct, but I didn't use the MAX intrinsic here since this test is for `atomic compare` (the Fortran intrinsics are allowed for `atomic update`).

spophale (approved):

fel-cab (approved):

PR #765, opened 2023-08-23
Add new_test/5.1/teams/test_target_teams_default_firstprivate.F90

seyonglee:
Added a new Fortran test and fixed minor bugs in the original C test.

[Evaluation Results on Summit]
- GCC 13.1.1:
- XL 16.1.1-10:
    - C test failed on the host with the following warning: Number of teams was unexpected, test results likely inconclusive
    - Fortran test passed.
- NVHPC 22.11:
    - C test failed: line 31: error: invalid text in pragma
    - Fortran test passed.
- LLVM 17.0.0:
    - C test passed.

### DIFF
--- a/tests/5.1/teams/test_target_teams_default_firstprivate.F90
+++ b/tests/5.1/teams/test_target_teams_default_firstprivate.F90
@@ -41,7 +41,7 @@ CONTAINS
         num_teams = omp_get_num_teams()
         ! check first private
         IF (not_shared .NE. 5) then
-          errors = error + 1
+          errors = errors + 1
         END IF
     END IF 
     DO i=1, omp_get_num_teams()
--- a/tests/5.1/teams/test_target_teams_default_firstprivate.F90
+++ b/tests/5.1/teams/test_target_teams_default_firstprivate.F90
@@ -1,0 +1,53 @@
+!===--- test_target_teams_default_firstprivate.F90 --------------------------===//
+!
+! OpenMP API Version 5.1 Nov 2020
+!
+! This test uses the default(firstprivate) clause on a teams directive. The test
+! validates that when the default(firstprivate) clause is present all
+! variables without explicit sharing rules are not avaialble outside the region
+! and are private to each team. The default(firstprivate) clause is tested using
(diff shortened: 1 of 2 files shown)


spophale (approved):

tests/5.1/teams/test_target_teams_default_firstprivate.F90
```diff
@@ -0,0 +1,53 @@
...
+    num_teams = 0
+
+    !$omp target teams default(firstprivate) map(tofrom: num_teams) shared(num_teams) num_teams(OMPVV_NUM_TEAMS_DEVICE)
+    IF (omp_get_team_num() .EQ. 0) then
+        num_teams = omp_get_num_teams()
+    END IF 
+    DO i=1, omp_get_num_teams()
+        not_shared = not_shared + 5
+    END DO
+    !$omp end target teams
```
fel-cab, line 46:
```suggestion
    !$omp target teams default(firstprivate) map(tofrom: num_teams,errors) shared(num_teams,errors) num_teams(OMPVV_NUM_TEAMS_DEVICE)
    IF (omp_get_team_num() .EQ. 0) then
        num_teams = omp_get_num_teams()
        ! check first private
        IF (not_shared .NE. 5) then
          errors = error + 1
        END IF
    END IF 
    DO i=1, omp_get_num_teams()
        not_shared = not_shared + 5
    END DO
    !$omp end target teams
```
I'm not sure this is the best way, but I think we should add the check for not_shared to be 5 inside the target region, to make sure it is first private.
Furthermore, in the comments at lines 11 and 12, it says that the test checks for race conditions, I don't see that.

andrewkallai: On line 44 errors = error +1 should be changed to errors = errors + 1
Otherwise, the test looks good, but the C test should be modified according to the additional check for firstprivate semantics added in the Fortran test.

seyonglee: Updated as suggested.

fel-cab (approved):

PR #767, opened 2023-08-30
Add new_test/5.1/target_update/test_target_update_to_present.F90

seyonglee:
One common issue on both C and Fortran tests is that the tests assume that a target device with discrete memory exists; these tests will fail if running on the host or the target device with unified shared memory.

[Results on Summit]
- GCC 13.1.1:
- XL 16.1.1-10:
    - C test failed on the host.
    - Fortran test failed: line 68.13: 1516-036 (S) Entity omp_target_is_present has undefined type.
- NVHPC 22.11:
    - C test failed: line 55: error: identifier "present" is undefined
    - Fortran test failed: NVFORTRAN-S-0034-Syntax error at or near :
- LLVM 17.0.0:
    - C test passed.

### DIFF
--- a/tests/5.1/target_update/test_target_update_to_present.F90
+++ b/tests/5.1/target_update/test_target_update_to_present.F90
@@ -93,7 +93,7 @@ CONTAINS
             errors = errors + 1
         END IF
 
-        !$omp target map(tofrom: errors) defaultmap(none) map(from: scalar_var, A, new_struct)
+        !$omp target map(tofrom: errors) defaultmap(none) map(to: scalar_var, A, new_struct)
         IF(scalar_var .NE. 1) THEN
             errors = errors + 1
         END IF
--- a/tests/5.1/target_update/test_target_update_to_present.F90
+++ b/tests/5.1/target_update/test_target_update_to_present.F90
@@ -31,8 +31,6 @@ PROGRAM main
     USE omp_lib
     implicit none
     
-    OMPVV_TEST_OFFLOADING
-
     OMPVV_TEST_VERBOSE(test_target_update_to_present() .ne. 0)
 
     OMPVV_REPORT_AND_RETURN()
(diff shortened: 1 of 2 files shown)


seyonglee:
Change the code to skip the first test if target offloading is not enabled (running on the host).
[Results on Summit]
- GCC 13.1.1: 
- XL 16.1.1-10:
    - Fortran test failed: line 72.17: 1516-036 (S) Entity omp_target_is_present has undefined type.
- NVHPC 22.11:
    - C test failed: line 61: error: identifier "present" is undefined
    - Fortran test failed: NVFORTRAN/power Linux 22.11-0: compilation completed with severe errors
- LLVM 17.0.0: 
    - C test passed.

tests/5.1/target_update/test_target_update_to_present.F90
```diff
@@ -0,0 +1,117 @@
...
+            errors = errors + 1
+        END IF
+        IF (omp_target_is_present(A_cptr, t) .EQ. 0) THEN
+            errors = errors + 1
+        END IF
+        IF (omp_target_is_present(new_struct_cptr, t) .EQ. 0) THEN
+            errors = errors + 1
+        END IF
+
+        !$omp target map(tofrom: errors) defaultmap(none) map(from: scalar_var, A, new_struct)
```
fel-cab, line 96:
```suggestion
        !$omp target map(tofrom: errors) defaultmap(none) 
```
The` scalar_var, A, new_struct ` variables are not modified inside the target region and there is no need to map them from it

spophale (approved):
With the modification suggested.

tests/5.1/target_update/test_target_update_to_present.F90
seyonglee: We cannot delete the map clause because of the `defaultmap(none)`.
In fact, it does not matter which map type is used in the `map(from: scalar_var, A, new_struct)` since these variables are already present. Anyway, I changed the map type from `from` to `to`.

PR #768, opened 2023-08-30
test_allocate_allocator.c: Fix 'omp allocate' usage

tob2:
There were two issues with `omp allocate`:

First, the variable `x` is declared (`int *x;`) before the the `omp_allocator_handle_t` variable `x_alloc` is declared and initialized; while the `#pragma omp allocate(x) allocator(x_alloc)` line comes after the latter, there is a potential ordering issue: Per C semantic, `x` is available before `x_alloc` but before it can exist and be initialized, it has to be allocated - but for this `x_alloc` has to exist.

Solution: Moving the `x` declaration after the `x_alloc` declaration and initialization solves the issue.

Secondly, the OpenMP semantic states that the lifetime of an omp-allocated automatic (stack) variable such as `x` is its scope, i.e. it ends at the enclosing `}`. However, as allocator used for allocation is also used for deallocation, its lifetime must exceed the liftime of `x`. Thus, `omp_destory_allocator` cannot be in in the same scope as 'x'.

Solution: Place the omp-allocated variables in an inner scope such that 'x_alloc' has a longer lifetime.

_Additionally, some comments have been added to make it clearer to the code reader what happens and why code is as it is._
* * *
@spophale @fel-cab @ — please review. Thanks!

* * *

While the second issue did not show up when trying it with GCC (not even with `-fsanitize=address,undefined`), the first one shows up in the current draft patch as:
```
test_allocate_allocator.c:31:21: error: allocator variable ‘x_alloc’ must be declared before ‘x’`.
```
_[BTW: This topic has now an associated OpenMP spec issue (Issue 3676).]_

While a compiler could handle the first issue for this test case (by doing what the patch does), the following example causes an unsolvable hen-and-egg problem:
```C
int n=4;
omp_allocator_handle_t n_alloc = (omp_allocator_handle_t) (omp_default_mem_alloc + n);
#pragma omp allocate(n) allocator(n_alloc)
```
_This assumes that the allocator exists, but in most/all implementation, `omp_default_mem_alloc + 4` = `omp_low_lat_mem_alloc` as that's the rather obvious choice when implementing the predefined allocators (as enum values). – But playing around with arguments to omp_init_allocator, e.g. for the second argument (traits size), a variant of this issue can be created. – Both are not very sensible, but illustrate the issue._

### DIFF
--- a/tests/5.0/allocate/test_allocate_allocator.c
+++ b/tests/5.0/allocate/test_allocate_allocator.c
@@ -15,6 +15,7 @@
 #include <omp.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdint.h>
 #include "ompvv.h"
 
 #define N 1024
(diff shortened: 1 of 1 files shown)


tob2:
@spophale @fel-cab @ — please review. Thanks

tests/5.0/allocate/test_allocate_allocator.c
```diff
@@ -15,32 +15,56 @@
...
 int test_allocate_allocator() {
   int errors = 0;
 
-  int* x;
+  int n = N;
   omp_memspace_handle_t x_memspace = omp_default_mem_space;
   omp_alloctrait_t x_traits[1] = {{omp_atk_alignment, 64}};
   omp_allocator_handle_t x_alloc = omp_init_allocator(x_memspace, 1, x_traits);
+  {
+    // x and y must be in an inner scope as their lifetime must exceed the
```
fel-cab, line 31:
```suggestion
    // x and y must be in an inner scope as their lifetime mustn't exceed the
```
?

fel-cab (approved):

spophale (approved):

PR #769, opened 2023-09-06
Add new_test/5.1/target_update/test_target_update_iterator.F90

seyonglee:
[Results on Summit]
- GCC 13.1.1:
    - C test failed: warning: implicit declaration of function 'iterator'; error: 'it' undeclared
    - Fortran test failed: Error: Syntax error in OpenMP variable list at (1)
- XL 16.1.1-10:
    - Fortran test failed: line 79.47: 1515-019 (S) Syntax is incorrect.
- NVHPC 22.11:
    - C test failed: line 62: error: identifier "iterator" is undefined
    - Fortran test failed: NVFORTRAN/power Linux 22.11-0: compilation completed with severe errors (in line 79)
- LLVM 17.0.0:
    - C test failed: error: call to undeclared function 'iterator'; ISO C99 and later do not support implicit function declarations

### DIFF
--- a/tests/5.1/target_update/test_target_update_iterator.F90
+++ b/tests/5.1/target_update/test_target_update_iterator.F90
@@ -50,7 +50,6 @@ CONTAINS
         TYPE(test_struct), INTENT(IN), POINTER :: s
         INTEGER :: i
         s%len = N
-        ALLOCATE(s%data(N))
         DO i = 1, s%len
             s%data(i) = i + 1
         END DO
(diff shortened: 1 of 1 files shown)


tests/5.1/target_update/test_target_update_iterator.F90
```diff
@@ -0,0 +1,93 @@
...
+        DO i = 1, s%len
+            s%data(i) = i
+        END DO
+    END SUBROUTINE
+
+    SUBROUTINE init_again(s)
+        TYPE(test_struct), INTENT(IN), POINTER :: s
+        INTEGER :: i
+        s%len = N
+        ALLOCATE(s%data(N))
```
fel-cab, line 53: is it necessary to allocate data again in this subrutine?, given that the `s` (`new_struct`) was allocated in `init`?

seyonglee: You are correct; we don't need to reallocate the data.

spophale (approved):

fel-cab (approved):

PR #770, opened 2023-09-06
Add new_test/5.1/teams/test_target_teams_thread_limit.F90

seyonglee:
[Results on Summit]
- GCC 13.1.1: 
- XL 16.1.1-10:
    - C test passed but ran on the host with the following warning: The number of teams was unexpected, the test results are likely inconclusive
    - Fortran test failed: line 45.48: 1515-022 (S) Syntax Error: Extra token " thread_limit " was found. The token is ignored.
- NVHPC 22.11:
    - C test failed: line 39: error: invalid text in pragma (in thread_limit)
    - Fortran test failed: NVFORTRAN-S-0533-Clause 'THREAD_LIMIT' not allowed in OMP TARGET
- LLVM 17.0.0: 
    - C test passed.

### DIFF
--- a/tests/5.1/teams/test_target_teams_thread_limit.F90
+++ b/tests/5.1/teams/test_target_teams_thread_limit.F90
@@ -1,0 +1,69 @@
+!===--- test_target_teams_thread_limit.F90 --------------------------===//
+!
+! OpenMP API Version 5.1 Nov 2020
+!
+! This test uses the thread_limit clause on the target construct. Specifically
+! testing if a thread_limit on a target construct properly carries
+! down to the nested teams construct, as if it were directly on the construct
+! as defined in the spec. The test validates that only the specified 
+! threads are created by summing a shared variable across all threads 
+! (and teams). If the threads are correctly limited, this should produce the
+! expected value. Additional warnings are sent if specific issues occur.
+!
+!//===----------------------------------------------------------------------===//
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_target_teams_thread_limit
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+  OMPVV_TEST_OFFLOADING
+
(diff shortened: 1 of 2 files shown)


spophale:
@seyonglee since both target and teams accepts thread_limit, we should rename this test to test_target_thread_limit[_teams] and have a separate test for teams +thread_limt. Thoughts ?

seyonglee:
> @seyonglee since both target and teams accepts thread_limit, we should rename this test to test_target_thread_limit[_teams] and have a separate test for teams +thread_limt. Also, not sure which folder this goes in.. target or teams.
> 
> Thoughts ?

How about this?
1) Copy these tests, rename the copies as `test_target_thread_limit`, and move them to` test/5.1/target` directory.
2) In these original tests, move `thread_limit` clause to the teams directive.

fel-cab:
> > @seyonglee since both target and teams accepts thread_limit, we should rename this test to test_target_thread_limit[_teams] and have a separate test for teams +thread_limt. Also, not sure which folder this goes in.. target or teams.
> > Thoughts ?
> 
> How about this?
> 
> 1. Copy these tests, rename the copies as `test_target_thread_limit`, and move them to` test/5.1/target` directory.
> 2. In these original tests, move `thread_limit` clause to the teams directive.

I think it is a good solution

fel-cab (approved):

seyonglee:
Updated the following things:
        - Move the thread_limit clause from the target directive to the teams directive.
        - Changed the comments accordingly.
        - If the num_teams is bigger than OMPVV_NUM_TEAMS_DEVICE, the array out of bounds error can occur; fixed this issue.

seyonglee:
[New results on Summit]
- GCC 13.1.1: 
- XL 16.1.1-10:
    - C test passed but ran on the host with the following warning: The number of teams was unexpected, the test results are likely inconclusive
    - Fortran test passed.
- NVHPC 22.11:
    - C test failed: The number of teams was unexpected, the test results are likely inconclusive; The value of sum_errors is 8.
    - Fortran test passed.
- LLVM 17.0.0: 
    - C test passed.

PR #771, opened 2023-09-06
Add new_test/5.1/metadirective/test_metadirective_nothing.F90

seyonglee:
[Results on Summit]
- GCC 13.1.1:
- XL 16.1.1-10:
    - Fortran test failed: line 46.10: 1515-019 (S) Syntax is incorrect.
- NVHPC 22.11:
    - C test failed: line 34: error: invalid text in pragma
    - Fortran test failed: NVFORTRAN-F-1241-OpenMP begin/end METADIRECTIVE not supported in this version of the compiler.
- LLVM 17.0.0:
    - C test passed but with a warning: extra tokens at the end of '#pragma omp nothing' are ignored [-Wextra-tokens]

### DIFF
--- a/tests/5.1/metadirective/test_metadirective_nothing.F90
+++ b/tests/5.1/metadirective/test_metadirective_nothing.F90
@@ -1,0 +1,102 @@
+!===--- test_metadirective_nothing.F90 -------------------------------------===//
+!
+! OpenMP API Version 5.1 Nov 2020
+!
+! Test for nothing directive within metadirectives. Runs a variety of
+! metadirectives that check if the nothing directive is properly rendered.
+! Primarily tests based on the fact that no matter what 'when' clause is 
+! rendered it should result in nothing, and thus no additional pragma should
+! be created. Thus, the threads should remain unchanged through this process
+! and the compiler should handle it properly. Handles both device and host
+! based tests.
+!
+!//===----------------------------------------------------------------------===//
+#include "ompvv.F90"
+
+#define N 1024
+
+PROGRAM test_metadirective_nothing
+  USE iso_fortran_env
+  USE ompvv_lib
+  USE omp_lib
+  implicit none
+  OMPVV_TEST_OFFLOADING
+
+  IF( omp_get_num_devices() .GT. 0 ) THEN
+    OMPVV_TEST_VERBOSE(metadirectiveOnDevice() .NE. 0)
+  ELSE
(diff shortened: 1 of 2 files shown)


spophale (approved):

andrewkallai:
This test should consider the actual number of threads in a parallel region since omp_in_parallel() will return true only if it is enclosed by an active parallel region, which is executed by a team consisting of more than one thread (I assume this applies to Fortran as well). Otherwise, if the nothing directive is not executed, the parallel for will be executed, but this fact will not be verified with single threaded execution (as omp_in_parallel() returns false).

andrewkallai:
It is true that it is not clear if parallel for was exectued single threaded or the nothing directive was executed. For that reason, I think it makes sense to still add a check for the number of threads to state that the results of the test cannot be verified if the test is running single threaded.
Adding this code after the variable definitions of each function could achieve this:
```
!beginning of function metadirectiveOnTYPE
INTEGER :: num_threads
!$omp parallel
IF(omp_get_thread_num() .EQ. 0) THEN
  num_threads = omp_get_num_threads()
END IF
!$omp end parallel

IF (num_threads .EQ. 1) THEN
  OMPVV_WARNING("test is running single threaded, results cannot be verified")
ELSE
  !code for test here
ENDIF
!return with errors count
```

spophale:
@seyonglee will add checks for omp_get_max_threads and set num_threads for parallel regions.

fel-cab (approved):

seyonglee:
Now, the C test uses `thread-limit-var` ICV and `omp_get_thread_limit()` API, while the Fortran test uses `nthreads-var` ICV  and `omp_get_max_threads()` API; each approach has its limit. We should decide which one to use.

seyonglee:
> Now, the C test uses `thread-limit-var` ICV and `omp_get_thread_limit()` API, while the Fortran test uses `nthreads-var` ICV and `omp_get_max_threads()` API; each approach has its limit. We should decide which one to use.

OpenMP Version 5.1, Section 2.4.4.1 How the Per-Data Environment ICVs Work says the following:

> If a teams construct with a thread_limit clause is encountered, the thread-limit-var ICV
> from the data environment of the initial task for each team is instead set to an implementation
> defined value between one and the value specified in the clause.

This means that the actual value of `thread-limit-var` can be smaller than the value in the `thread_limit` clause, in which case, we may not be able to check whether the `thread_limit` clause is applied or not.

PR #724, opened 2023-04-28
Adding test cases for target construct and applicable clauses

lthakur007:
Please review...

PR #773, opened 2023-09-18
test_target_parallel_linear.c: Fix memory allocation

tob2:
Fix a bug in https://github.com/SOLLVE/sollve_vv/tree/master/tests/5.0/target/test_target_parallel_linear.c — where the undefined THREADS is use instead of the N.

That test is rather newish and was added via Pull Req. #724

@seyonglee, @fel-cab, @spophale, @lthakur007 – please review.

### DIFF
--- a/tests/5.0/target/test_target_parallel_linear.c
+++ b/tests/5.0/target/test_target_parallel_linear.c
@@ -15,7 +15,7 @@
 
 
 int Runtst(int gpu) {
-  int *A = malloc(sizeof(int) * THREADS), errors = 0;
+  int *A = malloc(sizeof(int) * N), errors = 0;
   for (int i = 0; i < N; ++i) {
     A[i] = i;
   }
--- a/tests/5.0/target/test_target_parallel_linear.c
+++ b/tests/5.0/target/test_target_parallel_linear.c
@@ -46,8 +46,7 @@ int main() {
   OMPVV_TEST_OFFLOADING;
   int TotGpus = omp_get_num_devices();
   int errors = 0;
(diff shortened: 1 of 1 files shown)


tob2:
Looking at the testcase, I saw that it only run on non-initial devices - but the test function should also work on the host; hence, it is now tested as well. It also did reset 'errors', but I think it makes sense to have the overall result and not only the one of the last device. → Fixed in the follow-up commit.

fel-cab (approved):

spophale (approved):

tests/5.0/target/test_target_parallel_linear.c
```diff
@@ -46,8 +46,7 @@ int main() {
   OMPVV_TEST_OFFLOADING;
   int TotGpus = omp_get_num_devices();
   int errors = 0;
-  for (int gpu = 0; gpu < TotGpus; ++gpu) {
-    errors = 0;
+  for (int gpu = 0; gpu <= TotGpus; ++gpu) {
```
spophale, line 49: Why <= ?

tob2: No strong reason but as written above:
> Looking at the testcase, I saw that it only run on non-initial devices - but the test function should also work on the host; hence, it is now tested there as well.

If there is no code reason that something only works on non-host devices, I think it can also be run on the host – which is especially useful when one runs it on a system lacking offload devices.

It then at least tests whether `parallel for device(gpu) linear(val:2)` does somewhat sensible on the host (in a host-fallback target region).

PR #774, opened 2023-09-18
Adding Regression and Stress Tests to the V&V

ajarmusch:
We added a regression folder with regression and stress tests. We updated the makefile to include a flag `REGRESSION=1` if you want to run the regression tests. By default `REGRESSION` is assigned to 0 thus omitting all regression tests during the run. 

We tested this change on Frontier and Perlmutter without issues during the run.

### DIFF
--- a/regression/stress_tests/regression_distribute_for_collapse_N_TEAMS_N_THREADS.c
+++ b/regression/stress_tests/regression_distribute_for_collapse_N_TEAMS_N_THREADS.c
@@ -1,0 +1,98 @@
+//===--- regression_distribute_for_collapse_N_TEAMS_N_THREADS.c ----------------------------------------===//
+//
+// OpenMP API Version 4.5
+//
+// Test #pragma omp target teams distribute parallel for collapse() on an 
+// N_DIM*N_DIM*N_DIM*N_DIM array by writing to it. Testing on a gloabal 
+// array with at least 3-4 dimensions in order to do a for sollapse of
+// at least 3-4. To ensure that each write to array occurred only once.
+//
+////===----------------------------------------------------------------------===//
+
+#include <stdio.h>
+#include <omp.h>
+
+#ifndef DIM_SIZE
+#define DIM_SIZE 4
+#endif
+
+#ifndef N_THREADS
+#define N_THREADS 128
+#endif
+
+#ifndef N_TEAMS
+#define N_TEAMS 128
+#endif
+
+#ifndef COLLAPSE
+#define COLLAPSE 4
+#endif
+
+int calc_index(int dim_indices, int i, int j, int k, int l){
+  return dim_indices*dim_indices*dim_indices * i + dim_indices*dim_indices * j + dim_indices * k + l;
+}
+
+int test_collapse() {
+  int errors = 0;
+  int* teams = (int*)malloc(sizeof(int)*DIM_SIZE*DIM_SIZE*DIM_SIZE*DIM_SIZE);
+  int* threads = (int*)malloc(sizeof(int)*DIM_SIZE*DIM_SIZE*DIM_SIZE*DIM_SIZE);
+  int i,j,k,l;
+
+  if (teams == NULL || threads == NULL){
+    printf("malloc failed!\n");
+    return 1;
+  }
+
+  for (i = 0; i < DIM_SIZE*DIM_SIZE*DIM_SIZE*DIM_SIZE; ++i) {
+    teams[i] = -1;
+    threads[i] = -1;
+  }
+#pragma omp target teams distribute parallel for collapse(COLLAPSE) \
+  map(tofrom: teams[:DIM_SIZE*DIM_SIZE*DIM_SIZE*DIM_SIZE], \
+              threads[:DIM_SIZE*DIM_SIZE*DIM_SIZE*DIM_SIZE]) private(i,j,k,l) \
+  num_teams(N_TEAMS) num_threads(N_THREADS)
+
+  for (i = 0; i < DIM_SIZE; ++i) {
+    for (j = 0; j < DIM_SIZE; ++j) {
+      for (k = 0; k < DIM_SIZE; ++k) {
+        for (l = 0; l < DIM_SIZE; ++l) {
+          int index = calc_index(DIM_SIZE, i, j, k, l);
+          if(teams[index] == -1)
+            teams[index] = omp_get_team_num();
+          else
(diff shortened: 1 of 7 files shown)


regression/stress_tests/regression_distribute_for_collapse_N_TEAMS_N_THREADS.c
```diff
@@ -0,0 +1,87 @@
+#include <stdio.h>
```
seyonglee, line 1: Add header comments here.

regression/threaded_data_transfer/regression_large_data_async_transfers.c
```diff
@@ -0,0 +1,126 @@
...
+        if(i < array_size1){
+            host_memory1[i] = i;
+        }
+    }
+
+    /* copy to target */
+    omp_target_memcpy_async(device_memory8, host_memory8, sizeof(double)*array_size8, 0, 0, device, host, 0, NULL);
+
+    #pragma omp target is_device_ptr(device_memory8) device(device)
+
```
seyonglee, line 55: Delete the space between the target directive and the teams directive, which is unnessary and confusing.

regression/threaded_data_transfer/regression_large_data_async_transfers.c
```diff
@@ -0,0 +1,126 @@
...
+        }
+        if(i < array_size1){
+            host_memory1[i] = i;
+        }
+    }
+
+    /* copy to target */
+    omp_target_memcpy_async(device_memory8, host_memory8, sizeof(double)*array_size8, 0, 0, device, host, 0, NULL);
+
+    #pragma omp target is_device_ptr(device_memory8) device(device)
```
seyonglee, line 54: This target task is dependent on the asynchronous `omp_target_memcpy_async()` task; therefore, appropriate synchronization mechanism is needed (either creating an explicit dependency between the two tasks or adding `taskwait` directive before this target task.

regression/threaded_data_transfer/regression_large_data_async_transfers.cpp
```diff
@@ -0,0 +1,126 @@
+//this tests async data transfer between the host and the device using 4 host arrays and 1 device array
```
seyonglee, line 1: This C++ test is identical to the above C test; do we need both C and C++ files? (all the C and C++ file pairs in this directory are identical to each other) 
If the purpose of the C++ file is to test the C++ compiler, we can compile the same C file using the C++ compiler.

regression/threaded_data_transfer/regression_large_data_async_transfers.c
```diff
@@ -0,0 +1,126 @@
...
+
+    #pragma omp teams distribute parallel for num_threads(number_of_threads)
+    for(int i = 0; i < N; i++){
+        device_memory8[i] = i*2; // initialize data on device
+    }
+
+    /* copy to host */
+    omp_target_memcpy_async(host_memory8, device_memory8, sizeof(double)*array_size8, 0, 0, host, device, 0, NULL);
```
seyonglee, line 62: If the above target task becomes asynchronous, the following asynchronous memcpy tasks will depend on the target task.

regression/threaded_data_transfer/regression_qmcpack_target_math.c
```diff
@@ -0,0 +1,51 @@
...
+  // This is intentional
+  int c99_zero = FP_ZERO;
+  
+#pragma omp target map(tofrom: array[0:N]) 
+  for (int i = 0; i < N; ++i) {
+    array[i] = pow((double)i,2.0);
+  }
+
+  for (int i = 0; i < N; ++i) {
+    OMPVV_TEST_AND_SET(errors, (array[i] - pow((double)i,2)) > 0.000009);
```
seyonglee, line 40: Change `pow((double)i,2)` to `pow((double)i,2.0)`

regression/threaded_data_transfer/regression_small_data_async_transfers.c
seyonglee, line 1: This file is identical to `regression/threaded_data_transfer/regression_small_data_async_transfer.c`; delete this.

ajarmusch:
rebase

regression/threaded_data_transfer/regression_large_data_async_transfers.c
```diff
@@ -0,0 +1,156 @@
+//===--- regression_large_data_async_transfers.c ------------------------------===//
+//
+// OpenMP API Version 4.5
```
seyonglee, line 3: `omp_target_memcpy_async()` API is added from OpenMP API Version 5.1,

regression/threaded_data_transfer/regression_large_data_async_transfers.c
```diff
@@ -0,0 +1,156 @@
...
+      host_memory1[i] = i;
+    }
+  }
+
+  /* copy to target */
+  omp_target_memcpy_async(device_memory8, host_memory8,
+                          sizeof(double) * array_size8, 0, 0, device, host, 0,
+                          NULL);
+
+#pragma omp target is_device_ptr(device_memory8) device(device)
```
seyonglee, line 68: This target task is dependent on the `asynchronous omp_target_memcpy_async()` task; therefore, appropriate synchronization mechanism is needed (either creating an explicit dependency between the two tasks or adding `taskwait` directive before this target task.

regression/stress_tests/regression_distribute_for_collapse_N_TEAMS_N_THREADS.c
```diff
@@ -0,0 +1,98 @@
...
+    for (j = DIM_SIZE-1; j < DIM_SIZE; ++j) {
+      for (k = DIM_SIZE-1; k < DIM_SIZE; ++k) {
+        for (l = DIM_SIZE-1; l < DIM_SIZE; ++l) {
+          int index = calc_index(DIM_SIZE, i, j, k, l);
+          printf("%d, %d, %d, %d, %d, %d, %d\n",i,j,k,l,index,teams[index],threads[index]); 
+        }
+      }
+    }
+  }
+
```
fel-cab, line 80:
```suggestion
```
I think this for is not necessary, since it only prints the results. Usually tests only prints warnings and errors.
Furthermore, the array threads doesn't seem to be useful, unless it is important for the error printf in the following for

spophale:
May come back to this later.

PR #779, opened 2023-10-10
Providing option to the user an option to set time limit using SOLLVE_TIMELIMIT env

lthakur007:
Providing option to the user an option to set time limit using SOLLVE_TIMELIMIT env


seyonglee (approved):

spophale (approved):

spophale:
Thanks @lthakur007 !

PR #783, opened 2023-11-15
Adding test case for target teams distribute parallel for default con…

lthakur007:
Please review...


ronlieb (approved):

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,83 @@
...
+  int CONST = 123;
+  int Arr[32];
+  for (int i = 0; i < 32; ++i) {
+    Arr[i] = i;
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(private)
+  for (int i = 0; i < 32; ++i) {
+    CONST += 10;
```
spophale, line 44: since CONST is privatized, its value is undefined (could be garbage).

spophale, line 44:
```suggestion
    CONST = 10;
```

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,83 @@
...
+  int CONST = 123;
+  int Arr[32];
+  for (int i = 0; i < 32; ++i) {
+    Arr[i] = i;
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(shared)
+  for (int i = 0; i < 32; ++i) {
+    CONST += 10;
```
spophale, line 66: Updates need to be protected by atomic.

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,83 @@
...
+  int Arr[32];
+  for (int i = 0; i < 32; ++i) {
+    Arr[i] = i;
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(private)
+  for (int i = 0; i < 32; ++i) {
+    CONST += 10;
+    Arr[i] += CONST;
```
spophale, line 45: Same would be true for Arr.

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,83 @@
...
+  for (int i = 0; i < 32; ++i) {
+    Arr[i] = i;
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(firstprivate)
+  for (int i = 0; i < 32; ++i) {
+    if (Arr[i] != i) {
+#pragma omp atomic
+      ErrCount += 1;
```
spophale, line 27: ErrCount should be both shared and mapped tofrom to get the correct values on the host.

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(private)
+  for (int i = 0; i < 32; ++i) {
+    CONST += 10;
+    Arr[i] += CONST;
+    if (Arr[i] != (i + 10)) {
+#pragma omp atomic
+      ErrCount += 1;
```
spophale, line 48: ErrCount has to be shared and mapped tofrom.

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(shared)
+  for (int i = 0; i < 32; ++i) {
+    CONST += 10;
+    Arr[i] += CONST;
+    if (Arr[i] != (i + 10 + 123)) {
+#pragma omp atomic
+      ErrCount += 1;
```
spophale, line 70: same as above.

spophale:
@lthakur007 , this appears to be a OpenMP 5.1 test as before that only shared or none was allowed with default clause for C/C++.
According to 5.1 the default clause applies to both teams and parallel. The test logic needs to be checked for such usage.

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
lthakur007: Hi @spophale ,
I observed that, when we use private for any variable, Inside each threads its variable is '0'. I thought I can use this to test "default(private)". Should I avoid assuming the values would be zero ?

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
lthakur007: Added #pragma omp atomic....

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
lthakur007: mapped ErrCount tofrom using map()

lthakur007:
I have moved the test to 5.1 folder. Will observe CI machines report with 5.1 version of this test...

lthakur007:
HI @spophale ,
I have addressed the review comments, please consider the changes...

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
seyonglee: OpenMP Version 4.5, Section 2.15.3.3 private Clause, page 194, line 6-7 say the following:

> The new list item is initialized, or has an undefined initial value, as if it had been locally declared without an initializer.

The C11 standard says that non-static objects at block scope declared without explicit initializers are left uninitialized.
Therefore, we should avoid assuming the values would be zero.

seyonglee:
Please delete the obsolete test:    tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,89 @@
...
+  int Arr[32];
+  for (int i = 0; i < 32; ++i) {
+    Arr[i] = i;
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(private)
+  for (int i = 0; i < 32; ++i) {
+    CONST = 10;
+    Arr[i] += CONST;
```
seyonglee, line 49: `Arr` should be explicitly mapped (e.g., `map(to: Arr)`).

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,89 @@
...
+  for (int i = 0; i < 32; ++i) {
+    if (Arr[i] != i) {
+#pragma omp atomic
+      ErrCount += 1;
+    }
+  }
+  return ErrCount;
+}
+
+int DefaultPrivate() {
```
seyonglee, line 37: This test does not actually check the `default(private)` behavior; one way to check the behavior is to check the value of `CONST` after the target region, which should not have been modified.

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,89 @@
...
+  int Arr[32];
+  for (int i = 0; i < 32; ++i) {
+    Arr[i] = i;
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(shared) map(tofrom: ErrCount)
+  for (int i = 0; i < 32; ++i) {
+#pragma omp atomic
+    CONST += 10;
```
seyonglee, line 71: `default(shared)` clause will be applied to both `teams` and `parallel` constructs. Therefore, `CONST` will be shared in both the teams region and parallel region.
    ==> `CONST` in line 71 will be updated multiple times.
    ==> We don't know the value of `CONST` in line 73, and thus we cannot check the value of `Arr[i]` in line 74.

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
spophale: @lthakur007 yes, don't assume value is zero.

lthakur007:
> Please delete the obsolete test: tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c

Done!! Thank you @seyonglee  :) ..

tests/4.5/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
lthakur007: Hi @spophale ,
Thank you for pointing out. I did the required changes...

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
lthakur007: Hi @seyonglee ,
Thank you for pointing out, mapped the variables appropriately.
Please review the changes...

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
lthakur007: Hi @seyonglee ,
Added check for CONST for its original value as suggested...
Thank you!!

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
lthakur007: Hi @seyonglee ,
Made changes such that CONST's usage is limited to read-only.
        default(shared) map(tofrom: ErrCount)
  for (int i = 0; i < 32; ++i) {
#pragma omp atomic
    Arr[i] += CONST;
    if (Arr[i] != (i + 123)) {
#pragma omp atomic
      ErrCount += 1;
    }
  }

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,89 @@
+//-----------------test_target_teams_distribute_parallel_for_default.c-----
+//
+// OpenMP API Version 5.1 Mar 2023
+//
+// This test intends to test the functionality of an indirect
```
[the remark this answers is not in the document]
lthakur007: Done!!!

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,89 @@
+//-----------------test_target_teams_distribute_parallel_for_default.c-----
+//
+// OpenMP API Version 5.1 Mar 2023
```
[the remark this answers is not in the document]
lthakur007: Done!!!

lthakur007:
Hi @spophale ,
Addressed all the review comments.. Please consider the updates ...

tests/5.1/target_teams_distribute_parallel_for/test_target_teams_distribute_parallel_for_default.c
```diff
@@ -0,0 +1,88 @@
...
+  int CONST = 123;
+  int Arr[32];
+  for (int i = 0; i < 32; ++i) {
+    Arr[i] = i;
+  }
+
+#pragma omp target teams distribute parallel for num_teams(2) thread_limit(10)\
+        default(shared) map(tofrom: ErrCount)
+  for (int i = 0; i < 32; ++i) {
+#pragma omp atomic
```
spophale, line 71: atomic is not needed.

lthakur007: Removed "atomic" construct...

lthakur007:
Hi @spophale ,
Please consider the changes submitted....

fel-cab:
It is not clear to me how this test is testing the `default` clause, since the test would pass if the `default` clauses are removed, or they are changed for the other possible `default` (e.g. if the `default(private)` is changed for `default(shared)` or the other way around). Is it possible to change the test in a way that the consequence of `default` is carefully tested?

lthakur007:
> It is not clear to me how this test is testing the `default` clause, since the test would pass if the `default` clauses are removed, or they are changed for the other possible `default` (e.g. if the `default(private)` is changed for `default(shared)` or the other way around). Is it possible to change the test in a way that the consequence of `default` is carefully tested?

Hi @spophale ,
I agree with the comment of @fel-cab mentioned above. However, the test passes with/without "default(firstprivate)" and also with/without presence of "map(to: Arr[0:32])". I have attached a file for reference. The terminal output when the file is run is as follows:
@spophale , could you please help me improve the test case.....

_/opt/rocm/llvm/bin/clang -O2 -fopenmp --offload-arch=gfx906 -fopenmp-version=51 dist_paralel_default.c
 ./a.out
Test Passed!!
Test Passed!!
Test Passed!!
Test Passed!!_
__
[dist_paralel_default.txt](https://github.com/SOLLVE/sollve_vv/files/14163367/dist_paralel_default.txt)

seyonglee:
I also agree that this test does not actually test the `default` clause, since the `default` clause is not applied to the `target` construct (and this test checks the data mapping behaviors only.)

spophale:
@lthakur007 , here is my suggestion:
For default(firstprivate)
- have a scalar initialized to some value outside the target region
- since firstprivate applies to both teams as well as parallel; update this scalar variable say 'CONST += team_num + thread_num'
- Update Arr (should be mapped tofrom)
- Check updates Arr on host


Same steps can be used for default(private) but CONST will be assigned inside the target region -- 'CONST = team_num + thread_num'

spophale (approved):

spophale:
This test was merged with PR #793 .This is a duplicate.

PR #788, opened 2023-12-20
Flipped values for USM macro var, now default value is zero to indica…

andrewkallai:
…te no USM present

### DIFF
--- a/ompvv/ompvv.h
+++ b/ompvv/ompvv.h
@@ -98,24 +98,24 @@ _Pragma("omp target map (from: _ompvv_isOffloadingOn)") \
 
 // Macro to check if it is a shared data environment
 #define OMPVV_TEST_SHARED_ENVIRONMENT_PROBE \
-  int _ompvv_isSharedEnv = 1; \
+  int _ompvv_isSharedEnv = 0; \
   _ompvv_isOffloadingOn = 0; \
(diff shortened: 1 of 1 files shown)


fel-cab (approved):

fel-cab:
@andrewkallai is going to add the tests that rely on the macro before we approve it.

andrewkallai:
All C tests which rely on the macro have been reviewed.

fel-cab:
@andrewkallai are you going to add, to this commit, the C tests that rely this macro and have to be changed?

andrewkallai:
No changes are needed so no tests need to be added.

spophale:
@andrewkallai , since the behavior is reversed, will the tests not need updating as well ?

andrewkallai:
Case 1: existing tests make use of OMPVV_TEST_SHARED_ENVIRONMENT;
Because this macro is used for reporting purposes and has no effect on the program besides defining internal macro variables, no change is needed as long as the program does not depend on those internal variables (which is the actual case).
Case 2: existing tests make use of OMPVV_TEST_AND_SET_SHARED_ENVIRONMENT(var2set);
This can affect program behavior if var2set is used in the program. Thus, if the program relies on the previous definition logic, the test has to be changed. I found this to not be the case for existing tests, so no changes have to be made.
Note: I have not included the Fortran tests into consideration, which I will change in a seperate PR.

spophale (approved):

PR #789, opened 2023-12-22
test_dispatch_is_device_ptr.c: Avoid accessing device var on the host

tob2:
_The main change is to avoid `arr[0] == 2` on the host, i.e. accessing a device-allocated variable on the host. The rest is only polishing:_
 
tests/5.1/dispatch/test_dispatch_is_device_ptr.c:
* Don't access a device variable on the host
* Free the device-allocated memory
* Add a comment with regard to the expected value

Also, the `is_device_ptr(arr)` is required because:

1. in case of _not_ having 'unified_address' and

2. as OpenMP 5.1 states in '2.21.7.2 Pointer Initialization for Device Data Environments' that if the pointer does not point to associated storage, it is set to _NULL_. With OpenMP 5.2, that changed to: "the pointer retains its original value"; hence, since 5.2 it would work here, if (1) is true [which is commonly the case].

@spophale, @andrewkallai @nolanbaker31 @seyonglee  – please review!

### DIFF
--- a/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
+++ b/tests/5.1/dispatch/test_dispatch_is_device_ptr.c
@@ -42,6 +42,7 @@ void add_dev(int *arr){
 
 int test_wrapper() { 
     int t;
+    int called_add = 0;
     errors = 0;
     t = omp_get_default_device();
     arr = (int *)omp_target_alloc( sizeof(int)*N, t);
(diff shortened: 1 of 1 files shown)


tob2:
_I also want to give kudos to a colleague who found the issue and diagnosed the main/original issue. With this fix applied, this testcase now passes with a GCC that includes the 'dispatch'-support feature (which the colleague developed)._

spophale:
Makes sense to me. @tob2 thanks to you and your colleague  for the fix.

fel-cab (approved):

PR #794, opened 2024-02-12
Moving omp_get_max_teams outside teams.

### LINKED ISSUE #792: 5.1/teams/test_target_set_num_teams.c: error: OpenMP runtime API call ‘omp_get_max_teams’ strictly nested in a ‘teams’ region
Pull Request #728 added **5.1/teams/test_target_set_num_teams.c** – which fails to compile with GCC mainline/14 with:

```C
tests/5.1/teams/test_target_set_num_teams.c: In function ‘main’:
tests/5.1/teams/test_target_set_num_teams.c:28:37: error: OpenMP runtime API call ‘omp_get_max_teams’ strictly nested in a ‘teams’ region
   28 |                         num_teams = omp_get_max_teams();
      |                                     ^~~~~~~~~~~~~~~~~~~
```
 
https://github.com/SOLLVE/sollve_vv/blob/802c0a8b3e3ae08d0691acd55e40c940aea57768/tests/5.1/teams/test_target_set_num_teams.c#L25-L30

As the compiler notes, the code violates the following as it does not list **omp_get_max_teams** as being permitted:

> Restrictions to the **teams** construct are as follows:
> ...
> * **distribute**, **distribute simd**, distribute parallel worksharing-loop, distribute parallel
worksharing-loop SIMD, **parallel** regions, including any **parallel* regions arising from
combined constructs, **omp_get_num_teams()** regions, and **omp_get_team_num()**
regions are the only OpenMP regions that may be strictly nested inside the teams region.

→ OpenMP 5.1, "2.7 teams Construct", [103:30–33], https://www.openmp.org/spec-html/5.1/openmpse15.html#x62-620002.7

@jrreap @andrewkallai @spophale @fel-cab

spophale:
Closes #792

### DIFF
--- a/tests/5.1/teams/test_target_set_num_teams.c
+++ b/tests/5.1/teams/test_target_set_num_teams.c
@@ -27,14 +27,25 @@ int main() {
 	OMPVV_ERROR_IF(num_teams != 8, "Upper bound of the number of teams is not the number set by omp_set_num_teams()");
 	OMPVV_TEST_AND_SET(errors, num_teams != 8);
 
+	#pragma omp teams 
+	{
+		if (omp_get_team_num() == 0 ) {
+			num_teams = omp_get_num_teams();
+		}
+	}
+	OMPVV_ERROR_IF(num_teams > 8, "Incorrect number of teams detected when num_teams clause was not specified");	
+	OMPVV_TEST_AND_SET(errors, num_teams > 8);
+
 	#pragma omp teams num_teams(OMPVV_NUM_TEAMS_HOST)
 	{
 		if (omp_get_team_num() == 0 ) {
 			num_teams = omp_get_num_teams();
 		}
 	}
+	OMPVV_ERROR_IF(num_teams != OMPVV_NUM_TEAMS_HOST, "The number of teams was not overriden by the num_teams clause");	
+	OMPVV_TEST_AND_SET(errors, num_teams != OMPVV_NUM_TEAMS_HOST);
 
-	OMPVV_ERROR_IF(num_teams <= OMPVV_NUM_TEAMS_HOST, "The number of teams was not overriden by the num_teams clause");	
-	OMPVV_TEST_AND_SET(errors, num_teams <= OMPVV_NUM_TEAMS_HOST);
+	
+	
 	OMPVV_REPORT_AND_RETURN(errors);
 }
--- a/tests/5.1/teams/test_target_set_num_teams.c
+++ b/tests/5.1/teams/test_target_set_num_teams.c
@@ -22,12 +22,7 @@ int main() {
 	int num_teams = 0;
 	omp_set_num_teams(8);
         	
-	#pragma omp teams 
-	{
-		if (omp_get_team_num() == 0 ) {
-			num_teams = omp_get_max_teams();
-		}
-	}               
+	num_teams = omp_get_max_teams();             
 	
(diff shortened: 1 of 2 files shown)


tob2 (approved):
LGTM.

From the spec:
> The value returned by **omp_get_max_teams** is the value of the _nteams-var_ ICV of the current device. This value is also an upper bound on the number of teams that can be created by a teams construct without a **num_teams** clause that is encountered after execution returns from this routine.

Thus, this ICV value should be the same whether being inside some explicit region or not. → Hence, this change is fine. _(Besides fixing Issue #792, which is the whole point of this PR.)_

I wonder whether there should be an additional test of a **teams** without a **num_teams** clause, testing whether then omp_get_num_teams() <= 8 or not, i.e. similar to the test further below in the source file but without the clause `num_teams(OMPVV_NUM_TEAMS_HOST)` and checking for `<= 8` instead of `<= OMPVV_NUM_TEAMS_HOST`.

_(Side remark: OMPVV_NUM_TEAMS_HOST is 4)_

tob2:
Can you additionally change the following?

_I get num_teams == 4 == OMPVV_NUM_TEAMS_HOST, which is the expected result. But sollve_vv prints an error. → Thus, the condition should not be an assert/expected good value condition but an error condition._

Namely: `<=`  should be changed to `>` — That's for the check at the very bottom of the file for the **omp teams** – and not for the check directly after your change.

```diff
@@ -42,2 +43,2 @@ int main() {
-       OMPVV_ERROR_IF(num_teams <= OMPVV_NUM_TEAMS_HOST, "The number of teams was not overriden by the num_teams clause");     
-       OMPVV_TEST_AND_SET(errors, num_teams <= OMPVV_NUM_TEAMS_HOST);
+       OMPVV_ERROR_IF(num_teams > OMPVV_NUM_TEAMS_HOST, "The number of teams was not overriden by the num_teams clause");      
+       OMPVV_TEST_AND_SET(errors, num_teams > OMPVV_NUM_TEAMS_HOST);
```

Thanks for this pull request and all your previous patch reviews!

spophale:
@tob2 On further inspection, I think the error conditions should be:
```
OMPVV_ERROR_IF(num_teams != OMPVV_NUM_TEAMS_HOST, "The number of teams was not overriden by the num_teams clause");     
OMPVV_TEST_AND_SET(errors, num_teams != OMPVV_NUM_TEAMS_HOST);

```
Since both the upper and lower bounds for num_teams will evaluate to OMPVV_NUM_TEAMS_HOST.

tob2:
> @tob2 On further inspection, I think the error conditions should be: [...]

I concur that the stricter condition is fine as well – and, hence, can/should be used.

Can you change this as well? _(And, possibly, duplicate the bottom part of the test without the **num_teams** clause as additional test? For that one, the condition would then be indeed `OMPVV_ERROR_IF(num_teams > 8, …`)_


_In 5.1, the wording is makes clear that the used `num_teams(OMPVV_NUM_TEAMS_HOST)` sets also the lower bound:_
> If the **num_teams** clause is present, _lower-bound_ is the specified lower bound and _upper-bound_ is the specified upper bound on the number of teams requested. If a lower bound is not specified, the lower bound is set to the specified upper bound. The number of teams created is implementation defined, but it will be greater than or equal to the lower bound and less than or equal to the upper bound.

_For completeness: In 5.0, `>` would be the correct condition as 5.0 only allows to set an upper bounds (but this is a 5.1 test case):_
> The number of teams created is implementation defined, but is less than or equal to the value specified in the **num_teams** clause.

seyonglee (approved):

Issue #795, opened 2024-02-29
Missing taskwait in test_target_memcpy_async_no_obj.c


brianwhitney59:
There appears to be a missing $!omp taskwait after the first omp_target_memcpy_async call in test_target_memcpy_async_no_obj.c (5.1 test)

```
40     /* copy to target */
 41     omp_target_memcpy_async(mem_dev_cpy, mem, sizeof(double)*N,
 42                                 0,          0,
 43                                 t,          h,
 44                                 0,          NULL);
 45 
 46     #pragma omp taskwait   //<<<<<<<<<<<<<< this is not there and it is believed it is needed to allow the async to finish
 47     #pragma omp target is_device_ptr(mem_dev_cpy) device(t)
 48     #pragma omp teams distribute parallel for
 49     for(i = 0; i < N; i++){
 50         mem_dev_cpy[i] = i*2; // initialize data on device
```

This is using the Intel icpx compiler (internal version 2024.02.28) on Intel PVC GPUs.

The similar test test_target_memcpy_rect_async_obj_no.c includes the taskwait.

fel-cab:
The fixed was merged in the Pull request #723

PR #797, opened 2024-03-01
Fixing bug of test_target_memcpy_async_no_obj.c

fel-cab:
Fixing bug reported on issue #795

Adding `#pragma omp taskwait` after `omp_target_memcpy_async`


spophale (approved):

seyonglee:
We don't need this PR since the `task wait` directive is already applied in PR #723.
Moreover, PR #723  has a later version of the `test_target_memcpy_async_no_obj.c` than this PR.

PR #799, opened 2024-03-22
Taskloop updated

shivapdr:


tests/4.5/taskloop/test_taskloop_private.c
```diff
@@ -0,0 +1,52 @@
...
+#include "ompvv.h"
+
+#define NUM_THREADS 100
+#define NUM_TASKS 1000
+
+int test_taskloop_private() {
+
+  int errors = 0;
+
+  int private_var_sum;
```
tob2, line 22:
```suggestion
  int private_var_sum = 10;
```
... is required to initialize the variable – and it must be initialized (to that value) as there is later the check:
`OMPVV_TEST_AND_SET_VERBOSE(errors, val_not_in_openmp != private_var_sum);`

tests/4.5/taskloop/test_taskloop_private.c
```diff
@@ -0,0 +1,52 @@
...
+  int private_var_sum;  
+  int val_not_in_openmp = 10; 
+  
+   #pragma omp parallel num_threads(NUM_THREADS)
+   {
+      #pragma omp single
+      {
+	#pragma omp taskloop private(private_var_sum) 
+        for(int i = 0; i < NUM_TASKS; i++) {
+	  private_var_sum = i;
```
tob2, line 31: _I want to note for completeness:_
_(a) there is a minor race as the writing does not happen atomically ('#pragma omp atomic write'), but it shouldn't really matter._
_(b) With really bad luck, the value that remains at the end can be still 10 by chance, but that's also unlikely and would also only mean spurious PASS with a_ broken _implementation._
_(c) The name implies some summing up like `private_var_sum += i` — but we cannot change it to that value as the variable starts as being (locally) undefined due to the `private`._

Thus, the simplest is to **leave** the code **as is**  (or add the atomic write if someone feels that that makes more sense).

tests/4.5/taskloop/test_taskloop_num_tasks.c
```diff
@@ -0,0 +1,76 @@
...
+// It allows for efficient parallel execution of loop iterations by distributing them among multiple threads. 
+// The 'num_tasks' clause variable ensures that the loop iterations are shared amoung created 'num_tasks'. 
+// i.e the loop iterations are divided amoung number of tasks (each num_task group is run by same the thread).   
+//----------------------------------------------------------------------------------------------------------//
+
+#include <stdio.h>
+#include <omp.h>
+#include "ompvv.h"
+
+#define NUM_THREADS 100
```
spophale, line 16: Why do we need 100 threads ?

tests/4.5/taskloop/test_taskloop_num_tasks.c
```diff
@@ -0,0 +1,76 @@
...
+	    #pragma omp atomic 
+	    var = var + i;
+
+            thread_ids[i] = omp_get_thread_num();
+        }
+      }
+   }
+
+   //if all the tasks in a group are run by a same thread, get TRUE else FALSE
+   OMPVV_TEST_AND_SET_VERBOSE(errors, (isGroupIdsSame(thread_ids) != 1));
```
spophale, line 61: We need to check if the num_threads actually executing the parallel region > 1.
If there is only one thread doing all the work, it is not an error. Crappy implementation, but still not an error.

tests/4.5/taskloop/test_taskloop_private.c
```diff
@@ -0,0 +1,58 @@
...
+      num_threads = omp_get_num_threads();
+      #pragma omp single
+      {
+	#pragma omp taskloop private(private_var) 
+        for(int i = 0; i < NUM_TASKS; i++) {
+	  private_var = omp_get_thread_num();
+          #pragma omp atomic
+	    shared_var_sum += private_var;
+	 }      
+      }
```
fel-cab, line 37:
```suggestion
      #pragma omp single
      {
	#pragma omp taskloop private(private_var) 
        for(int i = 0; i < NUM_TASKS; i++) {
          int sum = 0;
	  for(private_var = 0; private_var < 5; private_var++)
	    sum += private_var;
          #pragma omp atomic
	    shared_var_sum += sum;
	 }      
      }	   
```
This way we would use `private_var` to add `sum` to `10 (1+2+3+4)`. Each task would do the same. And below we can check that the each taskloop iteration did the increment using the `private_var`. Below we can test that the end value of `shared_var_sum` was `NUM_TASKS * 10`.

tests/4.5/taskloop/test_taskloop_private.c
```diff
@@ -0,0 +1,58 @@
...
+          #pragma omp atomic
+	    shared_var_sum += private_var;
+	 }      
+      }	   
+   }
+  
+   //check if the private variable is same before and after openmp region. i.e it should be same
+   OMPVV_TEST_AND_SET_VERBOSE(errors, private_var != -2);
+   if(num_threads == 100){
+     OMPVV_TEST_AND_SET_VERBOSE(errors, shared_var_sum != NUM_THREADS * ((NUM_THREADS + 1)/2) );
```
fel-cab, line 43:
```suggestion
     OMPVV_TEST_AND_SET_VERBOSE(errors, shared_var_sum != (NUM_TASKS * 10) ) );
```
As I mentioned above, we have to change the check.

tests/4.5/taskloop/test_taskloop_num_tasks.c
shivapdr: Just for testing purpose considered 100 threads.

tests/4.5/taskloop/test_taskloop_private.c
shivapdr: @fel-cab: updated with your comments.

spophale (approved):

fel-cab (approved):

tests/4.5/taskloop/test_taskloop_num_tasks.c
```diff
@@ -59,11 +59,14 @@ int test_taskloop_num_tasks() {
...
    }
 
-   //if all the tasks in a group are run by a same thread, get TRUE else FALSE
-   OMPVV_TEST_AND_SET_VERBOSE(errors, (isGroupIdsSame(thread_ids) != 1));
+   
 
    //To check if the num_threads actually executing the parallel region > 1
-   OMPVV_TEST_AND_SET_VERBOSE(errors, (num_threads_in_parallel_zone < 1));
+   if(num_threads == 1) 
+     OMPVV_INFOMSG("Only a single thread executed the parallel region");
```
mhalk: Hi there,
FYI when looking at our CI, I see `test_taskloop_num_tasks.c` test fails compilation (with AOMP as well as with vanilla LLVM-main):
```
sollve_vv/tests/4.5/taskloop/test_taskloop_num_tasks.c:67:4: error: expected expression
   67 |    else
      |    ^
```
When enclosing `OMPVV_INFOMSG(...)` with braces: `{ OMPVV_INFOMSG(...) }` it compiles and passes.
If you don't mind could you look into that?
Otherwise, please let me know and I'll investigate further from our end.

fel-cab: You are right, I just created a PR #803 to solve the problem. Thanks

PR #800, opened 2024-03-27
Change proposal for test_capture_omp_affinity.c

fel-cab:
The original test FAILS sometimes in some systems. 
Furthermore, I think the original test does not checks the affinity correctly, it only limits to check the size of the format.
With this change I expect to make the test more reliable and check the format.

### DIFF
--- a/tests/5.0/program_control/test_capture_omp_affinity.c
+++ b/tests/5.0/program_control/test_capture_omp_affinity.c
@@ -37,6 +37,11 @@ int main() {
    #pragma omp parallel private(thrd_num) num_threads(threads) 
    {
       thrd_num = omp_get_thread_num();
+      if(thrd_num > threads) { //safety: dont exceed # of buffers
+        OMPVV_ERROR("Number of thread greater than requested");
+        exit(1); 
+      }
+
       if(thrd_num == 0)
         num_threads = omp_get_num_threads();
 
--- a/tests/5.0/program_control/test_capture_omp_affinity.c
+++ b/tests/5.0/program_control/test_capture_omp_affinity.c
@@ -20,36 +20,39 @@
 
 int main() {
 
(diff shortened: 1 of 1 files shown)


tests/5.0/program_control/test_capture_omp_affinity.c
```diff
@@ -20,36 +20,39 @@
...
-   max_req_store = 0;
 
-   #pragma omp parallel private(thrd_num,nchars) reduction(max:max_req_store)
+   #pragma omp parallel private(thrd_num) num_threads(threads) 
    {
-      if(omp_get_num_threads()>n) exit(1); //safety: dont exceed # of buffers
-
-      thrd_num=omp_get_thread_num();
-      nchars=omp_capture_affinity(buffer[thrd_num],(size_t)BUFFER_STORE,NULL);
+      if(omp_get_num_threads() > threads) exit(1); //safety: dont exceed # of buffers
```
spophale, line 39: Add a msg here before exiting.

spophale (approved):

tests/5.0/program_control/test_capture_omp_affinity.c
fel-cab: I removed this line, because it makes no sense to check if the number of threads is greater than the requested, since they have to be less or equal to the num_threads requested.
Instead, I added a variable that checks for the number of threads and only verifies on this number.

spophale: Exactly, if the number of threads is greater than the requested, then it is an error. I would keep it.

fel-cab: Updated the test with the check and the message

seyonglee (approved):

PR #802, opened 2024-04-11
Add uses_allocators to fix compiler error

fel-cab:
LLVM error: 
`error: allocator must be specified in the 'uses_allocators' clause`
adding `uses_allocators(omp_default_mem_alloc)` fixes the problem

### DIFF
--- a/tests/5.0/target/test_target_allocate.c
+++ b/tests/5.0/target/test_target_allocate.c
@@ -24,7 +24,7 @@ int test_target_allocate() {
       host_result += i;
    }
    
-   #pragma omp target allocate(omp_default_mem_alloc:x) firstprivate(x) map(from: device_result)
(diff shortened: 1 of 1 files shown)


tob2 (approved):
LGTM

OpenMP (here TR12) has in “Restrictions to the **target** construct are as follows:”
> Memory allocators that do not appear in a **uses_allocators** clause cannot appear as an allocator in an **allocate** clause or be used in the **target** region unless a **requires** directive with the **dynamic_allocators** clause is present in the same compilation unit.

It seems to be somewhat superfluous to require it for **allocate** clause _on_ **target** clause itself, especially for predefined allocators, but the wording is rather unambiguous.

spophale (approved):

seyonglee (approved):

PR #803, opened 2024-04-12
Fix issue created by mhalk on github

fel-cab:
Fixing compilation error of tests/4.5/taskloop/test_taskloop_num_tasks.c from PR #799

### DIFF
--- a/tests/4.5/taskloop/test_taskloop_num_tasks.c
+++ b/tests/4.5/taskloop/test_taskloop_num_tasks.c
@@ -51,10 +51,10 @@ int test_taskloop_num_tasks() {
         #pragma omp taskloop num_tasks(NUM_TASKS)
         for(int i = 0; i < NUM_ITERATIONS; i++)
         {
-	    #pragma omp atomic 
-	    var = var + i;
+	        #pragma omp atomic 
+	        var = var + i;
 
-            thread_ids[i] = omp_get_thread_num();
+          thread_ids[i] = omp_get_thread_num();
         }
       }
    }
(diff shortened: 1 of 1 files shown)


mhalk:
Thanks for the swift reaction -- much appreciated!

tob2 (approved):
For what it is worth: LGTM.

The issue is that OMPVV_WARNING expands to a `{…}` – which is then followed by a `;` which es a second (empty) statement – such that the `else` has no associated `if`.

* * *
_An alternative would be to use `do { … } while (false)` which always requires a `;` at the end. — Looks uglier in the macro definition  but avoids this issue when using the macro._

_Namely – example for this macro, only:_
```diff
--- a/ompvv/ompvv.h
+++ b/ompvv/ompvv.h
@@ -15,4 +15,4 @@ int _ompvv_isOffloadingOn = -1;
 #ifdef VERBOSE_MODE
-  #define OMPVV_WARNING(message, ...) { \
+  #define OMPVV_WARNING(message, ...) do { \
     printf("[OMPVV_WARNING: %s:%i] " message "\n", __FILENAME__, __LINE__, ##__VA_ARGS__); \
-  }
+  } while (0)
@@ -42,2 +42,2 @@ int _ompvv_isOffloadingOn = -1;
 #else
-  #define OMPVV_WARNING(message, ...) {}
+  #define OMPVV_WARNING(message, ...) do {} while (0)
```

spophale (approved):

seyonglee (approved):

PR #804, opened 2024-04-12
Allow not expected result for const allocator

fel-cab:
The test tests/5.0/target/test_target_uses_allocators_const.c
was checking that  writing to `x` was successful even though the specification says that
the `omp_const_mem_space`:
Represents storage optimized for variables with constant values. The result of writing to this storage is unspecified.
Which means that if the variable `x` is not written, the result is also correct.
This PR seeks to solve this problem.

### DIFF
--- a/tests/5.0/target/test_target_uses_allocators_const.c
+++ b/tests/5.0/target/test_target_uses_allocators_const.c
@@ -39,7 +39,8 @@ int test_uses_allocators_const() {
   device_result = x;
 }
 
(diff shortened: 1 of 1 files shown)


seyonglee:
Is this test legal? For me, it does not make sense to evaluate "unspecified behavior". (We cannot say whether this test is passed or failed in any case.)

spophale (approved):

seyonglee (approved):

Issue #805, opened 2024-04-16
test_taskloop_num_tasks.c failing to compile


brianwhitney59:
test_taskloop_num_tasks.c is failing to compile.  There seems to be an unneeded ";" that is causing the if structure to fail

 64    //To check if the num_threads actually executing the parallel region > 1
 65    if(num_threads == 1)
 66      OMPVV_WARNING("Only a single thread executed the parallel region");
 67    else
 68      //if all the tasks in a group are run by a same thread, get TRUE else F    ALSE
 69      OMPVV_TEST_AND_SET_VERBOSE(errors, (isGroupIdsSame(thread_ids) != 1)); 

On line 66, the macro expands and causes multiple lines, so the else statement is left orphaned.  Removing the ";" or adding {} around the macro allow it to build.

This was found with the intel 2024.1 compiler.
icx -std=c99 -fiopenmp -fopenmp-targets=spir64 -I./ompvv -DVERBOSE_MODE=1 test_taskloop_num_tasks.c
test_taskloop_num_tasks.c:67:4: error: expected expression
   67 |    else 
      |    ^
1 error generated.

fel-cab:
The issue was solved with PR #803

PR #807, opened 2024-04-17
Fixing test_taskloop_num_tasks.c issue due to missing brackets on if/…

fel-cab:
Fixing test_taskloop_num_tasks.c issue due to missing brackets on if/else statement from issue #805


tob2 (approved):
LGTM.

This seems to be the very same as PR #803 – except that the title of the pull request / commit log is more helpful.

And there is now a spurious tailing space after '{' in the 'if' line.

Regarding the background, copying part of my comment from #803 (w/ a fixed typo):
> The issue is that OMPVV_WARNING expands to a {…} – which is then followed by a ';' (semicolon) which is a second (empty) statement – such that the else has no associated if.

PR #819, opened 2024-05-14
Break possible deadlock on lockstep execution

fel-cab:
Some of the target_requires_atomic tests get on infinite loop with LLVM/Clang compiler because lockstep execution on the target device and the `if-else` order of execution. These tests are written assuming that target threads are independent or that if they execute in a warp, in lockstep execution, the order of execution would be first the `if` part and then the `else` part.
The modifications on the tests are meant to make sure that this problem does not happens for LLVM/Clang or any compiler that could encounter this problem.

### DIFF
--- a/tests/5.0/target_requires/test_target_requires_atomic_default_mem_order_acq_rel.c
+++ b/tests/5.0/target_requires/test_target_requires_atomic_default_mem_order_acq_rel.c
@@ -15,8 +15,6 @@
 #include <stdlib.h>
 #include "ompvv.h"
 
-#define N 1024
-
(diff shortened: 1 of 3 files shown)


fel-cab:
The following Issue was created for LLVM issues https://github.com/llvm/llvm-project/issues/94047
Given the conclusion by the developers. I think this PR should be closed, but the problem with LLVM will always remain. The original code doesn't pass on LLVM either.

`This code cannot reliably work. You are trying to synchronize within a warp on an AMDGPU, that does not work. OpenMP 6.0 makes this explicit. If you want to test atomics, split the reader and writer into different warps at least.`

spophale:
Closing this for now.

PR #820, opened 2024-05-22
Resolving issue 817.

### LINKED ISSUE #817: tests/5.0/target_update/test_target_update_mapper_from_discontiguous should use map clause for target region
The problem I'm seeing is a seg fault on the AMDGPU. I think it is happening because the target region is referencing a pointer (as if it's an array) without providing any mapping information.

Here's a small snippet from the test case to illustrate the issue:

#define N 32
typedef struct{
  size_t len;
  double *data;
} T;

int main() {
  T s;
  s.len = N;
  s.data = (double *)calloc(N,sizeof(double));

  #pragma omp target
  for (int i = 0; i < s.len; i++)
    s.data[i] = i; **<<<<<** seg fault occurs here
}

The reference to "s.data", a pointer, according to the latest spec (OpenMP spec TR12, pg. 6, L16-18) is supposed to be implicitly mapped like this:

  "A variable that is of type pointer, but is neither a pointer to a function nor (for C++) a pointer to a member function, is treated as if it is the base expression of a zero-offset assumed-size array that appears in a map clause with the alloc map-type."

However, since there is no “size” to assume for this “array” (because it’s a pointer), an explicit "map(alloc:s.data[0:N])" is required on the target directive. When this is added, the target region executes without error and assigns the correct results to s.data.

spophale:
closes #817 

This resolves the missing data issue. But the test fails with both LLVM clang and amdclang as neither account for the stride specified in the target update.

Example:

`#pragma omp target update from(s.data[:N/2:2]) //only update even array elements`

Results in:
```
s.data[0]:10.000000 
s.data[1]:10.000000 
s.data[2]:10.000000 
s.data[3]:10.000000 
s.data[4]:10.000000 
s.data[5]:0.000000 
s.data[6]:0.000000 
s.data[7]:0.000000 
s.data[8]:0.000000 
s.data[9]:0.000000 
```

### DIFF
--- a/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
+++ b/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
@@ -38,7 +38,7 @@ int target_update_from_mapper() {
     #pragma omp target
     { 
       for (int i = 0; i < s.len; i++) {
-        s.data[i] = 10;
+        s.data[i] = i;
       }
     }//end target
   
@@ -50,7 +50,7 @@ int target_update_from_mapper() {
       OMPVV_TEST_AND_SET(errors, s.data[i] != 0);
     }
     else{
-      OMPVV_TEST_AND_SET(errors, s.data[i] != 10);
+      OMPVV_TEST_AND_SET(errors, s.data[i] != i);
     }
   }
 
--- a/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
+++ b/tests/5.0/target_update/test_target_update_mapper_from_discontiguous.c
@@ -14,17 +14,16 @@
 #include <stdlib.h>
 #include "ompvv.h"
 
-#define N 1024
+#define N 10
 
 typedef struct{
   size_t len;
   double *data;
 } T;
 
-int i;
 int errors = 0;
 
-#pragma omp declare mapper(S: T v) map(to: v, v.len, v.data[0:v.len])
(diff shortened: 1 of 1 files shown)


fel-cab:
I see that the same problem with Cray compiler on NVIDIA GPU system. While nvc, and gcc have not implemented mapper (it seems).
I would suggest to change the double to int, so we can compare equality without the problems that could happen with floating point numbers.

tob2 (commented):
> ..  gcc have not implemented mapper (it seems).

GCC supports 'mapper' but currently only in the OG vendor branches (e.g. OG13; GCC git: devel/omp/gcc-13; OG14 due soon). This feature is expected to get merged into mainline/GCC 15 relatively soon.

BTW: A quick test with GCC's OG13 shows: "Test passed on the device." :-)

>  I would suggest to change the double to int, so we can compare equality without the problems that could happen with floating point numbers.

Side remark: I concur although the first '2^(mantissa bits + 1)' integers can be represented exactly in IEEE 754 binary floating-point numbers – contrary to, e.g., 0.1 which cannot be represented at all. Thus, the original testcase should just work fine, especially as only assignments and no arithmetic operations were used _(2*i is still an integer and also 2.0*i should work) ._

_I do note that there is a minor test coverage loss: Before it had different values on the device side (0 ... N-1), now it has 10. — Thus, the corner case of copying multiple times the same device element  to the proper host array wouldn't be covered. — But, admittedly, this particular implementation bug is is extremely unlikely._

fel-cab (approved):

andrewkallai (approved):

PR #834, opened 2024-08-22
Declare Induction for 6.0

rjenaa:


rjenaa:
I will update the collector expression : The only variables allowed in a collector expression are omp_step and omp_idx.

spophale (approved):

andrewkallai (approved):

PR #848, opened 2025-01-06
Adding the tests for map-type modifier with default properties.

andrewkallai:
CC = clang 21.0.0git
test_map-type_modifier_default.c.o:
omptarget message: device mapping required by 'present' map type modifier does not exist for host address 0x00007ffe9ed6eda4 (4 bytes)
omptarget error: Call to getTargetPointer returned null pointer ('present' map type modifier).
omptarget error: Call to targetDataBegin failed, abort target.
omptarget error: Failed to process data before launching the kernel.
omptarget error: Consult https://openmp.llvm.org/design/Runtimes.html for debugging options.
omptarget error: Source location information not present. Compile with -g or -gline-tables-only.
omptarget fatal error 1: failure of target construct while offloading is mandatory
timeout: the monitored command dumped core

### DIFF
--- a/tests/6.0/map-type_modifier/test_map-type_modifier_default.c
+++ b/tests/6.0/map-type_modifier/test_map-type_modifier_default.c
@@ -20,11 +20,12 @@ int test_maptype_modifier_default() {
 
   #pragma omp target enter data map(to : a)
 
-  #pragma omp target map(present, from : a)
+  #pragma omp target map(present)
   {
     a = a * N;
   }
-  #pragma omp target update from(a)
+
+  #pragma omp target exit data map(from: a)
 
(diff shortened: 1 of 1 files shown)


tests/6.0/map-type_modifier/test_map-type_modifier_default.c
```diff
@@ -20,11 +20,12 @@ int test_maptype_modifier_default() {
 
   #pragma omp target enter data map(to : a)
 
-  #pragma omp target map(present, from : a)
+  #pragma omp target map(present)
```
tob2: You lost the variable name:
```suggestion
  #pragma omp target map(present : a)
```

spophale (approved):

seyonglee (approved):

PR #854, opened 2025-02-06
Inital test for local clause

rjenaa:


rjenaa:
Local isn't recognized yet with declare target on LLVM

spophale (approved):

PR #874, opened 2025-08-05
Adding the first version of the safesync test.

andrewkallai:
Compiles with clang 22, but ignores the safesync clause.
gcc 15.1 cannot compile the test.

### DIFF
--- a/tests/6.0/parallel/test_parallel_safesync.c
+++ b/tests/6.0/parallel/test_parallel_safesync.c
@@ -1,4 +1,4 @@
-//--------------- test_directive_clause.c ------------------------------------//
+//--------------- test_parallel_safesync.c------------------------------------//
 // OpenMP API Version 6.0 November 2024
 // Pg. 901, line 9
 // ***********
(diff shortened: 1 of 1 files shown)


tests/6.0/parallel/test_parallel_safesync.c
```diff
@@ -0,0 +1,64 @@
...
+//----------------------------------------------------------------------------//
+#include "ompvv.h"
+#include <omp.h>
+
+#define N 100
+
+int test_directive() {
+  int errors = 0;
+  int a[N], b[N];
+  int count1 = 0, count2 = 0;
```
seyonglee, line 25: In line 25, `count2 = 0;` should be changed to `count2 = 1;`. With `count2 = 0;` no thread can exit from the first do-while loop in line 37 - 40. The first thread entering the do-while loop will have `t = 1` and `u = 0` and thus cannot exit (`while (0 < 1)`), and all the later threads will be stuck in the same loop (`while(0<2)`, `while(0<3)`, ...). With `count2 = 1;` threads will exit from the first do-while loop one by one (`while(1<1)`, `while(2<2)`, …) (whenever previous thread increases `count2` in line 46).
(I think that the original example in the 6.0 example document has an error.)

PR #875, opened 2025-08-05
Adding the test for the self modifier on map.

andrewkallai:
Does not compile with clang 22 nor gcc 15.1.

### DIFF
--- a/tests/6.0/target/test_target_map_type_modifier_self.c
+++ b/tests/6.0/target/test_target_map_type_modifier_self.c
@@ -1,0 +1,44 @@
+//--------------- test_target_map_type_modifier_self.c -----------------------//
+// OpenMP API Version 6.0 November 2024
+// Pg. 899, line 14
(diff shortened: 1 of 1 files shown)


tests/6.0/target/test_target_map_type_modifier_self.c
```diff
@@ -0,0 +1,44 @@
...
+
+int test_map_type_modifier_self() {
+  int errors = 0;
+  int scalar_value = 0;
+  int *host_address = &scalar_value;
+  int *device_address = NULL;
+
+  #pragma omp target map(self : scalar_value) map(tofrom : device_address)
+  {
+    scalar_value += 1;
```
seyonglee, line 25: This may cause a segmentatiion fault if the target device does not support unified shared memory. 
Either delete this access statement or enable unified shared memory.

seyonglee (approved):

Issue #862, opened 2025-03-31
do-loop variable interaction with OpenMP loop directives


ivan-pi:
**Directive and clause it applies to**

Applies to the the `!$omp loop` directive

**Specification details**

In a Fortran do-loop, the loop iteration variable ("do-variable") are supposed to remain defined upon exit from the loop. The OpenMP standard tries to remain consistent with this behaviour with the statement, OpenMP 5.2 spec, Chapter 5, page 96, line 18, 

> The loop iteration variable in any associated loop of a `loop` construct is lastprivate.

**Pseudocode**

```
program test_omp_loop
!$ use omp_lib
   implicit none
   integer, parameter :: n = 20
   real :: a(n)
   integer :: i, nt

!$omp parallel

   !$omp single
   nt = omp_get_num_threads()
   !$omp end single

   !$omp loop shared(a)
   do i = 1, n
      a(i) = 42
   end do
!$omp end parallel

   if (i == n+1) then
      print *, "PASSED on ", nt, "threads"
   else
      print *, "FAILED"
      print *, "i = ", i, ", expected ", n+1
   end if

end program
```

**Any other comments**

When testing the program above using the Intel Fortran compiler, it failed depending on the optimization flag:

```
(base) di75yas@cm4login1:~/omp> make
ifx -fiopenmp -O0 -o test_omp_loop_O0 test_omp_loop.f90
ifx -fiopenmp -O2 -o test_omp_loop_O2 test_omp_loop.f90
(base) di75yas@cm4login1:~/omp> make run
./test_omp_loop_O0
 FAILED
 i =            0 , expected           21
./test_omp_loop_O2
 PASSED on            1 threads
```

This behavior should be resolved with the oneAPI HPC Toolkit 2025.0.

ivan-pi:
Thanks for adding the test. I noticed the added test applies to `target`, but AFAICU it is possible to also use `loop` within `parallel`. Should that be a separate test? 

Just for the record, some compilers still struggle on this: https://godbolt.org/z/exxGKW8dT

PR #876, opened 2025-08-12
Adding the initial test.

andrewkallai:
Test compiles and passes with gfortran 15.1
fails with flang
[header test_loop_iterator.F90:24]  Condition condition failed
[header test_loop_iterator.F90:23] Test is running on device
[header ompvv.F90:253] The value of errors is 2
[OMPVV_RESULT test_loop_iterator.F90] Test failed on the device.
Fortran STOP: code 1

This PR resolves issue #862

### DIFF
--- a/tests/5.0/loop/test_loop_iterator.F90
+++ b/tests/5.0/loop/test_loop_iterator.F90
@@ -28,23 +28,18 @@ CONTAINS
   INTEGER FUNCTION test_loop_iterator()
   INTEGER :: errors = 0
   REAL :: a(N)
-  INTEGER :: i, num_threads = 0
+  INTEGER :: i, temp = 0
 
-  !$omp parallel
-  !$omp single
-    num_threads = omp_get_num_threads()
-  !$omp end single
-
-  !$omp target map(a, i)
+  !$omp target map(from: temp)
   !$omp loop
   DO i = 1, N
     a(i) = i
   END DO
   !$omp end loop
+  temp = i
   !$omp end target
-  !$omp end parallel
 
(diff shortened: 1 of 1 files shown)


tests/5.2/loop/test_loop_iterator.F90
```diff
@@ -0,0 +1,48 @@
...
+  INTEGER :: errors = 0
+  REAL :: a(N)
+  INTEGER :: i, num_threads = 0
+
+  !$omp parallel
+  !$omp single
+    num_threads = omp_get_num_threads()
+  !$omp end single
+
+  !$omp loop shared(a)
```
seyonglee, line 37: The `loop` directive does not allow `shared` clause (OpenMP API Version 5.0, Section 2.9.5 loop Construct, page 128, line 13-19).

spophale, line 37: !$omp target loop map(a,i)

spophale (approved):

tests/5.0/loop/test_loop_iterator.F90
```diff
@@ -0,0 +1,51 @@
...
+  INTEGER :: errors = 0
+  REAL :: a(N)
+  INTEGER :: i, num_threads = 0
+
+  !$omp parallel
+  !$omp single
+    num_threads = omp_get_num_threads()
+  !$omp end single
+
+  !$omp target map(a, i)
```
spophale, line 38:
```suggestion
  !$omp target map(from: temp) //temp is an int declared outside
```

seyonglee (approved):

PR #878, opened 2025-09-04
Implement GPU check and skip test if none found

spophale:
Add warning and skip logic for tests without GPUs


andrewkallai (approved):
This test can be merged if the OMPVV_SKIPPED_EXIT_CODE macro is added to the header file first.

PR #882, opened 2025-09-22
Fix Makefile error: unterminated call to function 'if'

mhalk:
Fix if statement syntax to avoid error introduced by https://github.com/OpenMP-Validation-and-Verification/OpenMP_VV/pull/860:

```
Makefile:316: *** unterminated call to function 'if': missing ')'.  Stop.
```


andrewkallai:
@mhalk thank you, we are taking a look at this fix

andrewkallai (approved):

PR #889, opened 2025-11-06
Adding the first workdistribute test

andrewkallai:
Compiles and crashes with flang 22.
Compiles and dumps core with gfortran 14

For gfortran this happens when
integer, parameter :: N >= 1024 * 1016

### DIFF
--- a/tests/6.0/workdistribute/test_workdistribute_directive.F90
+++ b/tests/6.0/workdistribute/test_workdistribute_directive.F90
@@ -74,7 +74,7 @@ program test_omp_workdistribute
 
 
     integer :: errors = 0
-    integer, parameter :: N = 1024 * 1024
+    integer, parameter :: N = 1024
 
     real :: a
     real :: x0


tests/6.0/workdistribute/test_workdistribute_directive.F90
```diff
@@ -0,0 +1,56 @@
+!===----test_workdistribute_directive.F90-----------------------------------===//
+!
+! OpenMP API Version 6.0 
+!
+! This test checks if the workdistribute passed to the nowait clause
```
seyonglee: Update comments

tests/6.0/workdistribute/test_workdistribute_directive.F90
```diff
@@ -0,0 +1,57 @@
...
```
spophale:
```suggestion
! This test checks if the workdistribute construct works with target teams
```

seyonglee (approved):

PR #891, opened 2025-11-13
Remove shared memory checks from is_accessible tests w/o USM requirement

mhalk:
Remove the shared memory check, since page-migration and USM are different.

`HSA_XNACK=0/1` affects the return value of `omp_target_is_accessible` (false/true).
But `isSharedMemory` would remain false in both cases on a non-USM architecture.

Hence, non-USM target archs would fail these tests when page-migration is active
 - For example setting `HSA_XNACK=1` when using a gfx90a target would fail
 - Additonally forcing USM via pragma or `OMPX_APU_MAPS=1` would pass

### DIFF
--- a/tests/5.1/target/test_target_is_accessible.F90
+++ b/tests/5.1/target/test_target_is_accessible.F90
@@ -2,7 +2,7 @@
 !
 ! OpenMP API Version 5.1 Nov 2020
 !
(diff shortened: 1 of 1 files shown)


seyonglee (approved):

mhalk:
Thank you for the prompt support :)

PR #901, opened 2026-01-29
Taskgraph - matching taskgraph record in parallel threads

rpereira-dev:
See https://github.com/OpenMP-Validation-and-Verification/OpenMP_VV/pull/886#discussion_r2695473514

In other words: what should we expect from a taskgraph record creation/replay if it is matching in parallel on multiple threads?

### DIFF
--- a/tests/6.0/taskgraph/test_taskgraph_parallel.c
+++ b/tests/6.0/taskgraph/test_taskgraph_parallel.c
@@ -1,0 +1,81 @@
+//===-- test_taskgraph.c ------------------------------------------------===//
+//
+// OpenMP API Version 6.0 Nov 2024
+//
+// Description
+// testTaskgraphParallel():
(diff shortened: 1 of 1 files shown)


rpereira-dev:
Not much, but I am currently going intensively through past ARB discussions, minutes and specs to address on-going issues with data environment specifications.

I agree that these races are users responsibility.
I disagree on your implication: taskgraph can still appear in a parallel region (more generally, multiple concurrent replay of the same taskgraph is allowed). Programmers must ensure the absence of data races themselves too, with additional synchronizations.

Though, I think the current specs on data environment makes multiple concurrent replay hardly praticable for programmers. Still gotta do some archaeology on why specs are this way, and then I may suggest modifications.

rpereira-dev:
For OMPVV, I suggest we let this issue/pr open until further feedback on multiple concurrent replay of the same record

PR #909, opened 2026-06-10
Issue 908: Fix 5.1 test_target_update_iterator.F90

### LINKED ISSUE #908: 5.1/target_update/test_target_update_iterator.F90: Invalid use of map clause
The following seems to be a difference between Fortran and C – and seems to be resolvable using OpenMP 6.0's **ref_ptee** modifier.

Namely, the code (added in PR #769) does:
https://github.com/OpenMP-Validation-and-Verification/OpenMP_VV/blob/4843b5ad559ea45ad94e5c6e22c9baf0fe54d6b5/tests/5.1/target_update/test_target_update_iterator.F90#L67-L68

GCC fails when executing the second line with:
```
libgomp: Trying to map into device [0x7fff7cf68170..0x7fff7cf681b8) object when [0x7fff7cf68170..0x7fff7cf681b0) is already mapped
```

**The reason is that GCC maps:**
* First line: Map the pointee – and the pointer ('data' component), only
* Second line: Explicit map clause requires that the whole derived type gets mapped → 'len' has to be additionally mapped but can't as 'new_struct%data' has already been mapped.

* * *
NOTE that compilers differ here. Assume:
```F90
  TYPE test_struct
    INTEGER, POINTER:: data(:)
    INTEGER :: len  ! NOTE Reversed order to OpenMP_VV
  END TYPE
…
print *, 'one'
        !$omp target enter data map(to: new_struct%data)
print *, omp_target_is_present(c_loc(new_struct%data(1)), omp_get_default_device())
print *, omp_target_is_present(c_loc(new_struct%len), omp_get_default_device())
print *, omp_target_is_present(c_loc(new_struct), omp_get_default_device())
print *, 'two'
        !$omp target map(to: new_struct, new_struct%data) map(tofrom: A) private(i)
…
        !$omp end target 
print *, 'three'
```

For GCC, the result is as described:
```
one 
          1  → pointee is mapped = c_loc(new_struct%data(1))  [note: identical to: c_loc(new_struct%data)]
          0  → 'len' is not mapped = c_loc(new_struct%len)
          1  → the array-descriptor / pointer 'new_struct%data' is mapped =~= c_loc(new_struct)  (as 'data' is the first component)
 two
libgomp: Trying to map into device [0x7fff7cf68170..0x7fff7cf681b8) object when [0x7fff7cf68170..0x7fff7cf681b0) is already mapped
```

However, trying HPE's CCE ftn (Cray 18.0.1) the output is the following _[I had to replace the 'iterator' as that's not supported in that version.]_
```
 one
 1
 1
 1
 two
 three
```
Namely: That compiler maps all of 'new_struct'.

IMHO, the latter is permitted but not required – such that GCC's behavior is conforming, which renders the program invalid.

***
BTW: 'i' is used initialized due to the implicit 'firstprivate(i)' – solution: Use 'private'. _(For 'target parallel for', it would be implicitly private as loop iteration variable but not for 'target'.)_
```diff
--- a/tests/5.1/target_update/test_target_update_iterator.F90
+++ b/tests/5.1/target_update/test_target_update_iterator.F90
@@ -68 +68 @@ CONTAINS
-        !$omp target map(to: new_struct, new_struct%data) map(tofrom: A) 
+        !$omp target map(to: new_struct, new_struct%data) map(tofrom: A) private(i)
@@ -78 +78 @@ CONTAINS
-        !$omp target map(tofrom: A)
+        !$omp target map(tofrom: A) private(i)
```

tob2:
fixes #908
* Issue #908

The code used `target enter data map(to: new_struct%data)` to map part of the derived type - and later tries to explicitly map all of the derived type: `target map(to: new_struct, new_struct%data)`.

A permissible implementation is to only map parts of 'new_struct' in the 'enter data' directive such that mapping all of 'new_struct' is not permitted (unspecified behavior), leading to a runtime fail.

**Solution** is to **either** only _implicitly map_ 'new_struct', which handles partial mapping - **or** as **done here:** `target map(to: new_struct%data)`, i.e. _only_ mapping what is already present on the device (no actual data copy happening here).

Additionally, `private(i)` has been added as otherwise an _implicit_ `firstprivate(i)` is added that uses uninitialized memory.

@fel-cab @spophale @seyonglee @andrewkallai

### DIFF
--- a/tests/5.1/target_update/test_target_update_iterator.F90
+++ b/tests/5.1/target_update/test_target_update_iterator.F90
@@ -65,7 +65,7 @@ CONTAINS
         CALL init(new_struct)
 
         !$omp target enter data map(to: new_struct%data)
-        !$omp target map(to: new_struct, new_struct%data) map(tofrom: A) 
+        !$omp target map(to: new_struct%data) map(tofrom: A) private(i)
         ! pointer attachment
         DO i = 1, N
             A(i) = new_struct%data(i)
@@ -75,7 +75,7 @@ CONTAINS
         CALL init_again(new_struct)
         ! update with new values, (i*2)+1
         !$omp target update to(iterator(it = 1:N): new_struct%data(it))
-        !$omp target map(tofrom: A)
+        !$omp target map(tofrom: A) private(i)
         DO i = 1, N
             A(i) = A(i) + new_struct%data(i)
         END DO


tob2:
@fel-cab @spophale @seyonglee @andrewkallai — **REVIEW PING**

spophale (approved):

seyonglee (approved):

tob2:
> @spophale  approved
> @seyonglee  approved

Thanks for the review and approval! Unless there is more work to do, someone needs now to actually merge the changes …

tob2:
@spophale @seyonglee – Thanks for the approvals.
As followup, can someone actually merge this change? Thanks!

PR #910, opened 2026-06-10
 Issue 907: Make 5.1/atomic/test_atomic_fail_*.{c,F90} less racy

### LINKED ISSUE #907: tests/5.1/atomic/test_atomic_fail_seq_cst.c has a race condition
The code following code has a race condition — there is an ordering imposed by the atomic access – but there is no guarantee that the memory for 'X' is has the updated value from the first thread in the second thread when 'Y' is updated.

It is somehow likely as the code executed between 'Y' is set in the first thread and 'X is atomically set is rather simple while the code to handle 'Y' and then process 'X' in the second thread is more involved, but still.

Indeed when running this testcase repeatedly in GCC, it will eventually fails because X is still 0 in the second thread and not yet 1. That happens relatively rare, but still!

```C
  int x = 0, y = 0; // […]
#pragma omp parallel num_threads(2)
   {
      int thrd = omp_get_thread_num();
       if (thrd == 0) {
          y = 1;
          #pragma omp atomic write seq_cst
          x = 10;
       } else {
          while (y != 5) {
            #pragma omp atomic compare acquire fail(seq_cst)
            if(y == 1){
               y = 5;
            }
          }
          OMPVV_TEST_AND_SET(errors, x != 10);
          OMPVV_TEST_AND_SET(errors, y != 5);
       }
   }
```

Note that the generated code by GCC seems to be fine:

The **first thread** does (`-fdump-tree-optimized`; C-like output):
```C
   _16 = *.omp_data_i_5(D).y;
  *_16 = 1;
  _18 = *.omp_data_i_5(D).x;
  __atomic_store_4 (_18, 10, 5);
// with '__atomic_store_n (type *ptr, type val, int memorder)' and 5 = MEMMODEL_SEQ_CST
```

If you are interested in GCC's assembly code, on x86-64 it is:
```as
.L2:
        .cfi_restore_state
        movq    8(%rbx), %rax
        movl    $1, (%rax)
        movq    (%rbx), %rdx
        movl    $10, %eax
        xchgl   (%rdx), %eax
        jmp     .L1
```
And Clang's version is
```as
# %bb.1:
	movq	-72(%rbp), %rcx                 # 8-byte Reload
	movq	-80(%rbp), %rax                 # 8-byte Reload
	movl	$1, (%rax)
	movl	$10, %eax
	xchgl	%eax, (%rcx)
	leaq	.L__unnamed_1(%rip), %rdi
	callq	__kmpc_flush@PLT
	jmp	.LBB1_6
```

For the **second thread,** there is first a `if (y != 5)` check — and then the loop:
```C
<bb 5> [local count: 955630224]:
  # _25 = PHI <_6(5), _21(3)>
  .ATOMIC_COMPARE_EXCHANGE (_25, 1, 5, 4, 5, 5);  /* →  'if (y == 1) y = 5' (4 = 4 byte type + weak false, 2 × '5' = 'seq_cst') */
  _6 = *.omp_data_i_5(D).y;
  y.0_7 = *_6;
  if (y.0_7 != 5)
    goto <bb 5>; [89.00%]
  else
    goto <bb 4>; [11.00%]

  <bb 4> [local count: 118111600]:
  _8 = *.omp_data_i_5(D).x;
  x.1_9 = *_8;
  _10 = x.1_9 != 10;
…
```
This is translated into the following assembly by GCC:
```as
.L3:
        movl    %esi, %eax
        lock cmpxchgl   %ecx, (%rdx)
        movq    8(%rbx), %rdx
        cmpl    $5, (%rdx)
        jne     .L3
.L5:
        movq    (%rbx), %rax
        cmpl    $10, (%rax)
        setne   %al
        movzbl  %al, %eax
        addl    %eax, 16(%rbx)
…
```
and by Clang, respectively:
```as
.LBB1_3:                                # =>This Inner Loop Header: Depth=1
	movq	-80(%rbp), %rax                 # 8-byte Reload
	cmpl	$5, (%rax)
	je	.LBB1_5
# %bb.4:                                #   in Loop: Header=BB1_3 Depth=1
	movq	-80(%rbp), %rcx                 # 8-byte Reload
	movl	$1, %eax
	movl	$5, %edx
	lock		cmpxchgl	%edx, (%rcx)
	jmp	.LBB1_3
.LBB1_5:
	movq	-64(%rbp), %rax                 # 8-byte Reload
	movq	-80(%rbp), %rcx                 # 8-byte Reload
	movq	-72(%rbp), %rdx                 # 8-byte Reload
	cmpl	$10, (%rdx)
	setne	%dl
	andb	$1, %dl
	movzbl	%dl, %edx
	movl	%edx, -48(%rbp)
	movl	-48(%rbp), %edx
	addl	(%rax), %edx
	movl	%edx, (%rax)
	cmpl	$5, (%rcx)
	setne	%cl
	andb	$1, %cl
	movzbl	%cl, %ecx
	movl	%ecx, -52(%rbp)
	movl	-52(%rbp), %ecx
	addl	(%rax), %ecx
	movl	%ecx, (%rax)
```

tob2:
fixes #907
* Issue #907

The current code sets `y = 1` and then atomically `x = 10` in the first thread, expecting that once 'y' shows up as updated in the second thread, 'x' is also updated. - But this is racy.

With this commit, 'y' is set non atomically after the atomic write.

This applies to all 6 `tests: tests/5.1/atomic/test_atomic_fail_{acquire,relexed,seq_cst}.{c,F90}.`

### DIFF
--- a/tests/5.1/atomic/test_atomic_fail_acquire.c
+++ b/tests/5.1/atomic/test_atomic_fail_acquire.c
@@ -25,9 +25,9 @@ int test_atomic_fail_acquire() {
    {
       int thrd = omp_get_thread_num();
        if (thrd == 0) {
-          y = 1;
           #pragma omp atomic write seq_cst
           x = 10;
+          y = 1;
        } else {
           while (y != 5) {
             #pragma omp atomic compare seq_cst fail(acquire)
--- a/tests/5.1/atomic/test_atomic_fail_acquire.F90
+++ b/tests/5.1/atomic/test_atomic_fail_acquire.F90
@@ -36,10 +36,10 @@ CONTAINS
     !$omp parallel num_threads(2) private(thrd)
     thrd = omp_get_thread_num()
     IF( thrd .EQ. 0 ) THEN
-      y = 1
       !$omp atomic write seq_cst 
       x = 10
       !$omp end atomic
+      y = 1
     ELSE
       DO WHILE ( y .NE. 5 )
         !$omp atomic compare seq_cst fail(acquire)
--- a/tests/5.1/atomic/test_atomic_fail_relaxed.F90
+++ b/tests/5.1/atomic/test_atomic_fail_relaxed.F90
@@ -36,10 +36,10 @@ CONTAINS
     !$omp parallel num_threads(2) private(thrd)
     thrd = omp_get_thread_num()
     IF( thrd .EQ. 0 ) THEN
-      y = 1
       !$omp atomic write seq_cst 
       x = 10
       !$omp end atomic
+      y = 1
     ELSE
       DO WHILE ( y .NE. 5 )
         !$omp atomic compare fail(relaxed)
--- a/tests/5.1/atomic/test_atomic_fail_relaxed.c
+++ b/tests/5.1/atomic/test_atomic_fail_relaxed.c
@@ -25,9 +25,9 @@ int test_atomic_fail_relaxed() {
    {
       int thrd = omp_get_thread_num();
        if (thrd == 0) {
-          y = 1;
           #pragma omp atomic write seq_cst
           x = 10;
+          y = 1;
        } else {
           while (y != 5) {
             #pragma omp atomic compare fail(relaxed)
--- a/tests/5.1/atomic/test_atomic_fail_seq_cst.F90
+++ b/tests/5.1/atomic/test_atomic_fail_seq_cst.F90
@@ -37,10 +37,10 @@ CONTAINS
     !$omp parallel num_threads(2) private(thrd, tmp)
     thrd = omp_get_thread_num()
     IF( thrd .EQ. 0 ) THEN
-      y = 1
       !$omp atomic write seq_cst 
       x = 10
       !$omp end atomic
+      y = 1
     ELSE
       tmp = 0
       DO WHILE ( y .NE. 5 )
--- a/tests/5.1/atomic/test_atomic_fail_seq_cst.c
+++ b/tests/5.1/atomic/test_atomic_fail_seq_cst.c
@@ -25,9 +25,9 @@ int test_atomic_fail_seq_cst() {
    {
       int thrd = omp_get_thread_num();
        if (thrd == 0) {
-          y = 1;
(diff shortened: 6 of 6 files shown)


tob2:
@spophale @fel-cab @seyonglee – Please review.

_(Cross-ref: last change to those files was done in Pull Req. #776)_

tob2:
@spophale @fel-cab @seyonglee – **Review Ping**

seyonglee (approved):

tests/5.1/atomic/test_atomic_fail_acquire.c
spophale, line 36:
```suggestion
} else {
while (y!= 5) {
#pragma omp atomic compare seq_cst fail(acquire)
if(x == 10){
y = 5;
}
```

spophale: @tob2 , wondering if this is a better check. Is there a possibility for the code to deadlock ?

tob2: I am not sure I fully understand this comment. If I read the diff correctly, the proposed change is:
```diff
  #pragma omp atomic compare seq_cst fail(acquire)
- if (y == 1) {
+ if (x == 10) {
    y = 5;
```
However, comparing 'x' is no longer a _conditional-update structured block_ of type _cond-update-stmt_, which is either of the following:
```C
if(expr ordop x) x = expr;
if(x ordop expr) x = expr;
if(x == e) x = d;
{ r = x == e; if(r) x = d; }
```
With 'y' → 'x' in the diff, there are now two atomic variables, which does no work.

However, I might have misunderstood the proposal, also because the diff is shown in a rather odd way in Github.

tob2: @spophale – ? Could you clarify?

spophale: That makes sense. The intent was to have some relationship/dependance between the if and else, but then as you pointed out, it is not longer a conditional-update structured block.

spophale (approved):

PR #913, opened 2026-07-21
test_parallel_num_threads_prescriptiveness.c: Fix modifier use

tob2:
The _prescriptiveness_ modifier to **num_threads** is `strict` (and since TR15 (OpenMP 6.1) also `relaxed`) and not `prescriptiveness(strict)`.

Hence, in the following testcase `num_threads(prescriptiveness(strict) : N)` should actually be `num_threads(strict : N)` [cross ref: the file was added in Pull Req. #912]:

https://github.com/OpenMP-Validation-and-Verification/OpenMP_VV/blob/658ec4186cb57037d4a70b7a897440bb2419341f/tests/6.0/parallel/test_parallel_num_threads_prescriptiveness.c#L25

If you look at the specification (here: 6.0) `strict` is the keyword – in 6.1 (TR15) there is also `relaxed` – and both keywords run under _prescriptiveness_. [Note also that a type font is used for the keyword and for italic for the name.] 

grafik

* * *
@andrewkallai @seyonglee @spophale – Please review.

### DIFF
--- a/tests/6.0/parallel/test_parallel_num_threads_prescriptiveness.c
+++ b/tests/6.0/parallel/test_parallel_num_threads_prescriptiveness.c
@@ -8,10 +8,10 @@
 // ***********
 //
 // This test checks the prescriptiveness modifier for the num_threads clause on the
-// parallel directive. When prescriptiveness(strict) is specified, the implementation
+// parallel directive. When 'strict' is specified, the implementation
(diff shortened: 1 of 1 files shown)


seyonglee (approved):
Thank you for catching this.

spophale (approved):

PR #919, opened 2026-09-17
Adding initial version of loop collapse test.

andrewkallai:
Compiles with gcc/16.1.0


tests/5.1/loop/test_loop_collapse.F90
```diff
@@ -0,0 +1,44 @@
+!===----test_workdistribute_directive.F90-----------------------------------===//
```
spophale:
```suggestion
!===----test_loop_iterator_lastprivate.F90-----------------------------------===//
```

tests/5.1/loop/test_loop_collapse.F90
```diff
@@ -0,0 +1,44 @@
...
+
+    integer, parameter :: m = 32, n = 10
+    real :: a(m,n)
+    integer :: i, j
+
+    ! Section 5.1.1, pg. 97, line 16-17
+    ! The loop iteration variables in the associated loops of a
+    ! simd construct with multiple associated loops are lastprivate.
+
+    !$omp loop collapse(2) bind(thread)
```
spophale, line 26: Also try with:
 - loop inside parallel
 - loop inside target teams

